PR #1, opened 2020-04-24
Implement check, install, uninstall and create source commands

stefanprodan:
Changes:
- add check command 
  + kubectl >=1.18.0
  + kustomize >=3.5.0
  + kubernetes >=1.14.0
- add install manifests
  + RBAC role for CRs
  + RBAC cluster role for kustomize reconciler
  + deny ingress network policy
- add install command with dry-run option
- add uninstall command with dry-run option
  + prompt for confirmation
- add create source command for Git repositories
  + generate secret for basic-auth credentials
  + generate SSH host and deploy keys
  + generate secret for SSH keys
  + prompt for deploy key setup
- add end-to-end tests for:
  + check
  + install (local manifests)
  + create source (public repository)

### DIFF
--- a/cmd/tk/create_source.go
+++ b/cmd/tk/create_source.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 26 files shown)


cmd/tk/create_source.go
```diff
@@ -0,0 +1,236 @@
...
+	logAction("generating host key for %s", host)
+
+	keyscan := fmt.Sprintf("ssh-keyscan %s > %s/known_hosts", host, tmpDir)
+	if output, err := execCommand(keyscan); err != nil {
+		return fmt.Errorf("ssh-keyscan failed: %s", output)
+	}
+
+	logAction("generating deploy key")
+
+	keygen := fmt.Sprintf("ssh-keygen -b 2048 -t rsa -f %s/identity -q -N \"\"", tmpDir)
```
hiddeco, line 184: This could be done without depending on `ssh-keygen` using one of the `crypto` packages.

stefanprodan: If we want to get rid of openssh we need to replace keyscan as well. We could generate the host keys with something like https://github.com/Eun/sshkeys/blob/master/sshkeys.go

PR #9, opened 2020-04-28
Add git tag option to create source

stefanprodan:
Changes:
- add git tag option to create source
- add examples to main command
- format kustomization examples

### DIFF
--- a/cmd/tk/create_source.go
+++ b/cmd/tk/create_source.go
@@ -28,16 +28,30 @@ The create source command generates a source.fluxcd.io resource and waits for it
 For Git over SSH, host and SSH keys are automatically generated and stored in a Kubernetes secret.
 For private Git repositories, the basic authentication credentials are stored in a Kubernetes secret.`,
 	Example: `  # Create a source from a public Git repository master branch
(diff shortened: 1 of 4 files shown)


cmd/tk/main.go
```diff
@@ -16,11 +16,35 @@ var VERSION = "0.0.1"
...
+  tk install --version=master
+
+  # Create a source from a public Git repository
+  tk create source webapp \
+    --git-url=https://github.com/stefanprodan/podinfo \
+    --git-branch=master \
+    --interval=5m
+
+  # Create a kustomization for deploying a series of microservices
+  tk create create kustomization webapp \
```
hiddeco, line 37:
```suggestion
  tk create kustomization webapp \
```

cmd/tk/create_source.go
```diff
@@ -53,6 +68,7 @@ var (
 func init() {
```
hiddeco, line 68: Should have noticed this in an early review, but would it make sense to rename this to `create_gitrepository.go` and make the default `create source` type a `GitRepository`?

stefanprodan: The create source will have a `--helm-url` and then it will create Helm repos. Another option would be to introduce a `--type=git|helm`.

PR #10, opened 2020-04-29
Implement sync commands

stefanprodan:
Changes:
- add sync commands
  + sync git sources
  + sync kustomizations along with their sources
- refactor create commands
  + format `tk create [kind] [name] [flags]` 
- end-to-end testing
  + add create kustomization test
  + add sync kustomization with source test
  + add control plane check test
- improve cmd action reporting


cmd/tk/sync_kustomization.go
```diff
@@ -0,0 +1,108 @@
...
+			kustomization.Annotations[kustomizev1.SyncAtAnnotation] = time.Now().String()
+		}
+		if err := kubeClient.Update(ctx, &kustomization); err != nil {
+			return err
+		}
+		logSuccess("kustomization annotated")
+	}
+
+	logWaiting("waiting for kustomization sync")
+	if err := wait.PollImmediate(2*time.Second, timeout,
```
hiddeco, line 85: Given the repeated usage I would say this is either a `var` or global flag candidate.

PR #12, opened 2020-04-29
Implement delete, suspend and resume kustomization

stefanprodan:
Changes:
- add delete kustomization command with confirmation and warning if not suspended
- add suspend kustomization command
- add resume kustomization command
- add suspend/resume/delete e2e tests

### DIFF
--- a/cmd/tk/resume_kustomization.go
+++ b/cmd/tk/resume_kustomization.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1alpha1"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var resumeKsCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


cmd/tk/resume_kustomization.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var resumeKsCmd = &cobra.Command{
+	Use:     "kustomization [name]",
+	Aliases: []string{"ks"},
+	Short:   "Resume kustomization",
+	Long:    "The resume command enables the reconciliation of a previously suspended Kustomization resource and waits for it to finish the apply.",
```
hiddeco, line 16:
```suggestion
	Long:    "The resume command marks a previously suspended Kustomization resource for reconciliation and waits for it to finish the apply.",
```

Not sure if the above is the best suggestion, but it ensures we only adopt a single verb (`resume`) for the action.

PR #23, opened 2020-05-06
Release v0.0.1-alpha.1

stefanprodan:
Changelog will be auto-generated and posted on the release page.

### DIFF
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -11,7 +11,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/tk/main.go
```diff
@@ -11,7 +11,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
-var VERSION = "0.0.1"
+var VERSION = "0.0.1-alpha.1"
```
hiddeco, line 14: Might be an idea to build with `-ldflags "-X main.VERSION=``git <intelligent way to get tag>``”` for future releases.

PR #39, opened 2020-06-12
Add target path to the bootstrap cmd

stefanprodan:
This PR adds an optional flag to the bootstrap command that allows specifying a repository path where the manifests will be placed. When a path is specified, the cluster sync is configured for the target path. Using a path per cluster, a repository can be used to define the desired state of a fleet of clusters.

### DIFF
--- a/cmd/tk/bootstrap_github.go
+++ b/cmd/tk/bootstrap_github.go
@@ -43,6 +43,9 @@ the bootstrap command will perform an upgrade if needed.`,
   # Run bootstrap for a private repo owned by a GitHub organization
   bootstrap github --owner=<organization> --repository=<repo name>
 
+  # Run bootstrap for a repository path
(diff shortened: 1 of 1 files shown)


cmd/tk/bootstrap_github.go
```diff
@@ -417,7 +423,7 @@ func generateGitHubKustomization(url, name, namespace, tmpDir string, interval t
 			Interval: metav1.Duration{
 				Duration: 10 * time.Minute,
 			},
-			Path:  "./",
+			Path:  "./" + targetPath,
```
hiddeco, line 426: `filepath.Join()`?

stefanprodan: Doesn't work, as join removes `./` and that breaks the CRD validation.

stefanprodan: I added a trim to guard against double `./`.

PR #41, opened 2020-06-16
Assign GitHub teams during bootstrap

stefanprodan:
This PR allows cluster admins to assign org teams during bootstrap. The teams are added with maintain permissions.

### DIFF
--- a/cmd/tk/bootstrap_github.go
+++ b/cmd/tk/bootstrap_github.go
@@ -43,6 +43,9 @@ the bootstrap command will perform an upgrade if needed.`,
   # Run bootstrap for a private repo owned by a GitHub organization
   bootstrap github --owner=<organization> --repository=<repo name>
 
+  # Run bootstrap for a private repo and assign organization teams to it
+  bootstrap github --owner=<organization> --repository=<repo name> --team=<team1 slug> --team=<team2 slug>
+
   # Run bootstrap for a repository path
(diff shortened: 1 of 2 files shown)


cmd/tk/bootstrap_github.go
```diff
@@ -118,6 +124,17 @@ func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// add teams to org repository
+	if !ghPersonal {
+		for _, team := range ghTeams {
+			if err := addGitHubTeam(ctx, ghHostname, ghOwner, ghRepository, ghToken, team, ghDefaultPermission); err != nil {
+				logFailure(err.Error())
```
hiddeco, line 131: I would expect this to return as soon as it runs into an error, as the current code could result in no-one getting granted access to the repository while the bootstrap continues.

stefanprodan: Yes that was the intention as you can't make the operations transactional I opted for best-effort. Note that the user that runs the command has full access to the repo.

hiddeco: Ack. In that case I would expect a partly successful bootstrap run to return a non-zero exit code as an indicator.

stefanprodan: Done!

PR #74, opened 2020-07-02
Add notifications guide

stefanprodan:


docs/guides/notifications.md
```diff
@@ -0,0 +1,96 @@
+# Setup Notifications
+
+When operating a cluster, different teams may wish to receive notification about
```
hiddeco, line 3:
```suggestion
When operating a cluster, different teams may wish to receive notifications about
```

docs/guides/notifications.md
```diff
@@ -0,0 +1,96 @@
...
+      name: webapp-backend
+    - kind: Kustomization
+      name: webapp-frontend
+```
+
+Apply the above files or commit them to the `fleet-infra` repository.
+
+To verify that the alert has been acknowledge by the notification controller do:
+
+```text
```
hiddeco, line 72:
```suggestion
```console
```

stefanprodan: console is not suitable for commands and outputs in the same block, I used text like in the get-started guide.

hiddeco: Console _is meant to be used_ for command and outputs in the same block. See:

```console
$ kubectl -n gitops-system get alerts
NAME             READY   STATUS        AGE
on-call-webapp   True    Initialized   1m
$ kubectl -n gitops-system get alerts
NAME             READY   STATUS        AGE
on-call-webapp   True    Initialized   1m
```

PR #77, opened 2020-07-03
explain a bit more about Flux Toolkit

dholbach:
Point to #flux-dev, point to video and relevant repositories.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ toolkit.
 
 The project uses Slack: To join the conversation, simply join the
 [CNCF](https://slack.cncf.io/) Slack workspace and use the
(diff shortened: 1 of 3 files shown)


dholbach:
Do we want to include CONTRIBUTING.md in the navigation as well?

hiddeco:
@dholbach yes, I think we do. The document still makes mention of the `#flux` Slack channel while you changed it to `#flux-dev` in the `README.md`.

dholbach:
Shall we move CONTRIBUTING.md under docs? Or shall we symlink it somewhere?

Re: flux and flux-dev: do we want a separation between users (regular #flux) and contributors (#flux-dev)? In that case, it should be fine.

dholbach:
→ went for the symlink approach, LMK if that works for you

Issue #53, opened 2020-06-22
Create dev guide


stefanprodan:
We should create a development and contributing guide to the toolkit docs that covers both controllers and CLI dev workflows.

PR #81, opened 2020-07-08
Add source watcher dev guide

stefanprodan:
Fix: https://github.com/fluxcd/source-controller/issues/76

Ref: #53

Preview: https://toolkit.fluxcd.io/dev-guides/source-watcher/

### DIFF
--- a/docs/dev-guides/source-watcher.md
+++ b/docs/dev-guides/source-watcher.md
@@ -1,0 +1,233 @@
+# Watching for source changes
+
+In this guide you'll be developing a Kubernetes controller with
+[Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder)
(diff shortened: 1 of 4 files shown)


docs/dev-guides/source-watcher.md
```diff
@@ -0,0 +1,231 @@
...
+Port forward to source-controller artifacts server:
+
+```sh
+kubectl -n gitops-system port-forward svc/source-controller 8080:80
+```
+
+Export the local address as `SOURCE_HOST`:
+
+```sh
+export SOURCE_HOST=localhost:8080
```
paulcarlton-ww, line 75: port 8080 clashes with the source-watcher's metrics address listener, I used 9090 and it worked fine

stefanprodan: Good catch!

PR #87, opened 2020-07-14
Add reconcile helm commands

stefanprodan:
Changes:
- add `tk reconcile source helm`
- add `tk reconcile hr`
- update tk docs


cmd/tk/reconcile_helmrelease.go
```diff
@@ -0,0 +1,120 @@
...
+
+	err = kubeClient.Get(ctx, namespacedName, &helmRelease)
+	if err != nil {
+		return err
+	}
+
+	if helmRelease.Status.LastAppliedRevision != "" {
+		logger.Successf("reconciled revision %s", helmRelease.Status.LastAppliedRevision)
+	} else {
+		return fmt.Errorf("HelmRelease sync failed")
```
hiddeco, line 117:
```suggestion
		return fmt.Errorf("HelmRelease reconciliation failed")
```

cmd/tk/reconcile_helmrelease.go
```diff
@@ -0,0 +1,120 @@
...
+			return err
+		}
+	} else {
+		logger.Actionf("annotating HelmRelease %s in %s namespace", name, namespace)
+		if helmRelease.Annotations == nil {
+			helmRelease.Annotations = map[string]string{
+				helmv2.ReconcileAtAnnotation: time.Now().String(),
+			}
+		} else {
+			helmRelease.Annotations[helmv2.ReconcileAtAnnotation] = time.Now().String()
```
hiddeco, line 93: Noticed the other day that the string this creates is kind of awkward: `2009-11-10 23:00:00 +0000 UTC m=+0.000000001`.

stefanprodan: Should we use `metav1.Now().String()` like we do in KS/HC?

hiddeco: Does that print something else? As `metav1.Now()` is just a wrapper around `time` to able to (un)marshal.

stefanprodan: I've use RFC3339Nano

PR #88, opened 2020-07-14
Add go-git-providers proposal

luxas:
Let's unify talking to Git providers once and for all with a new repo & library: `go-git-providers` :tada:

cc @stefanprodan @hiddeco @squaremo @martina-if @bboreham

### DIFF
--- a/docs/proposals/go-git-providers.md
+++ b/docs/proposals/go-git-providers.md
@@ -19,9 +19,10 @@ client for multiple providers and domains.
   - For a given **Organization**:
     - **Teams**: `GET` and `LIST`
   - **Repositories**: `GET`, `LIST` and `POST`
-    - **Teams**: `ADD` and `REMOVE`
-    - **Credentials**: `GET`, `LIST` and `POST`
+    - **Team Access**: `LIST`, `POST` and `DELETE`
+    - **Credentials**: `LIST`, `POST` and `DELETE`
 - Support sub-organizations (or "sub-groups" in GitLab) if possible
+- Support reconciling an object for idempotent operations
 - Pagination is automatically handled for `LIST` requests
 - Transparently can manage teams (collections of users, sub-groups in Gitlab) with varying access to repos
 - Follow library best practices in order to be easy to vendor (e.g. use major `vX` versioning & go.mod)
@@ -43,7 +44,6 @@ client for multiple providers and domains.
 - It should be possible to create a fake API client for testing, implementing the same interfaces
 - All type structs shall have a `Validate()` method, and optionally a `Default()` one
 - All type structs shall expose their internal representation (from the underlying library) through the `InternalGetter` interface with a method `GetInternal() interface{}`
-- By default, mutating methods (e.g. `Create()`) shall be idempotent, unless opted out in `*CreateOptions` field.
 - Typed errors shall be returned, wrapped using Go 1.14's new features
 - Go-style enums are used when there are only a few supported values for a field
 - Every field is documented using Godoc comment, including `+required` or `+optional` to clearly signify its importance
@@ -194,6 +194,7 @@ The `OrganizationsClient` provides access to a set of organizations, as follows:
 type OrganizationsClient interface {
 	// Get a specific organization the user has access to
 	// This might also refer to a sub-organization
+	// ErrNotFound is returned if the resource does not exist
 	Get(ctx context.Context, o OrganizationRef) (*Organization, error)
 
 	// List all top-level organizations the specific user has access to
@@ -219,6 +220,7 @@ type Organization struct {
 	// (Domain, Organization and SubOrganizations) required for being an OrganizationRef
 	OrganizationInfo `json:",inline"`
 	// InternalHolder implements the InternalGetter interface
+	// +optional
 	InternalHolder `json:",inline"`
 
 	// Name is the human-friendly name of this organization, e.g. "Weaveworks" or "Kubernetes SIGs"
@@ -273,6 +275,7 @@ type OrganizationTeamsClient interface {
 	// Get a team within the specific organization
 	// teamName may include slashes, to point to e.g. "sub-teams" i.e. subgroups in Gitlab
 	// teamName must not be an empty string
+	// ErrNotFound is returned if the resource does not exist
 	Get(ctx context.Context, teamName string) (*Team, error)
 
 	// List all teams (recursively, in terms of subgroups) within the specific organization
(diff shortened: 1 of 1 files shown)


docs/proposals/go-git-providers.md
```diff
@@ -0,0 +1,474 @@
...
+  - one collection-specific interface, with a plural name (e.g. `OrganizationsClient`), that has methods like `Get()` and `List()`
+  - one instance-specific interface, with a singular name (e.g. `OrganizationClient`), that operates on that instance, e.g. allowing access to child resources, e.g. `Teams()`
+- Every `Create()` signature shall have a `{Resource}CreateOptions` struct as the last argument.
+  - `Delete()` and similar methods may use the same pattern if needed
+- All `*Options` structs shall be passed by value (i.e. non-nillable) and contain only nillable, optional fields
+- All optional fields in the type structs shall be nillable
+- It should be possible to create a fake API client for testing, implementing the same interfaces
+- All type structs shall have a `Validate()` method, and optionally a `Default()` one
+- All type structs shall expose their internal representation (from the underlying library) through the `InternalGetter` interface with a method `GetInternal() interface{}`
+- By default, mutating methods (e.g. `Create()`) shall be idempotent, unless opted out in `*CreateOptions` field.
```
luxas, line 45: let me know how you'd like to handle idempotency by default

docs/proposals/go-git-providers.md
```diff
@@ -0,0 +1,474 @@
...
+- Typed errors shall be returned, wrapped using Go 1.14's new features
+- Go-style enums are used when there are only a few supported values for a field
+- Every field is documented using Godoc comment, including `+required` or `+optional` to clearly signify its importance
+
```
luxas, line 49: Possibly add "Serialize nicely to JSON (if needed for e.g. debugging)"

docs/proposals/go-git-providers.md
```diff
@@ -0,0 +1,474 @@
...
+    - **Teams**: `ADD` and `REMOVE`
+    - **Credentials**: `GET`, `LIST` and `POST`
+- Support sub-organizations (or "sub-groups" in GitLab) if possible
+- Pagination is automatically handled for `LIST` requests
+- Transparently can manage teams (collections of users, sub-groups in Gitlab) with varying access to repos
+
+## Non-goals
+
+- Support for features not mentioned above
+
```
luxas, line 31: Maybe we should specify some kind of versioning pattern, e.g. go.mod and with `v1`, ... `v30` in the import path like e.g. the GitHub library does?

docs/proposals/go-git-providers.md
```diff
@@ -0,0 +1,474 @@
...
+`CredentialsClient` allows adding & removing credentials (e.g. deploy keys) from accessing a specific repository.
+
+```go
+// CredentialsClient operates on the access credential list for a specific repository
+type CredentialsClient interface {
+	// Create a credential with the given human-readable name, the given bytes and optional options
+	Create(ctx context.Context, c RepositoryCredential, opts CredentialCreateOptions) error
+
+	// Lists all credentials
+	List(ctx context.Context) ([]RepositoryCredential, error)
```
luxas, line 427: Do we have a practical use case of List or can we make these just `Add` and `Remove`?
If so, this List call probably should also take in a `type RepositoryCredentialType`
cc @martina-if @stefanprodan

stefanprodan:
For `tk` the create functions are in fact reconcile, as in create or update or no-op. Maybe add `Reconcile` to each interface?

docs/proposals/go-git-providers.md
```diff
@@ -0,0 +1,474 @@
...
+(import name: `gitprovider`), which provides an abstraction layer for talking to Git providers
+like GitHub, GitLab and Bitbucket.
+
+This would become a new repository, specifically targeted at being a general-purpose Git provider
+client for multiple providers and domains.
+
+## Goals
+
+- Support multiple Git provider backends (e.g. GitHub, GitLab, Bitbucket, etc.) using the same interface
+- Support talking to multiple domains at once, including custom domains (e.g. talking to "gitlab.com" and "version.aalto.fi" from the same client)
```
squaremo, line 15: What is the motivation for this? It sounds tricky to get right.

stefanprodan: Indeed, I see no use-case for this.

luxas:
> For tk the create functions are in fact reconcile, as in create or update or delete/create or no-op. Maybe add Reconcile to each interface?

`Reconcile(ctx, objs, opts)` is a good fit for the RepositoryTeams and RepositoryCredentials, but not for e.g. all personal repos or all organizations IMO. I'll add the method to those two interfaces.

luxas:
@stefanprodan did you envision reconcile would be for one object, or a whole list?

stefanprodan:
> did you envision reconcile would be for one object, or a whole list?

I would say per object, for example repo/team/cred reconciliation could be bundled in tk by calling into each object reconcile func.

docs/proposals/go-git-providers.md
luxas: @dinosk expressed some interest in automatically using different backends for parsed clone/repo URLs. This is what the `MultiClient` is doing. I was thinking of it to be dead simple, in that it just checks at the domain of the Git URL, and uses the right client for the right URL. However, `MultiClient` is purely additive, it doesn't affect the rest of the proposal. I thought it might save some typing.

If you strongly feel against it, lmk and I can remove it, at least for the first iteration.

squaremo:
> with a new repo & library

One last thing: I don't think this needs to have its own repository. There's a place for reusable bits of library, that's in `fluxcd/pkg`.

docs/proposals/go-git-providers.md
squaremo: Oh, it's a _factory_ for clients. I have no strong objection to code that examines the supplied URL and decides which implementation to use. Might be expedient to leave it for now though.

squaremo:
> I don't think this needs to have its own repository. There's a place for reusable bits of library, that's in `fluxcd/pkg`.

Repositories are not a scarce resource, but they do come with a certain amount of overhead, so we need to be deliberate about creating new ones. There are _some_ reasons it might be sensible to create a new one:

 - a) it's technically difficult to use this library unless it's in its own repo (not the case, but probably it would need its own go.mod if it were under pkg/)
 - b) it drags in dependencies that make life difficult for other packages (not the case, and also not a problem if it were its own module in pkg/)
 - c) it will have a distinct set of committers and contributors

Of these, I think only c) really applies here. The code itself would be completely fine in pkg/ -- but it will be easier to manage permissions and what-not if it lives in a separate repository.

luxas:
Yes, let's move forward with the separate repo as per the arguments you listed. Thanks.

I think this is ready to be merged, and then I'll populate the repo with your help, and then we can start moving code over :smile:

stefanprodan:
@luxas I've create the repo here https://github.com/fluxcd/go-git-providers and I've invited you to join the project as maintainer.

luxas:
Thanks!

PR #108, opened 2020-07-17
Link to image update discussion

squaremo:
.. and rearrange the line items a little. I removed the mention of source-controller because it doesn't feature in the design as drafted.

### DIFF
--- a/docs/roadmap/index.md
+++ b/docs/roadmap/index.md
@@ -41,7 +41,7 @@ Tasks
 
 Goals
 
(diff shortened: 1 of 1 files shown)


hiddeco:
Can you please add two spaces to every line of my to my not-so-nicely rendering sublist? That would safe me from a(nother) PR.

squaremo:
> Can you please add two spaces to every line of my to my not-so-nicely rendering sublist?

Which bit is the sublist?

hiddeco:
L79-L81 in your revision.

PR #128, opened 2020-07-23
Add container registry and image pull secret as install options

stefanprodan:
Changes to install and bootstrap commands:
- Add container registry optional arg 
- Add image pull secret optional arg

Example of installing the toolkit using images from a private repository:

```sh
tk install \
--namespace=gitops-system \
--components=source-controller,kustomize-controller \
--registry=registry.internal/fluxcd \
--image-pull-secret=regcred

kubectl -n gitops-system create secret generic regcred \
    --from-file=.dockerconfigjson=/.docker/config.json \
    --type=kubernetes.io/dockerconfigjson
```

### DIFF
--- a/cmd/tk/bootstrap.go
+++ b/cmd/tk/bootstrap.go
@@ -47,6 +47,7 @@ var bootstrapCmd = &cobra.Command{
 var (
 	bootstrapVersion    string
 	bootstrapComponents []string
+	bootstrapRegistry   string
 )
 
 const (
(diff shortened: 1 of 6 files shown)


stefanprodan:
@hiddeco I don't want to allow people to change the image names or tags, they should pull our official images and push them to their repos maintaining our version, otherwise it will not be possible to track bugs to a version. It also fits with corporate image mirrors where the name:tag is preserved.

paulcarlton-ww:
LGTM

PR #131, opened 2020-07-25
Remove tk sync example from help command

mmorejon:
Remove `tk sync` example from help command


mmorejon:
> In addition, to resolve the CI failure: please run `make api-docs` after making the change and before committing the changes to align the generated docs with the updated examples.

I used `make docs`, the entry `api-docs` doesn't appear in the `Makefile`.

Issue #24, opened 2020-05-06
Add checksum verification to install script


stefanprodan:
The install script should download the checksums.txt from the release page and verify the tar.gz before extracting it.

PR #163, opened 2020-08-18
Add checksum verification to install script #24

mberwanger:
Add support for checksum verification

### DIFF
--- a/install/tk.sh
+++ b/install/tk.sh
@@ -1,51 +1,183 @@
 #!/usr/bin/env bash
-
 set -e
 
 DEFAULT_BIN_DIR="/usr/local/bin"
-BIN_DIR=${1:-"$DEFAULT_BIN_DIR"}
+BIN_DIR=${1:-"${DEFAULT_BIN_DIR}"}
+GITHUB_REPO="fluxcd/toolkit"
 
-opsys=""
-if [[ "$OSTYPE" == linux* ]]; then
-  opsys=linux
-elif [[ "$OSTYPE" == darwin* ]]; then
-  opsys=darwin
-fi
-
-if [[ "$opsys" == "" ]]; then
-  echo "OS $OSTYPE not supported"
-  exit 1
-fi
-
-if [[ ! -x "$(command -v curl)" ]]; then
-    echo "curl not found"
-    exit 1
-fi
-
-tmpDir=`mktemp -d`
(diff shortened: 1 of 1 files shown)


install/tk.sh
```diff
@@ -1,50 +1,83 @@
...
 
-curl -s https://api.github.com/repos/fluxcd/toolkit/releases/latest |\
-  grep browser_download |\
-  grep $opsys |\
-  cut -d '"' -f 4 |\
-  xargs curl -sL -o tk.tar.gz
+# Verify that checksum matches
+CHECKSUMS_URL=$(echo "${DOWNLOAD_URLS}" | grep checksums)
+EXPECTED_CHECKSUM=$(curl -sL "${CHECKSUMS_URL}" | grep "${OS}" | grep "${ARCH}" | cut -d ' ' -f 1)
+if [[ $EXPECTED_CHECKSUM != "$(sha256sum tk.tar.gz | awk '{print $1}')" ]]; then
```
stefanprodan, line 73: We have to check if `sha256sum` is installed, if not we should switch to `shasum -a 256` or error out if `shasum` is not installed.

install/tk.sh
```diff
@@ -1,50 +1,83 @@
...
+if [[ $OS = unknown ]] || [[ $ARCH = unknown ]]; then
+    echo "Unable to detect your OS and CPU architecture."
+    exit 1
+fi
+if [[ $ARCH != amd64 ]]; then
+    echo "CPU architecture $ARCH is not supported."
+    exit 1
 fi
 
 if [[ ! -x "$(command -v curl)" ]]; then
```
hiddeco, line 43: Another improvement here may be to fallback to `wget` as either `curl` or `wget` is available on most system.

install/tk.sh
mberwanger: I rewrote the script and added support for shasum fallback.

install/tk.sh
mberwanger: I rewrote the script and added support for a downloader fallback.

hiddeco (approved):
Looks great, awesome improvements @mberwanger :tulip: 

Can you please smash your commits together (with a commit description that sums up the changes) so that the history in `master` stays clean? Thanks in advance.

stefanprodan (approved):
LGTM

Thanks @mberwanger 🥇

hiddeco:
Excellent, thanks a lot (again) @mberwanger :kiwi_fruit:

PR #180, opened 2020-08-25
Rename `tk` binary to `gotk`

### LINKED ISSUE #174: `tk` binary conflicts with Tanka
The toolkit binary `tk` conflicts with the binary from Tanka (https://tanka.dev/), which is unfortunately operating in the same area and might cause trouble for quite a bit of Kubernetes users.

hiddeco:
To avoid conflicts with the `tk` binary from the Tanka project.

Fixes #174


squaremo:
I know I said I preferred this name, but Leigh mentioned that it looks like `go`(lang) tooling and now I can't unsee it ...

hiddeco:
I can see why, @stefanprodan mentioned that he isn't a big fan of how it would be pronounced (although it reads nice). I am however very short on other options that make sense for a CLI tool that drives the "GitOps Toolkit".

stefanprodan (approved):
I can't say I'm happy with this change but we should play nice and avoid conflicting with Tanka

Issue #194, opened 2020-09-02
Publish linux/amd64 and arm64 images to GitHub Container Registry


stefanprodan:
To avoid Docker Hub [pull limits](https://github.com/cncf/foundation/issues/106#issuecomment-674129151) we should move to GitHub Container Registry and publish AMD64 and ARM64 images for the toolkit controllers:

- [x] source-controller
- [x] kustomize-controller
- [x] helm-controller
- [x] notification-controller

Reference implementation:
- Dockerfile: https://github.com/fluxcd/notification-controller/blob/master/Dockerfile
- GitHub workflow based on [docker/build-push-action v2](https://github.com/docker/build-push-action/pull/92): https://github.com/fluxcd/notification-controller/blob/master/.github/workflows/release.yml

The controller images will be made public under `ghcr.io/fluxcd/`:
* linux/amd64: `ghcr.io/fluxcd/<controller>:semver`
* linux/arm64: `ghcr.io/fluxcd/<controller>-arm64:semver`

Once the controller images are available on GitHub Container Registry, we should add a command flag to gotk install/bootstrap that allows users to chose between amd64 and arm64 e.g.:
```
gotk install --arch=arm64
```

PR #204, opened 2020-09-04
Use GitHub Container Registry for AMD64/ARM64 images

stefanprodan:
Use GitHub Container Registry instead of Docker Hub for the toolkit controller images linux/amd64 and arm64.

Fix: #194


hiddeco (approved):
:rocket:

hiddeco (approved):
:rocket:

PR #216, opened 2020-09-08
Add label arg to create commands

stefanprodan:
This PR adds support for defining labels when creating or updating the toolkit resources.

Example:

```sh
gotk create source git podinfo \
--label=toolkit.fluxcd.io/env=dev \
--label=toolkit.fluxcd.io/role=app
```

### DIFF
--- a/cmd/gotk/create.go
+++ b/cmd/gotk/create.go
@@ -17,6 +17,8 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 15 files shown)


cmd/gotk/create.go
```diff
@@ -31,10 +33,25 @@ var createCmd = &cobra.Command{
 var (
 	interval time.Duration
 	export   bool
+	labels   []string
 )
 
 func init() {
 	createCmd.PersistentFlags().DurationVarP(&interval, "interval", "", time.Minute, "source sync interval")
 	createCmd.PersistentFlags().BoolVar(&export, "export", false, "export in YAML format to stdout")
+	createCmd.PersistentFlags().StringSliceVar(&labels, "label", nil, "label in the format key=value, accepts comma-separated items")
```
hiddeco, line 42: Helm uses something like the following to make the "comma-separated items" more clear to new (CLI) users:

```suggestion
	createCmd.PersistentFlags().StringSliceVar(&labels, "label", nil, "set labels on the resource (can specify multiple labels with commas: label1=value1,label2=value2)")
```

stefanprodan: Cool, updated!

cmd/gotk/create_helmrelease.go
```diff
@@ -182,6 +180,14 @@ func createHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return exportHelmRelease(helmRelease)
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
hiddeco, line 183: :100: :1st_place_medal:

hiddeco (approved):
LGTM

PR #218, opened 2020-09-09
Add GitHub bootstrap e2e test

stefanprodan:
Add e2e tests for `gotk bootstrap github` using fluxcd-testing org.

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -3,8 +3,7 @@ name: bootstrap
 on:
   push:
     branches:
-      - master
-      - refac
+      - '*'
 
 jobs:
   github:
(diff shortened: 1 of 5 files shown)


.github/workflows/bootstrap.yaml
```diff
@@ -0,0 +1,58 @@
+name: bootstrap
+
+on:
+  push:
+    branches:
+      - '*'
```
hiddeco, line 6: Do we want to run this for _any_ branch (so also WIP ones), or just `master`?

stefanprodan: Using `*` means that at least for our own PRs we'll run the bootstrap tests before merge.

hiddeco (approved):

Issue #224, opened 2020-09-11
Watch all namespaces by default


stefanprodan:
Currently the toolkit controllers have a mix behaviour: `source-controller` and `kustomize-controller` resource watcher is restricted to the runtime namespace e.g. `gitops-system` while `helm-controller` and `notification-controller` are watching all namespaces. 

We should introduce a boolean command flag called `--watch-all-namespaces` across all toolkit controllers that allows toggling the resource watcher from cluster wide to a the runtime namespace only. By default, the controllers will watch resources in all namespaces, users can change the default with `gotk bootstrap/install`.

When `--watch-all-namespaces` is enabled, the `kustomize-controller` service account impersonation can't be enforced by cluster admins, as any cluster user could create `Kustomization` objects with a cluster-admin SA. In the future we'll be designing a ReconciliationPolicy API that will allow cluster admins to enforce service accounts.

Components affected by this change:
- [x] source-controller
- [x] kustomize-controller
- [x] helm-controller
- [x] notification-controller
- [x] gotk CLI

PR #227, opened 2020-09-13
Add watch all namespaces flag

stefanprodan:
This PR adds the `--watch-all-namespaces` flag to gotk install and bootstrap commands. 

With  `--watch-all-namespaces` (defaults to `true`) users can configure the toolkit controllers to watch for custom resources in all namespaces, if set to `false` it will only watch the namespace where the toolkit is installed.

Fix: #224


hiddeco (approved):
💯

PR #235, opened 2020-09-14
Support bash, fish, zsh, and powershell shell completions as separate sub-commands

stealthybox:
Was running through the `get-started.md` and noticed we didn't have zsh support.
I use zsh, so I figured I'd add it along /w Cobra's other supported shells.

This sub-command approach is the same one used by `kind`.
It's also similar to `kubectl`, except they just use an argument instead of dedicated sub-commands.

### DIFF
--- a/cmd/gotk/completion_bash.go
+++ b/cmd/gotk/completion_bash.go
@@ -17,15 +17,28 @@ limitations under the License.
 package main
 
 import (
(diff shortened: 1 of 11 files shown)


stefanprodan (approved):
LGTM

Thanks @stealthybox

PR #237, opened 2020-09-15
Rename 'gitops-system' namespace to 'gotk-system'

hiddeco:
To align with the project name, and the group introduced in #236.

Ref: #238


stefanprodan (approved):
LGTM

Issue #266, opened 2020-09-23
Add bucket commands


stefanprodan:
The gotk CLI should handle Bucket sources:
- [x] create/get/export/reconcile/delete source bucket
- [ ] create kusomization/helmrelease --source=Bucket/name

PR #270, opened 2020-09-23
Add support for Bucket sources to Kustomizations/HRs

stefanprodan:
Fix: #266

### DIFF
--- a/cmd/gotk/create_kustomization.go
+++ b/cmd/gotk/create_kustomization.go
@@ -39,7 +39,7 @@ var createKsCmd = &cobra.Command{
 	Use:     "kustomization [name]",
 	Aliases: []string{"ks"},
 	Short:   "Create or update a Kustomization resource",
-	Long:    "The kustomization source create command generates a Kustomize resource for a given GitRepository source.",
+	Long:    "The kustomization source create command generates a Kustomize resource for a given source.",
 	Example: `  # Create a Kustomization resource from a source at a given path
   gotk create kustomization contour \
     --source=contour \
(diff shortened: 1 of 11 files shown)


cmd/gotk/create_kustomization.go
```diff
@@ -111,6 +108,18 @@ func createKsCmdRun(cmd *cobra.Command, args []string) error {
 	if ksSource == "" {
 		return fmt.Errorf("source is required")
 	}
+
+	ksSourceKind := sourcev1.GitRepositoryKind
+	ksSourceName := ksSource
+	ksSourceElements := strings.Split(ksSource, "/")
```
hiddeco, line 114: We are doing this in so many places now that I think it could be added to the utils.

cmd/gotk/create_kustomization.go
```diff
@@ -88,7 +84,8 @@ var (
 )
 
 func init() {
-	createKsCmd.Flags().StringVar(&ksSource, "source", "", "GitRepository name")
+	createKsCmd.Flags().StringVar(&ksSource, "source", "",
+		"source that contains the Kubernetes manifests, format '<kind>/<name>' where kind can be GitRepository or Bucket, if kind is not specified it defaults to GitRepository")
```
hiddeco, line 88:
```suggestion
		"source that contains the Kubernetes manifests in format '[<kind>/]<name>', where kind can be GitRepository or Bucket, if kind is not specified it defaults to GitRepository")
```

hiddeco (approved):
:100:

PR #273, opened 2020-09-26
Add support for multi-namespace tenant ownership

stefanprodan:
Changes:
- add `--with-namespace` arg to create tenant command
- label namespaces and role bindings with the tenant name for discovery purposes
- allow changing the the cluster role name with `--cluster-role` arg, defaults to `cluster-admin` 
- add create tenant examples to docs

Ref: https://github.com/fluxcd/toolkit/discussions/263


hiddeco (approved):
LGTM

PR #276, opened 2020-09-28
Generate a kustomize package for the bootstrap manifests

stefanprodan:
This PR generates a kustomization.yaml file in the bootstrap target path, this enables further customisations of the bootstrap manifests.

### DIFF
--- a/cmd/gotk/utils.go
+++ b/cmd/gotk/utils.go
@@ -22,8 +22,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/gotk/utils.go
```diff
@@ -210,3 +217,81 @@ func (*Utils) makeDependsOn(deps []string) []dependency.CrossNamespaceDependency
 	}
 	return refs
 }
+
+func (*Utils) makeKustomizationYaml(dirPath string) error {
```
hiddeco, line 221: Nitpick, but I feel like `build` or `generate` instead of `make` is a better description of what the method does.

hiddeco (approved):

Issue #238, opened 2020-09-15
Preparation for v0.1.0 release


stefanprodan:
After working for six months on the GitOps Toolkit we've reached feature parity with Flux v1 read-only and Helm Operator v1. I think it's time to make a v0.1 release and promote the APIs to beta. Version 0.1.x will include breaking changes to v0.0.x, so until we release v0.1 we should freeze gotk to 0.0.28 and bundle all changes into 0.1.

Pre 0.1.0:
- [x] rename the default namespace from `gitops-system` to `gotk-system`
- [x] change `DependsOn` type from `[]string` to `[]CrossNamespaceDependencyReference`
- [x] incorporate `DependsOn` changes in `gotk create [...]` commands
- [x] replace `Kustomization.Status.Snapshot.Revision` with `Checksum` in Kustomization API
- [x] add support for CAPI remote clusters in Kustomnization API
- [x] add support for S3 Bucket sources
- [x] publish multi-arch ARMv7/v8 container images
- [x] promote source/kustomize/helm/notification APIs from `alpha1` to `beta1`
- [x] update Go to v1.15
- [x] make controller releases

Post 0.1.0:
- [ ] remove `Experimental` from the controllers descriptions and docs
- [x] extend documentation around managing Helm releases
- [ ] create a migration guide for Flux v1 read-only users
- [ ] create a migration guide for Helm Operator v1 users
- [ ] update notices in `fluxcd/flux` and `fluxcd/helm-operator` `README.md`s
- [ ] highlight the existence of the project on the fluxcd.io website

squaremo:
> [x] implement create tenant command

I don't recall tenants being a first-class thing, is there a design for it somewhere?

stefanprodan:
The tenant command is for creating a role binding inside a namespace that belongs to a tenant. This doesn't mean we have a Tenant API/CRD, it's more about providing a way to streamline the creation of restricted roles for the GOTK reconcilers. I've considered naming it `gotk create tenant-namespace-role` but that's so long...

squaremo:
I think discussion #178 needs to be resolved before releasing anything that's held up as having parity with Flux v1 (and before writing migration guides and so on.)

stefanprodan:
v0.1.0 is out https://github.com/fluxcd/toolkit/releases/tag/v0.1.0 🎉 

I propose that we close this and take the post v0.1 tasks and make them into separate issues.

PR #280, opened 2020-09-30
Update components to v0.1

stefanprodan:
Update api packages and controllers to v0.1.0.
This version is not backwards compatible with v0.0.28.

Ref: #238


cmd/gotk/create_source_git.go
```diff
@@ -373,7 +374,7 @@ func isGitRepositoryReady(ctx context.Context, kubeClient client.Client, name, n
 		}
 
 		for _, condition := range gitRepository.Status.Conditions {
```
hiddeco, line 376:
```suggestion
if c := meta.GetCondition(gitRepository.Status.Conditions, meta.ReadyCondition); c != nil {
	switch condition.Status {
	case corev1.ConditionTrue:
		return true, nil
	case corev1.ConditionFalse:
		return false, fmt.Errorf(c.Message)
	}
}
```

cmd/gotk/get_helmrelease.go
```diff
@@ -68,7 +69,7 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		}
 		isInitialized := false
 		for _, condition := range helmRelease.Status.Conditions {
-			if condition.Type == helmv2.ReadyCondition {
+			if condition.Type == meta.ReadyCondition {
```
hiddeco, line 72: This could also use `meta.GetCondition()` but looking at the logic in general, it could use an overall update to be more accurate.

cmd/gotk/reconcile_helmrelease.go
```diff
@@ -144,7 +144,7 @@ func isHelmReleaseReady(ctx context.Context, kubeClient client.Client, name, nam
 		}
 
 		for _, condition := range helmRelease.Status.Conditions {
-			if condition.Type == helmv2.ReadyCondition {
+			if condition.Type == meta.ReadyCondition {
```
hiddeco, line 147: This should take the new `LastHandledReconcilateAt` status field into account, and could make use of `meta.GetCondition()`.

cmd/gotk/reconcile_kustomization.go
```diff
@@ -94,10 +94,10 @@ func reconcileKsCmdRun(cmd *cobra.Command, args []string) error {
 		logger.Actionf("annotating kustomization %s in %s namespace", name, namespace)
 		if kustomization.Annotations == nil {
 			kustomization.Annotations = map[string]string{
-				consts.ReconcileAtAnnotation: time.Now().Format(time.RFC3339Nano),
+				meta.ReconcileAtAnnotation: time.Now().Format(time.RFC3339Nano),
 			}
 		} else {
-			kustomization.Annotations[consts.ReconcileAtAnnotation] = time.Now().Format(time.RFC3339Nano)
+			kustomization.Annotations[meta.ReconcileAtAnnotation] = time.Now().Format(time.RFC3339Nano)
```
hiddeco, line 100: Same for the logic below, although the current is less problematic given how the kustomize-controller handles state. It could however take the new `LastHandledReconcilateAt` status field into account, and also make use of `meta.GetCondition()`.

hiddeco (approved):
Couple of other places where the `meta` API package would likely improve the overall logic, and/or added status fields would improve the behaviour.

But the switch to the API itself looks OK.

PR #281, opened 2020-09-30
Make use of `apis/meta` helpers and improve `reconcile` logic

hiddeco:
Fixes https://github.com/fluxcd/toolkit/issues/215 based on discussion in https://github.com/fluxcd/toolkit/discussions/217#discussioncomment-86404


stefanprodan (approved):
LGTM

Nice refactoring 🥇

Issue #294, opened 2020-10-02
Network Policy issue with deploying gotk with k3s


onedr0p:
I recently chatted with @stefanprodan and he came to the conclusion that k3s might behave differently than k8s in terms of network policies.

k3s version: `1.19.2-k3s1`

Basically the bootstrap process times out.
```
◎ waiting for cluster sync
✗ context deadline exceeded
```

I was able to get this to work with Calico and kube-proxy by removing the gotk network policy `kubectl delete networkpolicy/deny-ingress -n gotk-system`

However it didn't work on my tests when kube-proxy was disabled when using Calico w/ eBPF or Cilium and kube-router.

More information may be be found [here](https://cloud-native.slack.com/archives/CLAJ40HV3/p1601658217127500) in the slack channel.

Related: https://github.com/rancher/k3s/issues/1308 and https://github.com/rancher/k3s/issues/947 which also links to a kube-router issue.

onedr0p:
@stefanprodan awesome! I'll give this a shot and report back if I see any issues.

stefanprodan:
You need to use the v0.1.4 CLI, it should be available in a couple of minutes

onedr0p:
Seems like my issue was Cilium and setting `masquerade` to `false` in their helm chart values. I don't know why this would have an effect on gotk completing the bootstrap process or not. I have turned back on masquerading and gotk finishes the bootstrap command.

https://docs.cilium.io/en/v1.8/concepts/networking/masquerading/

PR #297, opened 2020-10-03
Add option to disable the network policy at install time

stefanprodan:
This PR adds the `--network-policy` boolean flag (defaults to `true`) to the install and bootstrap commands. This allows users to disable the network policy that denies ingress access to the toolkit controllers from other namespaces. Disabling the network policy is not advised especially when running on a multi-tenancy cluster.

Fix: #294


hiddeco (approved):
LGTM

PR #299, opened 2020-10-03
Switch get commands to use tables for output

### LINKED ISSUE #292: Output tables for `gotk get` commands
The current `gotk get` commands write simple lines to `stdout`.Which is not very readable, and does not provide a great overview of the status the resources are in.

Instead, we should write the output to tables which (at least) should include:

* Name
* Current revision
* Ready status (`True`, `False`, `Unknown`)
* Ready message

circa10a:
Fix #292 

Happy Hacktoberfest! 🎃 

Signed-off-by: circa10a <caleblemoine@gmail.com>

### DIFF
--- a/cmd/gotk/get_helmrelease.go
+++ b/cmd/gotk/get_helmrelease.go
@@ -18,6 +18,10 @@ package main
 
 import (
 	"context"
+	"os"
+	"strconv"
+	"strings"
+
 	"github.com/fluxcd/pkg/apis/meta"
 
 	"github.com/spf13/cobra"
@@ -51,8 +55,12 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var listOpts []client.ListOption
+	if !allNamespaces {
+		listOpts = append(listOpts, client.InNamespace(namespace))
+	}
 	var list helmv2.HelmReleaseList
-	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
+	err = kubeClient.List(ctx, &list, listOpts...)
 	if err != nil {
 		return err
 	}
@@ -62,26 +70,35 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	for _, helmRelease := range list.Items {
-		if helmRelease.Spec.Suspend {
-			logger.Successf("%s is suspended", helmRelease.GetName())
-			continue
-		}
-		isInitialized := false
-		if c := meta.GetCondition(helmRelease.Status.Conditions, meta.ReadyCondition); c != nil {
-			switch c.Status {
-			case corev1.ConditionTrue:
-				logger.Successf("%s last applied revision %s", helmRelease.GetName(), helmRelease.Status.LastAppliedRevision)
-			case corev1.ConditionUnknown:
-				logger.Successf("%s reconciling", helmRelease.GetName())
-			default:
-				logger.Failuref("%s %s", helmRelease.GetName(), c.Message)
-			}
-			isInitialized = true
-		}
-		if !isInitialized {
-			logger.Failuref("%s is not ready", helmRelease.GetName())
-		}
+	header := []string{"Name", "Revision", "Suspended", "Ready", "Message"}
+	if allNamespaces {
+		header = append([]string{"Namespace"}, header...)
 	}
+	var rows [][]string
+	for _, helmRelease := range list.Items {
+		row := []string{}
+		if c := meta.GetCondition(helmRelease.Status.Conditions, meta.ReadyCondition); c != nil {
+			row = []string{
+				helmRelease.GetName(),
+				helmRelease.Status.LastAppliedRevision,
+				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
+				string(c.Status),
+				c.Message,
+			}
+		} else {
+			row = []string{
+				helmRelease.GetName(),
+				helmRelease.Status.LastAppliedRevision,
+				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
+				string(corev1.ConditionFalse),
+				"waiting to be reconciled",
+			}
+		}
+		if allNamespaces {
+			row = append([]string{helmRelease.Namespace}, row...)
+		}
+		rows = append(rows, row)
+	}
+	utils.printTable(os.Stdout, header, rows)
 	return nil
 }
--- a/cmd/gotk/get_kustomization.go
+++ b/cmd/gotk/get_kustomization.go
@@ -18,6 +18,10 @@ package main
 
 import (
 	"context"
+	"os"
+	"strconv"
+	"strings"
+
 	"github.com/fluxcd/pkg/apis/meta"
 
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
@@ -50,8 +54,12 @@ func getKsCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var listOpts []client.ListOption
+	if !allNamespaces {
+		listOpts = append(listOpts, client.InNamespace(namespace))
+	}
 	var list kustomizev1.KustomizationList
-	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
+	err = kubeClient.List(ctx, &list, listOpts...)
 	if err != nil {
 		return err
 	}
@@ -61,26 +69,35 @@ func getKsCmdRun(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	for _, kustomization := range list.Items {
-		if kustomization.Spec.Suspend {
-			logger.Successf("%s is suspended", kustomization.GetName())
-			continue
-		}
-		isInitialized := false
-		if c := meta.GetCondition(kustomization.Status.Conditions, meta.ReadyCondition); c != nil {
-			switch c.Status {
-			case corev1.ConditionTrue:
-				logger.Successf("%s last applied revision %s", kustomization.GetName(), kustomization.Status.LastAppliedRevision)
-			case corev1.ConditionUnknown:
-				logger.Successf("%s reconciling", kustomization.GetName())
-			default:
-				logger.Failuref("%s %s", kustomization.GetName(), c.Message)
-			}
-			isInitialized = true
-		}
-		if !isInitialized {
-			logger.Failuref("%s is not ready", kustomization.GetName())
-		}
+	header := []string{"Name", "Revision", "Suspended", "Ready", "Message"}
+	if allNamespaces {
+		header = append([]string{"Namespace"}, header...)
 	}
+	var rows [][]string
+	for _, kustomization := range list.Items {
+		row := []string{}
+		if c := meta.GetCondition(kustomization.Status.Conditions, meta.ReadyCondition); c != nil {
+			row = []string{
+				kustomization.GetName(),
+				kustomization.Status.LastAppliedRevision,
+				strings.Title(strconv.FormatBool(kustomization.Spec.Suspend)),
+				string(c.Status),
+				c.Message,
+			}
+		} else {
+			row = []string{
+				kustomization.GetName(),
+				kustomization.Status.LastAppliedRevision,
+				strings.Title(strconv.FormatBool(kustomization.Spec.Suspend)),
+				string(corev1.ConditionFalse),
+				"waiting to be reconciled",
+			}
+		}
+		if allNamespaces {
+			row = append([]string{kustomization.Namespace}, row...)
+		}
+		rows = append(rows, row)
+	}
+	utils.printTable(os.Stdout, header, rows)
 	return nil
 }
--- a/cmd/gotk/get_source_bucket.go
+++ b/cmd/gotk/get_source_bucket.go
@@ -18,6 +18,8 @@ package main
 
 import (
 	"context"
+	"os"
+
 	"github.com/fluxcd/pkg/apis/meta"
 
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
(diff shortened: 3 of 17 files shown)


circa10a:
Strange, CI failed on:

```
0s
Run ./bin/gotk get sources git
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2029b2f]

goroutine 1 [running]:
main.getSourceGitCmdRun(0x36f2a40, 0x37d7930, 0x0, 0x0, 0x0, 0x0)
	/home/runner/work/toolkit/toolkit/cmd/gotk/get_source_git.go:78 +0x75f
```

But I can't reproduce:

```bash
~/Desktop/code/toolkit main
❯ ./bin/gotk get sources git
+------+----------+---------+------------------+-------------+
| NAME | REVISION | STATUS  |     MESSAGE      |  NAMESPACE  |
+------+----------+---------+------------------+-------------+
| None | None     | Unknown | No sources found | gotk-system |
+------+----------+---------+------------------+-------------+

~/Desktop/code/toolkit main
❯ ./bin/gotk get sources git
+---------+-------------------------------------------------+--------+---------+-------------+
|  NAME   |                    REVISION                     | STATUS | MESSAGE |  NAMESPACE  |
+---------+-------------------------------------------------+--------+---------+-------------+
| podinfo | master/7411da595c25183daba255068814b83843fe3395 | true   | ready   | gotk-system |
+---------+-------------------------------------------------+--------+---------+-------------+
```

cmd/gotk/get_kustomization.go
```diff
@@ -50,37 +53,44 @@ func getKsCmdRun(cmd *cobra.Command, args []string) error {
...
 	var list kustomizev1.KustomizationList
 	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
 	if err != nil {
 		return err
 	}
 
 	if len(list.Items) == 0 {
-		logger.Failuref("no kustomizations found in %s namespace", namespace)
+		table.Append([]string{"None", "None", "Unknown", "No kustomizations found", namespace})
+		table.Render()
```
hiddeco, line 68: If there are no kustomizations found, this should not print a table but simply log a string to `stderr` (as it did before the change).

cmd/gotk/get_source_bucket.go
```diff
@@ -49,36 +52,39 @@ func getSourceBucketCmdRun(cmd *cobra.Command, args []string) error {
...
 	var list sourcev1.BucketList
 	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
 	if err != nil {
 		return err
 	}
 
 	if len(list.Items) == 0 {
-		logger.Failuref("no sources found in %s namespace", namespace)
+		table.Append([]string{"None", "None", "Unknown", "No sources found", namespace})
+		table.Render()
```
hiddeco, line 67: If there are no sources found, this should not print a table but simply log a string to `stderr` (as it did before the change).

The printed string could be more informative about the resource type, e.g. `no bucket sources found in %s namespace`.

cmd/gotk/get_source_git.go
```diff
@@ -49,36 +52,39 @@ func getSourceGitCmdRun(cmd *cobra.Command, args []string) error {
...
 	var list sourcev1.GitRepositoryList
 	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
 	if err != nil {
 		return err
 	}
 
 	if len(list.Items) == 0 {
-		logger.Failuref("no sources found in %s namespace", namespace)
+		table.Append([]string{"None", "None", "Unknown", "No sources found", namespace})
+		table.Render()
```
hiddeco, line 67: If there are no sources found, this should not print a table but simply log a string to `stderr` (as it did before the change).

The printed string could be more informative about the resource type, e.g. `no git sources found in %s namespace`.

cmd/gotk/get_source_helm.go
```diff
@@ -49,36 +52,39 @@ func getSourceHelmCmdRun(cmd *cobra.Command, args []string) error {
...
 	var list sourcev1.HelmRepositoryList
 	err = kubeClient.List(ctx, &list, client.InNamespace(namespace))
 	if err != nil {
 		return err
 	}
 
 	if len(list.Items) == 0 {
-		logger.Failuref("no sources found in %s namespace", namespace)
+		table.Append([]string{"None", "None", "Unknown", "No sources found", namespace})
+		table.Render()
```
hiddeco, line 67: If there are no sources found, this should not print a table but simply log a string to `stderr` (as it did before the change).

The printed string could be more informative about the resource type, e.g. `no helm sources found in %s namespace`.

cmd/gotk/get_kustomization.go
```diff
@@ -50,37 +53,44 @@ func getKsCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "Revision", "Ready Status", "Message", "Namespace"})
+	table.SetRowLine(true)
```
hiddeco, line 58: With the below change, this can be moved above the loop that goes through the `list.Items`.

cmd/gotk/get_source_bucket.go
```diff
@@ -49,36 +52,39 @@ func getSourceBucketCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "Revision", "Ready Status", "Message", "Namespace"})
+	table.SetRowLine(true)
```
hiddeco, line 57: With the below change, this can be moved above the loop that goes through the `list.Items`.

cmd/gotk/get_source_git.go
```diff
@@ -49,36 +52,39 @@ func getSourceGitCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "Revision", "Ready Status", "Message", "Namespace"})
+	table.SetRowLine(true)
```
hiddeco, line 57: With the below change, this can be moved above the loop that goes through the `list.Items`.

cmd/gotk/get_helmrelease.go
```diff
@@ -51,37 +54,43 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "Revision", "Ready Status", "Message", "Namespace"})
+	table.SetRowLine(true)
```
hiddeco, line 59: With the below change, this can be moved above the loop that goes through the `list.Items`.

cmd/gotk/get_helmrelease.go
```diff
@@ -51,37 +54,43 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "Revision", "Ready Status", "Message", "Namespace"})
```
hiddeco, line 58:
```suggestion
	table.SetHeader([]string{"Name", "Revision", "Ready", "Message", "Suspended"})
```

Given the user already specifies a mandatory namespace selector, as I do not think that displaying the namespace offers any additional value (`kubectl` as an example behaves the same, but _does_ display the name if the `--all-namespaces` flag is passed).

cmd/gotk/get_helmrelease.go
```diff
@@ -51,37 +54,43 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
...
 			default:
-				logger.Failuref("%s %s", helmRelease.GetName(), c.Message)
+				table.Append([]string{helmRelease.GetName(), helmRelease.Status.LastAppliedRevision, "false", c.Message, namespace})
 			}
 			isInitialized = true
 		}
 		if !isInitialized {
-			logger.Failuref("%s is not ready", helmRelease.GetName())
+			table.Append([]string{helmRelease.GetName(), helmRelease.Status.LastAppliedRevision, "false", "not ready", namespace})
 		}
```
hiddeco, line 92: I think this can be reduced to (this should be applicable to all other `get` commands, with some minor variable name changes):

```suggestion
		if c := meta.GetCondition(helmRelease.Status.Conditions, meta.ReadyCondition); c != nil {
			table.Append([]string{
				helmRelease.GetNamespace(),
				helmRelease.Status.LastAppliedRevision,
				string(c.Status),
				c.Message,
				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
			})
		} else {
			table.Append([]string{
				helmRelease.GetName(),
				helmRelease.Status.LastAppliedRevision,
				"False",
				"waiting to be reconciled",
				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
			})
		}
```

hiddeco (commented):
Thank you very much for picking this up @circa10a :1st_place_medal:, please have a look at my (minor) suggestions. :jack_o_lantern:

cmd/gotk/get_helmrelease.go
circa10a: Did you mean to have `helmRelease.GetNamespace()` and `helmRelease.GetName()` differ in the if/else here?

cmd/gotk/get_helmrelease.go
circa10a: Just to clarify, you're thinking to only remove the `Namespace` column in `get_helmrelease.go` correct? or remove the `Namespace` column in every `get` command?

hiddeco: Sorry if my comment was vague, I meant:

> remove the `Namespace` column in every `get` command?

When the namespace flag is passed, it should however show for every `get`  command, hope this makes it more clear.

cmd/gotk/get_helmrelease.go
hiddeco: I indeed made a typo, good catch :+1:

cmd/gotk/get_helmrelease.go
circa10a: Gotcha, thoughts on the new changes?

cmd/gotk/get_helmrelease.go
```diff
@@ -62,26 +73,52 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	tableRowHeaders := []string{"Name", "Revision", "Ready", "Message", "Suspended"}
+	if allNamespaces {
+		tableRowHeaders = append(tableRowHeaders, "Namespace")
+	}
+	table.SetHeader(tableRowHeaders)
+	table.SetRowLine(true)
```
hiddeco, line 82: Had a look at the available `tablewriter` configuration options, and the following is what matches the output of `kubectl`:

```go
	table.SetAutoWrapText(false)
	table.SetAutoFormatHeaders(true)
	table.SetHeaderAlignment(tablewriter.ALIGN_LEFT)
	table.SetAlignment(tablewriter.ALIGN_LEFT)
	table.SetCenterSeparator("")
	table.SetColumnSeparator("")
	table.SetRowSeparator("")
	table.SetHeaderLine(false)
	table.SetBorder(false)
	table.SetTablePadding("\t")
	table.SetNoWhiteSpace(true)
```

This renders:

```console
$ ./bin/gotk get helmreleases -n helm-system
NAME    REVISION        READY   MESSAGE                                                 SUSPENDED 
podinfo                 False   Helm test failed: timed out waiting for the condition   False
```

---

You probably want to factor this out to `utils.go` to prevent repetitive code. E.g:

- Prepare just the configuration: `func (*Utils) newTableWriter(writer io.Writer) *tablewriter.Table`)
- Do everything in one run (likely the better option): `func (*Utils) printTable(writer io.Writer, header []string, rows [][]string)`. The contents of the function would look something like the following (pseudocode):

  ```go
    table := tablewriter.NewWriter(writer)
	table.SetHeader(header)
	// configure options from above
	table.AppendBulk(rows)
	table.Render()
  ```

  In the commands, the code would then become something like this (pseudocode):

  ```go
  header := []string{"Name", ...}
  var rows [][]string
  for _, i := range list.Items {
	row := []string{
    	i.GetName(), ...
	}
    rows = append(rows, row)
  }
  utils.printTable(io.Stdout, header, rows)
  ```

cmd/gotk/get_helmrelease.go
```diff
@@ -62,26 +73,52 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	table := tablewriter.NewWriter(os.Stdout)
+	tableRowHeaders := []string{"Name", "Revision", "Ready", "Message", "Suspended"}
+	if allNamespaces {
+		tableRowHeaders = append(tableRowHeaders, "Namespace")
+	}
```
hiddeco, line 80:
```suggestion
	if allNamespaces {
		tableRowHeaders = append([]string{"Namespace"}, tableRowHeaders...)
	}
```

Had another look at `kubectl`, and it does add the namespace as the first column, this does just that. My apologies for not checking this beforehand.

cmd/gotk/get_helmrelease.go
```diff
@@ -62,26 +73,52 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
...
+				string(corev1.ConditionFalse),
+				"waiting to be reconciled",
+				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
 			}
-			isInitialized = true
 		}
-		if !isInitialized {
-			logger.Failuref("%s is not ready", helmRelease.GetName())
+		if allNamespaces {
+			row = append(row, helmRelease.Namespace)
```
hiddeco, line 118:
```suggestion
			row = append([]string{helmRelease.Namespace, row...})
```

Companion to the above column change.

cmd/gotk/get_helmrelease.go
```diff
@@ -62,26 +73,52 @@ func getHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
...
+		} else {
+			row = []string{
+				helmRelease.GetName(),
+				helmRelease.Status.LastAppliedRevision,
+				string(corev1.ConditionFalse),
+				"waiting to be reconciled",
+				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
 			}
-			isInitialized = true
 		}
```
hiddeco, line 116: This can theoretically be changed to:

```suggestion
		row := []string{
			helmRelease.GetName(),
			helmRelease.Status.LastAppliedRevision,
			string(corev1.ConditionFalse),
			"waiting to be reconciled",
			strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
		}
		if c := meta.GetCondition(helmRelease.Status.Conditions, meta.ReadyCondition); c != nil {
			row = []string{
				helmRelease.GetName(),
				helmRelease.Status.LastAppliedRevision,
				string(c.Status),
				c.Message,
				strings.Title(strconv.FormatBool(helmRelease.Spec.Suspend)),
			}
		}
```

hiddeco (commented):
I reviewed the `get helmreleases` command, but the review comments apply to all others. Open to discussions or questions if anything I commented does not make sense :sun_with_face:

circa10a:
> I reviewed the `get helmreleases` command, but the review comments apply to all others. Open to discussions or questions if anything I commented does not make sense 🌞

All valuable feedback, I appreciate it! I really like following the `kubectl` output format. I pushed all the suggested changes.

hiddeco:
@circa10a awesome, output looks great! :1st_place_medal: 

There is one last thing I am having doubts about: the placement of the `Suspend` column. Given the `Message` column can become lengthy, I think that from an UX perspective we may want to place it between `Revision` and `Ready` (as the latter has a 1-1 relationship with the `Message`). What do you think?

circa10a:
@hiddeco Completely agree. Updated along with docs generated from `make test`.

hiddeco (approved):
Looks good to me, thanks for being patient with my review comments :sunflower: :100: 

One last request, can you please smash your commits together into one (`git rebase -i main`), and reword your first commit into something that does not refer to the issue?

This keeps the history in `main` clean from "review commits", and prevents endless pings from GitHub when the commit is moved around through e.g. different branches, as it is a bit too enthusiastic with those.

circa10a:
No no no, thank you for your patience and guidance since this was my contribution to this repository. Sincerely appreciate it

PR #305, opened 2020-10-05
Update guide to include column output commands

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco 🎁

PR #307, opened 2020-10-05
Display proper revision for sources

hiddeco:
Includes a change to an empty revision string if the reconciler has not
produced an artifact yet, as this will otherwise result in a nil
pointer dereference.

Follow up on #299


stefanprodan (approved):
LGTM

PR #315, opened 2020-10-07
Update api versions in guides

rieger-jared:


stefanprodan (approved):
LGTM

Thanks @rieger-jared

PR #318, opened 2020-10-08
Introduce install package

stefanprodan:
Changes:
- extract manifests generation to a dedicated package
- refactor the install and bootstrap command

The `install` package is meant as a replacement for Flux v1 `install` package.


hiddeco (approved):
Much better :1st_place_medal:

Issue #47, opened 2020-06-18
Windows support


stefanprodan:
The toolkit CLI [shells-out](https://github.com/fluxcd/toolkit/blob/master/cmd/tk/utils.go#L32) to kubectl using `sh -c` as a wrapper, we should create a kubectl helper that works on Windows/macOS/Linux, similar to [kustomizer](https://github.com/stefanprodan/kustomizer/blob/master/pkg/engine/kubectl.go).

PR #334, opened 2020-10-14
Get rid of shell-out and invoke command directly via exec

StupidScience:
Hello.

Please let me know if this is something you expect here.
Or would be better to switch `execCommand` to something like `execCommand(ctx context.Context, mode ExecMode, commandName string, args ...string)`?

Related to #47

### DIFF
--- a/cmd/gotk/utils.go
+++ b/cmd/gotk/utils.go
@@ -60,9 +60,10 @@ const (
 	ModeCapture  ExecMode = "capture.stderr|stdout"
 )
 
-func (*Utils) execCommand(ctx context.Context, mode ExecMode, command string) (string, error) {
+func (*Utils) execKubectlCommand(ctx context.Context, mode ExecMode, args ...string) (string, error) {
 	var stdoutBuf, stderrBuf bytes.Buffer
-	c := exec.CommandContext(ctx, "/bin/sh", "-c", command)
+
+	c := exec.CommandContext(ctx, "kubectl", args...)
 
 	if mode == ModeStderrOS {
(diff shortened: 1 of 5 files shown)


cmd/gotk/utils.go
```diff
@@ -62,7 +62,12 @@ const (
 
 func (*Utils) execCommand(ctx context.Context, mode ExecMode, command string) (string, error) {
```
stefanprodan, line 63: Can you please change the signature to:
```go
 execKubectlCommand(ctx context.Context, mode ExecMode, arg ...string) (string, error)
```

stefanprodan: After the signature change we need to pass an args array, so the current string concatenation should be rewrite to array append.

StupidScience: I've changed it

stefanprodan: Can you please rebase with main.

StupidScience: done

StupidScience: Should I squash it or smth?

stefanprodan: Yes please squash

stefanprodan:
@StupidScience can you give this a try on Windows and see if it works? Thanks

StupidScience:
Squashed. Also I tried to run `gotk check -pre` on Win10 and it successfully checked kubectl version. @stefanprodan

stefanprodan (approved):
LGTM

Thanks @StupidScience 🎖️

PR #350, opened 2020-10-16
Add secret-ref flag to create source git

allymparker:
## What? 
Add secret-ref flag to `create source` options

## Why?
This allows passing the name of an existing secret that already exists in the cluster (perhaps created with terraform)

## Tested?
`make test` is passing
Manually tested with `--export`

### DIFF
--- a/cmd/gotk/create_source_git.go
+++ b/cmd/gotk/create_source_git.go
@@ -87,15 +87,17 @@ For private Git repositories, the basic authentication credentials are stored in
 }
 
 var (
-	sourceGitURL          string
-	sourceGitBranch       string
-	sourceGitTag          string
-	sourceGitSemver       string
-	sourceGitUsername     string
-	sourceGitPassword     string
+	sourceGitURL      string
+	sourceGitBranch   string
+	sourceGitTag      string
+	sourceGitSemver   string
+	sourceGitUsername string
+	sourceGitPassword string
+
 	sourceGitKeyAlgorithm flags.PublicKeyAlgorithm = "rsa"
 	sourceGitRSABits      flags.RSAKeyBits         = 2048
 	sourceGitECDSACurve                            = flags.ECDSACurve{Curve: elliptic.P384()}
+	sourceGitSecretRef    string
 )
 
 func init() {
@@ -108,6 +110,7 @@ func init() {
 	createSourceGitCmd.Flags().Var(&sourceGitKeyAlgorithm, "ssh-key-algorithm", sourceGitKeyAlgorithm.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitRSABits, "ssh-rsa-bits", sourceGitRSABits.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitECDSACurve, "ssh-ecdsa-curve", sourceGitECDSACurve.Description())
+	createSourceGitCmd.Flags().StringVarP(&sourceGitSecretRef, "secret-ref", "", "", "the name of an existing secret containing SSH or basic credentials")
 
 	createSourceCmd.AddCommand(createSourceGitCmd)
 }
@@ -162,6 +165,11 @@ func createSourceGitCmdRun(cmd *cobra.Command, args []string) error {
 	}
 
 	if export {
+		if sourceGitSecretRef != "" {
+			gitRepository.Spec.SecretRef = &corev1.LocalObjectReference{
+				Name: sourceGitSecretRef,
+			}
+		}
 		return exportGit(gitRepository)
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/gotk/create_source_git.go
```diff
@@ -237,8 +244,13 @@ func createSourceGitCmdRun(cmd *cobra.Command, args []string) error {
 	logger.Generatef("generating source")
 
 	if withAuth {
+		secretName := name
+		if sourceGitSecretRef != "" {
+			secretName = sourceGitSecretRef
+		}
```
hiddeco, line 250: Instead of overwriting the generated secret here, I would expect the command to not generate a and upsert a secret at all if an existing secret name is given.

allymparker: So drop scenario 2 above?  Can do if that's what you prefer!

hiddeco: It does not do scenario 2 at the moment, as on L234 the secret that is upserted does not take the configured value into account.

allymparker: 🤦 Oops missed it, sorry.  Gimme a mo

cmd/gotk/create_source_git.go
```diff
@@ -105,6 +106,7 @@ func init() {
 	createSourceGitCmd.Flags().Var(&sourceGitKeyAlgorithm, "ssh-key-algorithm", sourceGitKeyAlgorithm.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitRSABits, "ssh-rsa-bits", sourceGitRSABits.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitECDSACurve, "ssh-ecdsa-curve", sourceGitECDSACurve.Description())
+	createSourceGitCmd.Flags().StringVarP(&sourceGitSecretRef, "secret-ref", "s", "", "the to use for the source secret or an existing secret containing SSH or basic credentials")
```
hiddeco, line 109:
```suggestion
	createSourceGitCmd.Flags().StringVarP(&sourceGitSecretRef, "secret-ref", "s", "", "the name to use for the source secret, or an existing secret containing SSH or basic credentials")
```

cmd/gotk/create_source_git.go
hiddeco: Adding just the name in there would not be sufficient by the way, as this would result in the overwrite of existing data.

It would require some additional logic to only create (and not update) a secret with the given name if it does not exist yet.

allymparker: What's the current behaviour if a secret with the same name as the git source exists?  Would that not have the same behavior?

Maybe I should just simplify this to just scenario 1?

allymparker: Let me know what you would prefer :)

cmd/gotk/create_source_git.go
```diff
@@ -105,6 +106,7 @@ func init() {
 	createSourceGitCmd.Flags().Var(&sourceGitKeyAlgorithm, "ssh-key-algorithm", sourceGitKeyAlgorithm.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitRSABits, "ssh-rsa-bits", sourceGitRSABits.Description())
 	createSourceGitCmd.Flags().Var(&sourceGitECDSACurve, "ssh-ecdsa-curve", sourceGitECDSACurve.Description())
+	createSourceGitCmd.Flags().StringVarP(&sourceGitSecretRef, "secret-ref", "s", "", "the name to use for the source secret, or an existing secret containing SSH or basic credentials")
```
stefanprodan, line 109: Please remove the `s` shorthand, this conflict with `--silent`.

stefanprodan (changes requested):

cmd/gotk/create_source_git.go
stefanprodan: Please drop option 2. I see no reason why would you want to setup a name for the generated secret. When `secret-ref` is specified, then we assume it's up to the user to create it on the cluster by other means than gotk. If we make this change then it should be for all `create` commands not only Git.

allymparker:
Hope you all had a good weekend! 

I've removed changing the name of generated secrets and added the same flag to Helm and Bucket sources. 

Let me know if you need any other changes

stefanprodan:
@allymparker you need to rerun the docs gen `make docs`

allymparker:
I think that's sorted @stefanprodan

hiddeco:
@allymparker any idea why cbba648 seems to touch all LOC? Is it a disguised merge commit?

stefanprodan (approved):
LGTM

Thanks @allymparker can you please squash all your commits into one and rebase please?

allymparker:
@stefanprodan done, thanks! @hiddeco sorry, I wasn't sure, might have been the rebases, but it seems to be happier after squashing

hiddeco (approved):

PR #352, opened 2020-10-16
Refactor create, reconcile and resume commands

hiddeco:
* Take ObservedGeneration into account in readiness checks where
  applicable
* Reduce amount of code (and duplicate GETs) by working with pointers
  where possible
* Improve logged messages to properly take resource names into account
  and better describe processes

First point addresses https://github.com/fluxcd/helm-controller/issues/112


stefanprodan (approved):
LGTM

Great refactoring 💯

Issue #308, opened 2020-10-05
Create custom flags for options with validation rules


hiddeco:
We have [a couple of flags in the GitOps Toolkit](https://github.com/fluxcd/toolkit/blob/6cce0a39011fc6878981a9c5c57092486c18ba1c/cmd/gotk/main.go#L113-L118) (there may be more) that have complex validation rules.

To describe the available options in the help menus, ensure we are consistent in our validation (messages), and make them more reusable, we should turn these into custom flags [as we did for the SSH key flags](https://github.com/fluxcd/toolkit/blob/main/cmd/gotk/flags.go).

rimaddo:
I would be interested to try this. Just to clarify that I understand correctly, you want to have and equivalent method to `Description` on `PublicKeyAlgorithm` for the other flags and then presumably you would call that [like you're doing here in the init](https://github.com/fluxcd/toolkit/blob/6cce0a39011fc6878981a9c5c57092486c18ba1c/cmd/gotk/create_source_git.go#L105)? Thanks!

hiddeco:
@rimaddo thanks for your interest! :kiwi_fruit: 

The highlighted lines in the first link are all tied to certain flags used within the commands. To de-duplicate this logic, we want a flag for every validation option.

One example:

```go
var supportedArchs = []string{"amd64", "arm", "arm64"}

type Arch string

func (a *Arch) String() string {
	return string(*a)
}

func (a *Arch) Set(str string) error {
	if strings.TrimSpace(str) == "" {
		return fmt.Errorf("no arch given, must be one of: %s",
			strings.Join(supportedArch, ", "))
	}
	for _, v := range supportedArchs {
		if str == v {
			*a = Arch(str)
			return nil
		}
	}
	return fmt.Errorf("unsupported arch '%s', must be one of: %s",
		str, strings.Join(supportedArchs, ", "))
}

func (a *Arch) Type() string {
	return "arch"
}

func (a *Arch) Description() string {
	return fmt.Sprintf("cluster architecture (%s)", strings.Join(supportedArchs, ", "))
}
```

You can than make the following change to initialize the default:

https://github.com/fluxcd/toolkit/blob/5efa1ebe88e1a4271e58293bf9f0a976e5ae4d93/cmd/gotk/install.go#L65

```go
	installArch               Arch = "string"
```

Plus (as you mentioned), register the new custom flag type in the command's `init` (including the nice description):

[...]

PR #356, opened 2020-10-17
Implement custom flags for options with validation rules

StupidScience:
Fix: #308 

Please look if you're ok with decoupling utils and flags to internal packages

### DIFF
--- a/internal/flags/log_level.go
+++ b/internal/flags/log_level.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2020 The Flux CD contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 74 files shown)


internal/flags/log_level.go
```diff
@@ -0,0 +1,54 @@
...
+	*l = LogLevel(str)
+	return nil
+}
+
+func (l *LogLevel) Type() string {
+	return "logLevel"
+}
+
+func (l *LogLevel) Description() string {
+	return fmt.Sprintf("log level, available options are: (%s)", strings.Join(supportedArchs, ", "))
```
stefanprodan, line 53:
```suggestion
	return fmt.Sprintf("log level, available options are: (%s)", strings.Join(supportedLogLevels, ", "))
```

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @StupidScience nice refactoring 🏅

hiddeco (approved):
Thank you @StupidScience :tulip:

Issue #359, opened 2020-10-19
gotk uninstall fails for installations with cherry-picked components


wingsofovnia:
## Problem
`gotk uninstall` cmd assumes all toolkit components are installed.

Where: [here](https://github.com/fluxcd/toolkit/blob/4eaf72fa3e017ad5b5d1c405692838996e1c2da1/cmd/gotk/uninstall.go#L101)

## Preconditions
```
gotk install  --components=source-controller,helm-controller --version=latest
```

## Actual behaviour
```
gotk uninstall --crds
Are you sure you want to delete the gotk-system namespace: y
► uninstalling custom resources
error: the server doesn't have a resource type "Kustomization"
```

## Expected behaviour
```
gotk uninstall --crds
Are you sure you want to delete the gotk-system namespace: y
► uninstalling custom resources
No resources found
No resources found
No resources found
No resources found
► uninstalling components
customresourcedefinition.apiextensions.k8s.io "buckets.source.toolkit.fluxcd.io" deleted
customresourcedefinition.apiextensions.k8s.io "gitrepositories.source.toolkit.fluxcd.io" deleted
customresourcedefinition.apiextensions.k8s.io "helmcharts.source.toolkit.fluxcd.io" deleted
customresourcedefinition.apiextensions.k8s.io "helmreleases.helm.toolkit.fluxcd.io" deleted
customresourcedefinition.apiextensions.k8s.io "helmrepositories.source.toolkit.fluxcd.io" deleted
clusterrolebinding.rbac.authorization.k8s.io "cluster-reconciler-gotk-system" deleted
namespace "gotk-system" deleted
✔ uninstall finished
```

## Workaround
Install the expecting component that is missing:
```
gotk install  --components=kustomize-controller --version=latest
```

rumsrami:
I am currently still facing the same problem, and this is the output from the command line
`
$ flux uninstall --resources --crds --namespace=flux-system
Are you sure you want to delete the flux-system namespace: y█
► uninstalling custom resources
error: the server doesn't have a resource type "GitRepository"
✗ kubectl: exit status 1
error: the server doesn't have a resource type "HelmRepository"
✗ kubectl: exit status 1
error: the server doesn't have a resource type "Bucket"
✗ kubectl: exit status 1
kustomization.kustomize.toolkit.fluxcd.io "flux-system" deleted

`

PR #361, opened 2020-10-19
Ignore missing kinds when uninstalling

stefanprodan:
Ignore missing kinds when uninstalling, do not assume `bootstrap` was used.

Fix: #359

### DIFF
--- a/cmd/gotk/uninstall.go
+++ b/cmd/gotk/uninstall.go
@@ -22,7 +22,9 @@ import (
 
 	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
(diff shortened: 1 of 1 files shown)


cmd/gotk/uninstall.go
```diff
@@ -83,37 +83,55 @@ func uninstallCmdRun(cmd *cobra.Command, args []string) error {
...
+	err = kubeClient.Get(ctx, releaseName, &release)
+	if err == nil {
+		release.Spec.Suspend = true
+		if err := kubeClient.Update(ctx, &release); err != nil {
+			return fmt.Errorf("unable to suspend release '%s': %w", releaseName.String(), err)
+		}
+	}
+	if err == nil || apierrors.IsNotFound(err) {
+		namespacedKinds = append(namespacedKinds, helmv2.HelmReleaseKind)
+	}
```
hiddeco, line 128: Given we do not bootstrap using a `HelmRelease` ourselves, this name assumption is likely wrong and the code thus obsolete?

stefanprodan: The code tests if the HelmRelease Kind exists, should I remove the suspend update?

hiddeco: Yes, I think that would be less confusing (at least to me).

hiddeco (approved):
Much easier to follow :100:

PR #376, opened 2020-10-26
Refactor manifest generation into a separate package

phillebaba:
The reason the manifest generation is refactored into a separate package is to make it possible for other projects to use the code.

The change also includes changing the return parameters as the file path and the content is important for the Terraform provider rather than reading files from a temp directory.

### DIFF
--- a/pkg/sync/options.go
+++ b/pkg/sync/options.go
@@ -1,0 +1,23 @@
+package sync
+
+import "time"
+
+type Options struct {
+	Interval   time.Duration
+	URL        string
+	Name       string
+	Namespace  string
+	Branch     string
+	TargetPath string
+}
+
+func MakeDefaultOptions() Options {
+	return Options{
+		Interval:   1 * time.Minute,
+		URL:        "",
+		Name:       "gotk-system",
+		Namespace:  "gotk-system",
+		Branch:     "main",
+		TargetPath: "",
+	}
+}
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -1,0 +1,88 @@
+package sync
+
+import (
+	"fmt"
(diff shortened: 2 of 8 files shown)


pkg/sync/sync.go
```diff
@@ -0,0 +1,88 @@
...
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
+)
+
+const (
+	bootstrapSourceManifest        = "toolkit-source.yaml"
+	bootstrapKustomizationManifest = "toolkit-kustomization.yaml"
+)
+
+func Generate(options Options) ([]map[string]string, error) {
```
phillebaba, line 22: Should we run a struct here with the parameters? Then we can use it in the other generate function.

pkg/sync/options.go
```diff
@@ -0,0 +1,23 @@
...
+import "time"
+
+type Options struct {
+	Interval   time.Duration
+	URL        string
+	Name       string
+	Namespace  string
+	Branch     string
+	TargetPath string
+}
```
phillebaba, line 12: Should there be an option to generate bucket manifests?

stefanprodan: No since bootstrap only works with Git.

pkg/sync/sync.go
stefanprodan: Yes, let's create an `Output` struct that contains an array of `Manifest` objects, the `Manifest` could have `Path` and `Content`.

stefanprodan (approved):
LGTM

I'll get this in and try to refactor it further, thanks @phillebaba 🍰

PR #378, opened 2020-10-27
change the windows binary release format to .zip

chanwit:
This PR changes the Windows binary release format from .tar.gz to .zip.


.goreleaser.yml
```diff
@@ -1,22 +1,37 @@
...
     goos:
-      - darwin
       - linux
-      - windows
     goarch:
       - amd64
       - arm64
       - arm
-    goarm:
-      - 7
```
stefanprodan, line 15: Removing this produces v6 binaries and that breaks ARMv7 compatibility.

chanwit: my bad. it's a mistake.

stefanprodan (approved):
LGTM

Thanks @chanwit 🍰

PR #394, opened 2020-10-29
Allow HTTP01 challenge for ingresses

gaetanars:
I propose this NetworkPolicy to allow Acme HTTP01 challenge for those who wants to add an ingress in gotk namespace.

### DIFF
--- a/manifests/policies/allow-acme.yaml
+++ b/manifests/policies/allow-acme.yaml
@@ -1,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-acme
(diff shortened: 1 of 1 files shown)


shanesveller:
Just another user here from the peanut gallery - this seems like something you would commit to your own local GitOps repo to me? Not everyone uses LetsEncrypt, HTTP challenge, or Ingress v1. Those are (popular, reasonable) implementation details of a particular cluster/stack. I personally see this more as `contrib`-style content as a result.  Perhaps public documentation for how to do this when desired would be good instead?

gaetanars:
You’re right, maybe in doc section.

PR #396, opened 2020-10-29
Delete generated docs before regenerating them

hiddeco:
This ensures no stray files are kept when for example a command name
changes.


stefanprodan (approved):
💯

PR #401, opened 2020-10-30
Add migration guide for Flux v1 users

stefanprodan:
This PR contains the migration guide from Flux v1 (read-only mode/without automated image updates) to Flux v2.

Preview: https://toolkit.fluxcd.io/guides/flux-v1-migration/


hiddeco (approved):
This is a step in the right direction, but as noted by our early adopters in some discussions and/or issues, we should aim for a declarative first and disaster resilient documentation.

PR #416, opened 2020-11-03
Fix datasource for cluster Grafana dashboard

L3o-pold:


stefanprodan (approved):
LGTM

Thanks @L3o-pold

PR #425, opened 2020-11-05
docs: fixing command for expected output

plod:
When following the guide for generic git server, I came across this problem which I solved, thought I'd do a PR to update the docs. 

The \ before the comment here means the subsequent lines end up getting commented out (and then not exported), see: https://stackoverflow.com/questions/9522631/how-to-put-a-line-comment-for-a-multi-line-command


hiddeco (approved):
Good catch, thank you :fishing_pole_and_fish:

PR #435, opened 2020-11-07
Update authentication hint for Azure

### LINKED ISSUE #427: Document Azure SOPS authentication
The current documentation for how to use the SOPS plugin with the kustomize-controller does not describe any methods for authenticating with Azure.

We should add a note about how to use aad-pod-identity together with the kustomize-controller.

phillebaba:
Updates the authentication hint for Azure, as it is not possible to add IAM permissions to a service account in Azure.

Fixes #427

### DIFF
--- a/docs/guides/mozilla-sops.md
+++ b/docs/guides/mozilla-sops.md
@@ -1,12 +1,12 @@
 # Manage Kubernetes secrets with Mozilla SOPS
 
 In order to store secrets safely in a public or private Git repository, you can use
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Changes look OK to me, but given this is more @stefanprodan his area of expertise. I'll request an additional review. :tulip:

stefanprodan (approved):
LGTM 

Thanks 🍰

Issue #445, opened 2020-11-11
Question - use a RAW values.yaml for a Helm release


0x53fa29jhezc:
Hey guys,
like flux so far but I struggle a bit at the moment with integrating big value.yaml config files

For example for bitnami MongoDB Chart the production-values.yaml is very long (500 lines).
I do not want to copy paste this into the helm release yaml file

What is the best way to do that?

there is this section with

```
spec:
  valuesFrom:
  - kind: ConfigMap
    name: prod-env-values
    valuesKey: values-prod.yaml
```

BUT I do not want to create a ConfigMap for every Helm chart. In my opinion this would be a bit confusing to have the values somewhere outside the git repo

is there kustomize magic which merges the "production-values.yaml" (without changing it)  into helm-release-yaml?

stefanprodan:
If you want to use the `values-production.yaml` file from inside the chart then do:

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: mongodb
spec:
  chart:
    spec:
      chart: mongodb
      sourceRef:
        kind: HelmRepository
        name: bitnami
      valuesFile: "values-production.yaml"
```

> BUT I do not want to create a ConfigMap for every Helm chart. In my opinion this would be a bit confusing to have the values somewhere outside the git repo

All Kubernetes manifests should be in the Git repository, you can place the raw production-values.yaml inside the repo and use Kustomize to generate the configmap:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- mongo-namespace.yaml
- mongo-helm-releases.yaml

configMapGenerator:
- name: mongo-config
  files:
    - values.yaml=configs/production-values.yaml
```

Then in mongo-helm-releases.yaml:

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: mongodb
spec:
  chart:
    spec:
      chart: mongodb
      sourceRef:
        kind: HelmRepository
        name: bitnami
  valuesFrom:
  - kind: ConfigMap
    name: mongo-config
```

0x53fa29jhezc:
Thank you very much!

with the additional option

generatorOptions:
  disableNameSuffixHash: true

otherwise there will be a suffix mongo-config-234234234 and the config map can not be found in the release



```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
[...]

PR #446, opened 2020-11-11
Add values file example to Helm docs

stefanprodan:
Fix: #445


hiddeco (approved):
Thank you @stefanprodan 🌷

PR #468, opened 2020-11-16
Fix --dry-run option

RossyWhite:
Currently, the components are installed even with `--dry-run` option.


stefanprodan (approved):
LGTM

Thanks @RossyWhite 🍰

Issue #479, opened 2020-11-19
flux reconcile is performed too soon


chanwit:
According to this code:
https://github.com/fluxcd/flux2/blob/main/cmd/flux/reconcile_source_git.go#L76

when we force reconciliation, the CLI annotates timestamp to objects.
In many of my observation especially when using a local cluster, this process triggers reconciliation too soon (most of the time, it's done before the new commits arrive the remote host).

would it be possible to make sure that flux gets reconciled exactly the commits just pushed to the remote?

PR #484, opened 2020-11-20
Utilize LastHandledReconcileAt for source reconcile commands

hiddeco:
Fix: #479


stefanprodan (approved):
LGTM

Thanks @hiddeco 🍰

Issue #213, opened 2020-09-07
Cover more HelmRelease configuration options


hiddeco:
The `gotk create helmrelease` command currently offers the bare minimum of options [made available by the API](https://github.com/fluxcd/helm-controller/blob/master/docs/api/helmrelease.md).

We would like to extend the available flags to a balanced set that works from a CLI perspective. Given the verbosity of the API, this may mean that some options will be ignored or that there are certain flags that provide "shortcuts" to a configuration set (remediation strategies is one thing that comes to mind here).

Tasks:

- [ ] Discuss / agree on the configuration options that should be covered
- [ ] Discuss / agree on the flags that control the options
- [ ] Implement the flags

juliohm1978:
Just stumbled into this issue, looking for a way to use HelmRepositories from different namespaces.

Currently, the `flux create helmrelease` command assumes every HemRepository resides inside the `flux-system` namespace? In order to group manifests together, I have to first generate a HelmRelease yaml and manually edit the chart namespace reference.

kingdonb:
I ran into this issue last week when I was trying to create a HelmRelease with `spec.timeout` -- the `--timeout` flag does not have any effect, since it is the global timeout, it only determines how long the CLI waits before timing out. There is no way to set `spec.timeout` at create time with `flux create helmrelease`.

There are about half a dozen other timeouts to set in the HelmRelease between install, uninstall, upgrade, test, rollback, ... I think there is a case to be made for maybe not covering them all in the CLI, especially since these structs could change one day. But the most common ones I think should be covered, especially `spec.timeout`, since it is essential for Helm Controller users to set `spec.timeout` if they want to receive error messages when something went wrong, or debug health check failures in a defined amount of time.

PR #491, opened 2020-11-22
Add values-from arg to create helmrelease cmd

stefanprodan:
This PR adds `--values-from` arg to create hr command, example usage:

```sh
# create Kubernetes secret with values
kubectl -n app create secret generic my-secret-values \
  --from-file=values.yaml=/path/to/my-secret-values.yaml

# create a HelmRelease with values from the Kubernetes secret
flux -n app create helmrelease redis \
  --source=HelmRepository/bitnami \
  --chart=redis \
  --values-from=Secret/my-secret-values
```

Ref: #213

### DIFF
--- a/cmd/flux/create_helmrelease.go
+++ b/cmd/flux/create_helmrelease.go
@@ -68,6 +68,14 @@ var createHelmReleaseCmd = &cobra.Command{
     --chart=podinfo \
     --values=./my-values.yaml
 
+  # Create a HelmRelease with values from a Kubernetes secret
+  kubectl -n app create secret generic my-secret-values \
+	--from-file=values.yaml=/path/to/my-secret-values.yaml
+  flux -n app create hr podinfo \
+    --source=HelmRepository/podinfo \
+    --chart=podinfo \
+    --values-from=Secret/my-secret-values
+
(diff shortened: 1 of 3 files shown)


cmd/flux/create_helmrelease.go
```diff
@@ -171,6 +181,13 @@ func createHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		helmRelease.Spec.Values = &apiextensionsv1.JSON{Raw: json}
 	}
 
+	if hrValuesFrom.String() != "" {
```
hiddeco, line 184: I think it would be better if we create a custom type like [`stringSliceValue`](https://github.com/spf13/pflag/blob/master/string_slice.go) that accepts multiple `--valuesFrom` references.

stefanprodan: This could come later, as it will not break the current command if we allow multiple values later on right?

hiddeco: That is absolutely true, just foreseeing that once a user see this is available they expect what I describe to be possible.

hiddeco (approved):

PR #496, opened 2020-11-23
Add reference to the bootstrap migration

gtseres:
Adding a reference to the bootstrap migration, as the section is named `Gitops migration`, so that it's clear which is the bootstrap migration we refer to in the section.


docs/guides/flux-v1-migration.md
```diff
@@ -60,7 +60,7 @@ Flux v1 repository to the bootstrap one.
 ## In-place migration
 
 !!! warning
-    For production use we recommend using the **bootstrap** procedure,
+    For production use we recommend using the **bootstrap** procedure (see the `Gitops migration` section above),
```
stefanprodan, line 63:
```suggestion
    For production use we recommend using the **bootstrap** procedure (see the [Gitops migration](#gitops-migration) section above),
```

stefanprodan (approved):
LGTM

Thanks @gtseres

PR #505, opened 2020-11-25
Add service accounts to tenant command

stefanprodan:
Generate a service account with limited access to its namespace, for each namespace owned by a tenant.

```
flux create tenant dev-team --with-namespace apps --export
```

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    toolkit.fluxcd.io/tenant: dev-team
  name: apps
---
kind: ServiceAccount
metadata:
  labels:
    toolkit.fluxcd.io/tenant: dev-team
  name: dev-team
  namespace: apps
---
kind: RoleBinding
metadata:
  labels:
    toolkit.fluxcd.io/tenant: dev-team
  name: gotk-reconciler
  namespace: apps
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: User
  name: gotk:apps:reconciler
- kind: ServiceAccount
  name: dev-team
```

### DIFF
--- a/cmd/flux/create_tenant.go
+++ b/cmd/flux/create_tenant.go
@@ -38,7 +38,7 @@ var createTenantCmd = &cobra.Command{
 	Use:   "tenant",
 	Short: "Create or update a tenant",
 	Long: `
-The create tenant command generates namespaces and role bindings to limit the
+The create tenant command generates namespaces, service accounts and role bindings to limit the
 reconcilers scope to the tenant namespaces.`,
 	Example: `  # Create a tenant with access to a namespace 
   flux create tenant dev-team \
@@ -89,6 +89,7 @@ func createTenantCmdRun(cmd *cobra.Command, args []string) error {
 	}
 
 	var namespaces []corev1.Namespace
+	var accounts []corev1.ServiceAccount
 	var roleBindings []rbacv1.RoleBinding
 
 	for _, ns := range tenantNamespaces {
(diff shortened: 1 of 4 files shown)


cmd/flux/create_tenant.go
```diff
@@ -65,7 +65,6 @@ var (
 )
 
 func init() {
-	createTenantCmd.Hidden = true
 	createTenantCmd.Flags().StringSliceVar(&tenantNamespaces, "with-namespace", nil, "namespace belonging to this tenant")
 	createTenantCmd.Flags().StringVar(&tenantClusterRole, "cluster-role", "cluster-admin", "cluster role of the tenant role binding")
```
hiddeco, line 69:
```suggestion
	createTenantCmd.Flags().StringVar(&tenantClusterRole, "cluster-role", "admin", "cluster role of the tenant role binding")
```

Prevents users from being deleting their own namespace, or fiddle with resource quotas.

stefanprodan: No that's no good, admin doesn't allow access to custom resources.

hiddeco (approved):

cmd/flux/create_tenant.go
stefanprodan: Also a tenant can't delete the namespace, for that to happen it needs to apply it first and the service account doesn't allow that.

Issue #507, opened 2020-11-25
Do not mark suspended resource as not ready


hiddeco:
At present the controllers mark a resource with a `Ready==False` state when a user defines the resource as suspended using `.spec.suspend`.

Given the aim of this value is to pause all reconciliation actions, this behavior is not accurate to the goal of the field. We must simply stop reconciling and observing state, and leave any previous observed state untouched as a "freeze in time".

To still make the state visible to the user, we must add an addition printer column that shows the `.spec.suspend` state of the resource (example here: https://github.com/fluxcd/helm-controller/pull/162), and a column to the `flux get [source]` commands.

- [x] helm-controller
- [x] source-controller
- [x] kustomize-controler
- [ ] add "suspended" column to `flux get [source]` commands

PR #512, opened 2020-11-26
Implement suspend/resume commands for sources

stefanprodan:
Fix: #507


hiddeco (commented):
I think we also need to add the column to the `flux get [source]` command for visibility.

hiddeco (approved):
:100:

PR #521, opened 2020-11-27
Improve installer list match for arm arches

relu:
`uname -m` will print out architecture codenames based on UTS_MACHINE
and COMPAT_UTS_MACHINE kernel defined constants. These extra values will
ensure the right version of the arm binary is installed on most Linux
systems running on ARM CPUs.

I was only able to test this on my RPiv3, `uname -m` reports `armv7l`. With the tweak, the installation is able to complete.


hiddeco (approved):
LGTM :100:

PR #525, opened 2020-11-29
Fix tenant and reconcile commands

stefanprodan:
Changes:
- Add service account namespace to tenant role binding
- Rename role binding to match the tenant name
- Do not try to reconcile a suspended object

### DIFF
--- a/cmd/flux/create_tenant.go
+++ b/cmd/flux/create_tenant.go
@@ -55,8 +55,7 @@ reconcilers scope to the tenant namespaces.`,
 }
 
 const (
-	tenantLabel       = "toolkit.fluxcd.io/tenant"
-	tenantRoleBinding = "gotk-reconciler"
+	tenantLabel = "toolkit.fluxcd.io/tenant"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


cmd/flux/create_tenant.go
```diff
@@ -123,18 +122,20 @@ func createTenantCmdRun(cmd *cobra.Command, args []string) error {
...
 				Namespace: ns,
 				Labels:    objLabels,
 			},
 			Subjects: []rbacv1.Subject{
 				{
-					Kind: "User",
-					Name: fmt.Sprintf("gotk:%s:reconciler", ns),
+					APIGroup: "rbac.authorization.k8s.io",
+					Kind:     "User",
+					Name:     fmt.Sprintf("gotk:%s:reconciler", ns),
```
phillebaba, line 133: Can users have non dns-resolvable names?

stefanprodan: no, see https://github.com/fluxcd/flux2/blob/17ba4464929a45ae7f90d622fdcb232d0f2c1e90/cmd/flux/create_tenant.go#L77

phillebaba (approved):
LGTM

hiddeco (commented):
I would put the annotation change in a separate PR as it should be included in the next MINOR release.

stefanprodan:
@hiddeco the annotation change was included in 0.3, should we wait for 0.5?

hiddeco:
@stefanprodan the change itself can’t be included in a PATCH release as it is breaking, so yes.

stefanprodan:
@hiddeco I removed the annotations from this PR.

hiddeco (approved):
:+1:

PR #538, opened 2020-12-01
Add commands for image automation API

squaremo:
This will add commands of all the expected varieties, for dealing with the image automation API.

EDIT: removed the dithering over `auto` vs `image`; the latter is more fluent (and less typing).

NB the ImagePolicy object does not get {reconcile,suspend,resume} subcommands, since it doesn't follow the reconcile/suspend protocol. It's sufficient at present for the ImageRepository to have these.


cmd/flux/get.go
```diff
@@ -33,3 +44,71 @@ func init() {
...
+	if !allNamespaces {
+		listOpts = append(listOpts, client.InNamespace(namespace))
+	}
+	err = kubeClient.List(ctx, get.list.AsObject(), listOpts...)
+	if err != nil {
+		return err
+	}
+
+	if get.list.Len() == 0 {
+		logger.Failuref("no imagerepository objects found in %s namespace", namespace)
```
hiddeco, line 101:
```suggestion
		logger.Failuref("no ImageRepository objects found in %s namespace", namespace)
```

(I aligned this awhile ago as we were using a mixture of camelcase and lowercase names throughout the codebase).

cmd/flux/get.go
```diff
@@ -33,3 +44,71 @@ func init() {
 		"list the requested object(s) across all namespaces")
 	rootCmd.AddCommand(getCmd)
 }
+
+type summarisable interface {
+	AsObject() runtime.Object
+	Len() int
+	SummariseAt(i int, includeNamespace bool) []string
```
hiddeco, line 50: `SummariseItem` or `SummariseRow` maybe? The `At` did not really ring any bells with me.

squaremo:
Pinned the two image-* API module versions to v0.1.0, and rebassed on main.

stefanprodan (approved):
LGTM

Happy holidays @squaremo 🎄

squaremo:
The gift of a PR review! 🎁 🎄 🎁

Issue #360, opened 2020-10-19
gotk create source git does not create secret for ssh


vchrisb:
I'm using following gotk (version 0.1.8, macos) command to create a git source with ssh for a private github repo;
```
gotk create source git reponame \
  --url="ssh://git@github.com/vchrisb/<repo>" \
  --branch=master \
  --interval=30s \
  --export
```

But it somehow gotk does not recognise, that the scheme is `ssh` and hence doesn't trigger the creation of a deploy key.

The output is:
```
---
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: k8s-home
  namespace: gotk-system
spec:
  interval: 30s
  ref:
    branch: master
  url: ssh://git@github.com/vchrisb/<repo-name>
```

I looked at the sources, but can't figure out why it isn't triggered https://github.com/fluxcd/toolkit/blob/main/cmd/gotk/create_source_git.go#L188

stefanprodan:
When using export we don't generate secrets, as that means storing the secret in clear in your repo. If you want to do that, then run `gotk create source git` without `--export` then do:
```
gotk export source git reponame --with-credentials > source.yaml
```

vchrisb:
thx. Still not sure I understand the flow here.
Why not doing the same for bot cases, with `--export` and without? Just add the option `--with-credentials` if one wants the credentials exported?

Currently I understand I have two options to persist the source in the fleet repo:
```
gotk create source git reponame \
  --url="ssh://git@github.com/vchrisb/<repo>" \
  --branch=master \
  --interval=30s
gotk export source git reponame > ./dev-cluster/reponame.yaml
```
or alternatively:
```
opensssl ...
kubectl create secret --name reponame_secret ...
gotk create source git reponame \
  --url="ssh://git@github.com/vchrisb/<repo>" \
  --branch=master \
  --interval=30s  \
 --secret-ref=reponame_secret  \
 --export > ./dev-cluster/reponame.yaml
```

sorry I'm new with the toolkit and may misunderstand the flow. :/

PR #547, opened 2020-12-04
Add create secret git command

stefanprodan:
This PR adds a command for generating Kubernetes secrets with Git credentials.

```sh
# Create a Git SSH authentication secret using an ECDSA P-521 curve public key
  flux create secret git podinfo-auth \
    --url=ssh://git@github.com/stefanprodan/podinfo \
    --ssh-key-algorithm=ecdsa \
    --ssh-ecdsa-curve=p521

  # Create a secret for a Git repository using basic authentication
  flux create secret git podinfo-auth \
    --username=username \
    --password=password

  # Create a Git SSH secret on disk and print the deploy key
  flux create secret git podinfo-auth \
    --url=ssh://git@github.com/stefanprodan/podinfo \
    --export > podinfo-auth.yaml

  yq read podinfo-auth.yaml 'data."identity.pub"' | base64 --decode

  # Create a Git SSH secret on disk and encrypt it with Mozilla SOPS
  flux create secret git podinfo-auth \
    --namespace=apps \
    --url=ssh://git@github.com/stefanprodan/podinfo \
    --export > podinfo-auth.yaml

    --in-place podinfo-auth.yaml
```

Fix: #360
Fix: https://github.com/fluxcd/flux2-multi-tenancy/issues/3

### DIFF
--- a/cmd/flux/create_secret_git.go
+++ b/cmd/flux/create_secret_git.go
@@ -1,0 +1,173 @@
+/*
+Copyright 2020 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"crypto/elliptic"
+	"fmt"
+	"net/url"
+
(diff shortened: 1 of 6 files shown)


cmd/flux/create_secret_git.go
```diff
@@ -0,0 +1,170 @@
...
+
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    secretLabels,
+		},
+	}
+
+	if strings.HasPrefix(u.Scheme, "ssh") {
```
hiddeco, line 123: Think it would be more appropriate to do something like:

```
switch u.Scheme {
	case "http", "https":
		// HTTP auth logic
	case u.Scheme == "ssh":
		// key generator logic
	default:
		// throw err
}
```

hiddeco (approved):
LGTM :+1:

PR #548, opened 2020-12-04
Fix list parsing issue in the docs

relu:


docs/get-started/index.md
```diff
@@ -36,6 +36,7 @@ The install script downloads the flux binary to `/usr/local/bin`.
 
 If using Arch Linux, there are three packages published in AUR you can use to
 install `flux`:
+
```
hiddeco, line 39: The `scm` can be removed from the get started, but I would still name it in the `README.md`.

relu: Makes sense to me.

stefanprodan (approved):

hiddeco (approved):

PR #550, opened 2020-12-04
Publish install manifest to GitHub releases

stefanprodan:
Build and publish Flux `install.yaml` to GitHub release for those that don't want to use `flux bootstrap`.

After the next release, Flux can be installed with kubectl:

```
kubectl apply -f https://github.com/fluxcd/flux2/releases/latest/download/install.yaml
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,24 +59,9 @@ jobs:
 
           # create tarball
           cd ./output && tar -cvzf manifests.tar.gz $files
-      - name: Create release
-        id: create_release
-        uses: actions/create-release@latest
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
-      - name: Upload artifacts
-        id: upload-release-asset
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./output/manifests.tar.gz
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -59,6 +59,9 @@ jobs:
 
           # create tarball
           cd ./output && tar -cvzf manifests.tar.gz $files
+      - name: Generate install manifest
+        run: |
+          kustomize build ./manifests/install > ./output/install.yaml
```
hiddeco, line 64: Having looked more into how GoReleaser does things, I would maybe change the `./output` folder to `./dist` so that we have all things related to the release build in one place.

Another option would be to make it part of the `GoReleaser` process, as it can work with custom commands. That would make it reproducible and a one-stepper.

relu: I think you'd need to install `kustomize` inside the GoReleaser runner for that work.

Maybe `dist/` should be added to `.gitignore` as well? I see there is one entry for `output/` already.

hiddeco: Mmm, that's actually a valid concern. Does the action have access to the files generated prior?

stefanprodan: I removed the GH Actions!

.github/workflows/release.yaml
```diff
@@ -59,24 +59,9 @@ jobs:
 
           # create tarball
           cd ./output && tar -cvzf manifests.tar.gz $files
-      - name: Create release
```
relu, line 62: Wait, was this intentional? Or would goreleaser create the release instead?

stefanprodan: Yes goreleaser will create it

hiddeco (approved):
🤞

.github/workflows/release.yaml
relu: Got it, thanks for the clarifications.

PR #567, opened 2020-12-09
Add Kustomize FAQ to docs

stefanprodan:


hiddeco (approved):
We may want to bring this a bit more to the foreground on the next-gen documentation website.

PR #580, opened 2020-12-10
Include image controller config in `flux install`

squaremo:
This adds kustomizations for `image-{reflector,automation}-controller`, which will get bundled into the file fetched by `flux install`.

It also adds the corresponding APIs into the update workflow, so they don't get left out.

Lastly, this introduces an argument `--extra-components`, which can be used to add components to the defaults. E.g., you can do

    flux install --extra-components image-reflector-controller,image-automation-controller

rather than having to use `--components` and name all of them including the defaults.

### DIFF
--- a/manifests/bases/image-automation-controller/kustomization.yaml
+++ b/manifests/bases/image-automation-controller/kustomization.yaml
@@ -1,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- https://github.com/fluxcd/image-automation-controller/archive/v0.1.0.zip//image-automation-controller-0.1.0/config/crd
+- https://github.com/fluxcd/image-automation-controller/archive/v0.1.0.zip//image-automation-controller-0.1.0/config/manager
+patchesJson6902:
+- target:
+    group: apps
+    version: v1
+    kind: Deployment
+    name: image-automation-controller
+  path: patch.yaml
--- a/manifests/bases/image-reflector-controller/kustomization.yaml
+++ b/manifests/bases/image-reflector-controller/kustomization.yaml
@@ -1,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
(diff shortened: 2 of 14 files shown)


manifests/bases/image-automation-controller/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
hiddeco, line 1: You likely want to automate updates for this in future releases, this can be done by adding a line to:

https://github.com/fluxcd/flux2/blob/main/.github/workflows/update.yml#L42

manifests/bases/image-reflector-controller/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
hiddeco, line 1: You likely want to automate updates for this in future releases, this can be done by adding a line to:

https://github.com/fluxcd/flux2/blob/main/.github/workflows/update.yml#L42

manifests/bases/image-automation-controller/kustomization.yaml
squaremo: In a different way to https://github.com/fluxcd/flux2/pull/580/commits/9d93adae72c0670b91a1c294f8ef9d6ca9f4df40#diff-571e661e1640e6f1f415c087c9a72af346ea9fd4655ff5545e5ad6ea596bab97R43 ?

hiddeco: :100:

hiddeco (approved):
:+1: :100:

stefanprodan:
I think we need to validate the extra components as we know what things it accepts. What about Terraform, shouldn't this be part of the manifest package?

PS. I would hold on merging this until we have the bootstrap equivalent.

squaremo:
> I think we need to validate the extra components as we know what things it accepts.

Can do; do the --components get validated?

stefanprodan:
We validate only the required ones, but I think we should validate them all, this can come in a different PR.

stefanprodan (approved):
LGTM

Thanks @squaremo 🎖️

Issue #579, opened 2020-12-10
flux writes informational messages to stdout that should go to stderr


squaremo:
If I do `flux install --accidentally-wrong > flux2.yaml` then I get an error message in flux2.yaml, rather than an error message printed to the console. Messages about what it's doing should go to stderr, rather than stdout.

PR #581, opened 2020-12-10
Log to stderr

hiddeco:
This commit refactors the `printLogger` into a `stderrLogger` that
properly logs to `os.stderr` instead of `os.stdout`.

Fixes: #579

### DIFF
--- a/cmd/flux/create_source_git.go
+++ b/cmd/flux/create_source_git.go
@@ -187,13 +187,13 @@ func createSourceGitCmdRun(cmd *cobra.Command, args []string) error {
 	if sourceGitSecretRef != "" {
 		withAuth = true
 	} else if u.Scheme == "ssh" {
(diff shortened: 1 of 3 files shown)


squaremo:
Is there a reason to log _anything_ to stdout rather than stderr?

hiddeco:
@squaremo likely not, I have amended my commit and replaced two `fmt.Printf` calls that also should be logged to stderr.

drewfreyling:
This brings back https://github.com/fluxcd/flux/issues/1491 from fluxv1 now....

hiddeco:
How does this bring back that issue? The change has impact on the `flux` CLI output, but not on the controller logs.

From a CLI I would expect it to log information to `stderr` while writing things like YAML manifests to `stdout`, otherwise it would be hard to use it in combination with pipes, etc.

PR #582, opened 2020-12-11
Propose security model for impersonation/tenancy

stealthybox:
This proposal describes a potential API design /w supporting controller implementation details for secure, multi-tenant API primitives within Flux.

The examples clarify the utility of certain changes which may at first seem esoteric.
This implementation should allow people using and building platforms on Flux to make concise choices toward implementing their own security needs and behavioral constraints.

Ideally this document and the surrounding conversations can graduate to becoming documentation for various audiences.

The ideas represented here are the work of many folks -- not limited to @stefanprodan @squaremo @hiddeco @jonathan-innis and myself.
For previous comments, see #263

### DIFF
--- a/docs/proposals/secure-impersonation.md
+++ b/docs/proposals/secure-impersonation.md
@@ -1,0 +1,636 @@
+# Secure Impersonation
+
+* [Context](#context)
+* [Goals](#goals)
+* [Design](#design)
+  - [controller serviceAccounts](#controller-serviceaccounts)
+  - [user/group names](#user-group-names)
+  - [serviceAccount impersonation](#serviceaccount-impersonation)
+  - [defaulting](#defaulting)
+  - [kubeConfig](#kubeconfig)
+  - [sourceRefs](#sourcerefs)
+  - [kubectl](#kubectl)
+* [Concerns](#concerns)
+  - [compatibiliity](#compatibiliity)
+  - [performance](#performance)
+* [Decisions](#decisions)
+* [Tasks](#tasks)
+* [Examples](#examples)
+  - [default install](#default-install)
+  - [tenants](#tenants)
+  - [cross-namespace binding against Flux Groups](#cross-namespace-binding-against-flux-groups)
+  - [opening Source policies with RBAC](#opening-source-policies-with-rbac)
+  - [rejecting kubeConfigs](#rejecting-kubeconfigs)
+
+## Context
+
+Flux 2's installation model is built from CRD's.
+This allows cluster operators to granularly control who can create/modify Sources and Applier type objects such as
+GitRepository and Kustomization and for which Namespaces.
+This also helps out the Deployment model.
+While we could for example, still run independent instances of ex:kustomize-controller in each Namespace listening to only
+Kustomize resources for that Namespace, it can use fewer resources to deploy a single controller watching all (or many) Namespaces.
+Having a single controller reduces the onboarding steps necessary for a new Namespace owner to use Flux Custom Resources.
+A central set of controllers is the primary Deployment mode for Flux.
+
+In order to manage resources across many Namespaces, the default install gives kustomize-controller and helm-controller a full
+cluster-admin ClusterRoleBinding.
+This means cluster owners need to be careful who can create Kustomizations and HelmReleases as well as which Sources are in the cluster
+due to the exposure they provide to act as cluster-admin.
+This restricts who can create Source and Apply type objects into a burdensome approval flow for cluster owners and provides
+poor delegation of self-service permissions which is a proven elements of good platforms.
+
+
+## Goals
+
+Flux should remain easy to install but promote secure practices in its default configuration.
+Flux's security model should allow a cluster administrator to delegate self-service management of
+Sources, Applies, Webhooks, Automations, and Notifications to Namespace owners/users without having to review every detail for
+compliance to security policies.
+Tenants or Namespace owners should be able to act freely without fear that their changes will overreach their permissions.
+
+Flux APIs must structurally prevent privilege escalation and arbitrary code execution.
+Flux APIs should be congruent with Kubernetes core API's security model.
+
+The implementation should be usable and allow users quick, concise decisions regarding their security needs.
+
+
+## Design
+
+Currently kustomize-controller and helm-controller apply Kustomizations and HelmReleases using their own ServiceAccount by default.
+It's possible to drop privileges using a `serviceAccountName` from the same Namespace -- this is more secure but this is an optional, explicit behavior.
+
+The central controllers should change so they always assume a different User when applying and maintaining objects from manifests in Sources.
+They should never use a controller SA when managing these resulting Objects, even for Garbage Collection, Health-Checks, or "values-from" merging.
+This will allow the audit log to be more granularly examined for controllers' mutations on Flux APIs vs. GitOps-managed objects.
+This technique prevents cross-tenant information disclosure and privilege escalation through misused Health-Checks or Garbage Collection.
+
+#### controller serviceAccounts
+The controller ServiceAccounts are far overprivileged for Flux API operations.
+They should have a ClusterRoleBinding (or RoleBinding for single Namespace installs) for necessary Flux API Kinds and other needed objects such as ConfigMaps/Secrets + a separate ClusterRoleBinding for User Impersonation.
+
+#### user/group names
+All usernames impersonated by kustomize-controller and helm-controller for managing resources will be Namespaced relative to the `Kustomization`/`HelmRelease` object.
+This is similar to the mechanism Kubernetes uses for [ServiceAccount User names](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-subjects)).
+Kubernetes uses `system:serviceaccount:{{namespace}}:{{metadata.name}}` for ServiceAccounts.
+The Flux username format should be `flux:user:{{namespace}}:{{spec.user}}`.
+Alternatively flux could use the reserved prefix `system:flux:user:<...>` -- this would break the rules, but make it less likely for other systems to accidentally or maliciously use the FluxUser namespace when authenticating with kube-apiserver.
+The default `spec.user` can be "reconciler".
+A user field will be added to the Kustomization and HelmRelease spec.
+It enables a partial override for the namespaced username the controller uses for that particular `Kustomization`/`HelmRelease`.
+Namespace Owners can use the user override to optionally specify flux `Users` with different `Roles`/`RoleBindings` for managing each workload,
+similar to Pods specifying non-default serviceAccounts from the same Namespace.
+- **Con**: RoleBinding against a FluxUser requires specifying the namespace in the username string which can be hard to do in a generic way with kustomize -- would need to document a kustomize example
+
+Flux controllers will also impersonate two Groups.
+The first is `flux:users` which is a generic group, analogous to kubernetes' `system:authenticated` -- binding against this group allows all reconcilers to access particular resources via RBAC.
+The second is `flux:users:{{namespace}}`, a namespace specific group which enables Namespace Owners to rolebind for reconcilers from another Namespace without mandating a particular username.
+Neither of these groups are optional or overrideable -- they are a given property of each reconciler's Namespace and identity.
+Note `users` is plural in the Group names, but not for a single User -- this matches kubernetes SA's.
+It could cause typos, but not matching k8s plural semantics could also be confusing. We could be silly and add both without notable consequence. 🚲🏠
+
+Impersonation of Users and Groups is done via API Headers that are permitted for the controller ServiceAccounts.
+- **Con**: controller SA's can impersonate any user or group, not just flux:user namespaced ones is explicitly specified via a ClusterRole (impractical,bad-ux)
+       controller code needs to be trusted not to impersonate 
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flux-user-group-impersonator
+rules:
+- apiGroups: [""]
+  resources: ["users", "groups"]
+  verbs: ["impersonate"]
+```
+
+Using Namespaced Users and Groups provides a security boundary in the same way the ServiceAccount User names do.
+It's never allowed for a reconcile to impersonate a non-namespaced User or user from outside the Namespace of the particular reconcile resource.
+If a reconcile needs access to resources from another Namespace, the other Namespace can have a RoleBinding targeting the reconciler's differently namespaced User name.
+Alternatively, a ClusterRoleBinding on the reconciler's namespaced User name can grant access to all Namespaces.
+
+#### serviceAccount impersonation
+`spec.serviceAccountName` will remain supported, but impersonating Users is preferable because it reduces the attack surface within a Namespace.
+`ServiceAccounts` produce privileged token Secrets in the same Namespace often for use within `Pods`.
+It's possible to impersonate a `ServiceAccount` User name even when they don't exist, but it's possible to create the resource after the fact creating generated tokens without a change in the reconciler behavior.
+This doesn't happen with impersonated `Users` -- `Users` don't have any object representation. This prevents `Pods` in same Namespace as the `Kustomization`/`HelmRelease` from assuming the role of the privileged reconciler which would typically be an inappropriate privilege escalation.
+
+Impersonation of reconciler ServiceAccounts is done via API Headers that are permitted for the controller ServiceAccounts.
+The FluxUser Groups can still be added to requests for impersonated ServiceAccounts.
+The token-fetching impersonation method currently implemented within kustomize-controller and helm-controller should be removed or flagged off.
+This is a trade-off:
+- *Pro*: Simpler tenant configs -- possible to bind against and impersonate a ServiceAccount User name that "doesn't exist"/isn't represented by a Resource -- user not required to actually create an SA
+- *Pro*: No per-tenant crypto/random-number dependency due to SA-tokens
+- *Pro*: Controller SA's no longer require cluster-wide access to Secrets
+- *Pro*: Possible to configure controller to only have ServiceAccount impersonation via RBAC -- disable arbitrary User and Group Impersonation if controller code is untrusted (prevent use of `flux:user:*`)
+- **Con**: Single namespace installs of Flux controllers may depend on ClusterRoleBindings to flux-sa-impersonator which requires cluster-admin setup permissions.
+       Cannot(?) restrict ServiceAccount User impersonation to specific namespaces.
+       Compare to SA's and token Secrets which a Namespace admin has full access to *only within their Namespace*.
+       SA Token Impersonation allows single namespace installs to have less dependency on the cluster-admin (as long as Flux CRD's are already registered) while still keeping reconciler specific users.
+       For this reason, it may be nice to keep `--sa-token-impersonation` as a controller flag and support it with the single-namespace installation option, despite the faults /w using SA's.
+       It's possible to do some heuristic fallback here, but it could complicate things.
+
+If SA Token impersonation is enabled, impersonated FluxUser Groups are omitted from the request.
+ServiceAccounts still have a Group that you can rolebind against: `system:serviceaccounts:{{namespace}}`.
+
+#### defaulting
+Behavior must be clarified when both `user` and `serviceAccountName` are provided.
+It should probably be a validation or runtime error condition.
+If we make this a singleton in the API, we could prevent this structural issue, but it will be a bigger breaking change:
+```yaml
+# example singleton style API
+spec:
+  user:
+    kind: ServiceAccount  # defaults to User
+    name: reconciler-svc
+```
+If `user` and `serviceAccountName` are unset/empty-strings, the default FluxUser (proposed: `reconciler`) is used.
+Alternatively, we could use [CRD defaulting](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting), but this puts other considerations on the design with detecting empty values.
+
+#### kubeConfig
+Kustomization and HelmRelease already have a `kubeConfig` option in addition to `serviceAccountName` and the proposed `user`.
+KubeConfig's are loaded from Secrets in the same Namespace in the style of Cluster API.
+They are safe to use literally as their own credential, since the FluxUser can be granted access to it as opposed to problematically assuming a controller SA.
+Not all kubeConfigs may be assumed to have impersonation rights, but there is no field that explicitly disables User impersonation.
+When `user`/`serviceAccountName` is unset and `kubeConfig` is non-empty, we could decide to ignore impersonation, and use the literal credential.
+Impersonation of Users and ServiceAccounts as well as ServiceAccount Token Impersonation can still function with an explicitly set value and does work across cluster boundaries.
+An impersonated User for the remote cluster will always match the Namespace used for the reconciler in the source cluster.
+
+KubeConfigs that mention the controller SA token in `/var/run/secrets/kubernetes.io/serviceaccount` must be rejected, because this is a privilege escalation route.
+These filepaths need to be resolved to absolute paths before they are checked to prevent subversive path traversal via symlinks.
+
+KubeConfigs that set `users[].user.exec.command` or `users[].user.auth-provider.config.cmd-path` should be rejected unless all `[command|cmd-path]` values return
+an [`os.exec.LookPath()`](https://golang.org/pkg/os/exec/#example_LookPath) inside the `/kubeconfig-bin/` of the controller Pod filesystem.
+While executing the kubeConfig, the `$PATH` likely needs to be preserved in case the auth-helper needs to execute other POSIX binary dependencies.
+Restricting `[command|cmd-path]` to `/kubeconfig-bin/` allows the administrator an opt-in directory where they can add tools they trust/accept like `gcloud` and `aws-iam-authenticator` 
+while disallowing access to more dangerous binaries like `kubectl`. This mechanism is somewhat inspired by `/sbin` and `setuid` binaries.
+
+Using exec helpers is not an approach used by Cluster API.
+In addition to being insecure, it's not possible to distrubite controllers containing all of the necessary binaries to support all kubernetes platforms.
+Cluster API providers instead schedule regular updates to kubeConfig Secret data which is more tenant friendly and dependency-free.
+
+Regardless, we must sanitize the kubeConfig, so it's easy to add a restricted, PATH-based allow-list that the flux-system owner can control if they choose to manage remote clusters in this way.
+
+
+#### sourceRefs
+Controllers should impersonate the reconciler User when fetching sourceRefs.
+This allows for users to open up Source access policies via RBAC, rather than relying on a Gatekeeper or Kyverno installation to restrict default Flux behavior.
+ClusterRoles should exist for the Source controller API's `flux-source-viewer`, `flux-gitrepo-viewer`, `flux-helmrepo-viewer`, `flux-bucket-viewer`, etc.
+FluxUsers should be bound explicitly against these ClusterRoles, either with RoleBindings (for single Namespaces) or ClusterRoleBindings.
+Cross-Namespace access of Sources can be enabled per Namespace or across the whole cluster via RoleBindings/ClusterRoleBindings to a FluxUser, Flux Group, or All FluxUsers.
+Kinds of sources can be differentiated, and individual Sources can be permitted per-User/Group through `resourceNames` within Role rules.
+The current API objects that use sourceRefs are `Kustomization` and `HelmChart`.
+`Kustomization` and `HelmRelease` have the `serviceAccountName` proposed `user` fields for impersonation.
+Both of these impersonation fields should also be added to `HelmChart` because reconciling one requires accessing other Sources.
+source-controller will need impersonation behavior implemented /w additional RBAC and any supporting flags/options.
+Alternatively, `HelmChart` could be moved into the helm-controller control loops instead of source-controller.
+It's not sufficient to gate the creation of `HelmCharts` based off of a `sourceRef` permission check in `HelmRelease`;
+RBAC rules can change at any time, so the source-controller API client should have the constraints of the impersonated User.
+
+If a kubeConfig is specified for a resource that specifies a sourceRef, a separate client matching the same User or SA from the remote cluster is used in the source cluster.
+This is sensible because the User for the remote cluster must match the source cluster's reconciling Namespace.
+If no User is specified for the remote cluster, the default FluxUser is impersonated to fetch the sourceRef from the source cluster.  
+
+
+#### notifications
+notification-controller needs to be able to impersonate FluxUsers and ServiceAccounts.
+`Recievers` specify `resources` which are Sources -- `Alerts` specify `eventSources` which can be Sources or Applies.
+Both impersonation fields should also be added to `Recievers` and `Alerts` to allow for the same kind of access polices possible with sourceRefs.
+
+This will prevent any cross-tenant misuse of the notification API's: notably denial-of-service through Recievers and information-disclosure through Alerts.
+Providers appear to be safe as is.
+
+The default install's flux-system:cluster-admin FluxUser can still be used to configure Recievers and Alerts for all objects in the cluster.
+In this configuration, other Namespaces will need to be granted access.
+
+It might be possible to get `alert.spec.eventSources` and `reciever.spec.resources` to work across clusters by specifying `kubeConfig`;
+this could open up some interesting management cluster patterns for monitoring credentials and centralized webhooks.
+
+
+#### cluster-role aggregation
+We should use [ClusterRole Aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) to configure the the top-level flux ClusterRoles.
+
+This means we can define `flux-gitrepo-viewer`, `flux-helmrepo-viewer`, `flux-bucket-viewer` and use an aggregation label to define `flux-source-viewer` without repeating ourselves.
+Adding another source in the future does not require updating the `flux-source-viewer` object, just creating a new aggregated role.
+
+It's possible to take this one step further and aggregate `flux-source-viewer` into the kubernetes default role for `edit`.
+It would be useful for `view` as well, depending on your security model.
+`flux-apply-editor` and `flux-source-editor` might be a good ClusterRoles to aggregate under `admin` since it's useful for enabling tenants to be
+self-service within their Namespace from their own control-repo.
+
+We need to work out which roles are appropriate to aggregate under any default Kubernetes roles, and whether this should be done with a default
+flux install or taken as an install option.
+
+> Note: `admin` is an aggregated role containing `edit` which contains `view`, but `cluster-admin` is not an aggregated role -- it explicitly allows all resources including any Flux objects.
+
+
+#### kubectl
+`client-go` currently has an issue where `kubectl --as=user --as-group=group` is ignored when using the in-cluster config
+for Pod ServiceAccount tokens.
+This prevents kustomize-controller from using its Pod ServiceAccount to impersonate FluxUsers when applying resources
+by exec'ing `kubectl`.
+The workaround to this bug is to heuristically detect the in-cluster config and supply it explicitly with equivalent
+flags or a kubeConfig file for the token, CA, apiserver-endpoint, and protocol.
+This bug affects all binaries using client-go's flag-options parser, but does not affect libraries like helm that use
+injected clients.
+
+We can maintain a workaround library that shims the kubectl exec or move to using kubectl as a library directly.
+
+
+## Concerns
+#### compatibiliity
+These changes are mostly breaking. Existing Flux deployments are not going to have RBAC configured for every reconciler.
+Some reconciler ServiceAccounts may need `flux-source-viewer` bound to them.
+Others may prefer to update their serviceAccountNames to user.
+Migration/Upgrade documentation will be necessary.
+It's possible to build a tool-assisted migration that modifies resources in a git repo. (`flux migrate <git-repo-path>` ?)
+
+#### performance
+Flux controllers will be using many more Kubernetes client Objects, informers, and caches.
+Storing a map of Flux Users to clients/caches might be a good first step.
+kubeConfig client/caches are not stored currently, which was an explicit decision to support refreshing clients,
+this this doesn't have to be true for all kubeConfigs.
+
+We have done some performance testing in the past, but it would be good to have a strategy/tools to help us
+understand performance regressions.
+
+## Decisions
+Direction should be decided on the following:
+
+- [ ] `flux:user` vs. `system:flux:user` User prefix
+- [ ] `flux:users` vs. `flux:user` Group prefix
+- [ ] default FluxUser name (proposed: "reconciler")
+- [ ] source-controller impersonation vs. HelmChart moving to helm-controller
+- [ ] HelmChart templates specifying a different user/serviceAccountName than their parent HelmRelease
+- [ ] kubeConfig for management cluter usage of notification-controller Recievers and Alerts
+- [ ] user + serviceAccount validation vs. API change deprecating `serviceAccountName`
+- [ ] user / serviceAccount defaulting or empty value behavior
+- [ ] kubeConfigs require impersonation permission vs. no impersonation on empty or default API value
+- [ ] serviceAccount token impersonation feature-flag
+- [ ] ClusterRole aggregation for flux objects in k8s default `view`, `edit`, `admin`
+
+
+## Tasks
+- [ ] add `user` to Kustomization
+- [ ] add `user` to HelmRelease
+- [ ] add `user`, `serviceAccountName` to HelmChart
+- [ ] create client generation library for kubeConfig, user, serviceAccountName
+- [ ] use client-gen library in kustomize-controller
+  - [ ] sourceRefs
+  - [ ] garbage-collector
+  - [ ] prune
+- [ ] use client-gen library in helm-controller
+  - [ ] sourceRefs
+  - [ ] helm client
+  - [ ] valuesFrom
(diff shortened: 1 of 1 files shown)


docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+  - [tenants](#tenants)
+  - [cross-namespace binding against Flux Groups](#cross-namespace-binding-against-flux-groups)
+  - [opening Source policies with RBAC](#opening-source-policies-with-rbac)
+  - [rejecting kubeConfigs](#rejecting-kubeconfigs)
+
+## Context
+
+Flux 2's installation model is built from CRD's.
+This allows cluster operators to granularly control who can create/modify Sources and Applier type objects such as
+GitRepository and Kustomization and for which Namespaces.
```
squaremo, line 29:
```suggestion
GitRepository and Kustomization, and for which Namespaces.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+This also helps out the Deployment model.
+While we could for example, still run independent instances of ex:kustomize-controller in each Namespace listening to only
```
squaremo, line 31: I didn't catch what the shorthand `ex:kustomize-controller` stands for.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+Kustomize resources for that Namespace, it can use fewer resources to deploy a single controller watching all (or many) Namespaces.
+Having a single controller reduces the onboarding steps necessary for a new Namespace owner to use Flux Custom Resources.
+A central set of controllers is the primary Deployment mode for Flux.
+
+In order to manage resources across many Namespaces, the default install gives kustomize-controller and helm-controller a full
+cluster-admin ClusterRoleBinding.
+This means cluster owners need to be careful who can create Kustomizations and HelmReleases as well as which Sources are in the cluster
+due to the exposure they provide to act as cluster-admin.
+This restricts who can create Source and Apply type objects into a burdensome approval flow for cluster owners and provides
```
squaremo, line 40:
```suggestion
This restricts who can create Source and Apply type objects into a burdensome approval flow for cluster owners, and provides
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+poor delegation of self-service permissions which is a proven elements of good platforms.
```
squaremo, line 41: This sentence provoked a double-take -- it's not clear from the syntax whether _poor_ delegation of self-service permissions is a proven element of good platforms, or good delegation [...] would be. The meaning is obvious after the fact, but perhaps it could be rewritten to flow better.

squaremo, line 41:
```suggestion
poor delegation of self-service permissions which is a proven element of good platforms.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+#### user/group names
+All usernames impersonated by kustomize-controller and helm-controller for managing resources will be Namespaced relative to the `Kustomization`/`HelmRelease` object.
+This is similar to the mechanism Kubernetes uses for [ServiceAccount User names](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-subjects)).
+Kubernetes uses `system:serviceaccount:{{namespace}}:{{metadata.name}}` for ServiceAccounts.
+The Flux username format should be `flux:user:{{namespace}}:{{spec.user}}`.
+Alternatively flux could use the reserved prefix `system:flux:user:<...>` -- this would break the rules, but make it less likely for other systems to accidentally or maliciously use the FluxUser namespace when authenticating with kube-apiserver.
+The default `spec.user` can be "reconciler".
+A user field will be added to the Kustomization and HelmRelease spec.
+It enables a partial override for the namespaced username the controller uses for that particular `Kustomization`/`HelmRelease`.
```
squaremo, line 80: Does "partial override" mean that you can provide the name part of the username, but not the namespace? (or something else)

squaremo: (it might be clearer to spell out how the whole username is formed given the input, either way)

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+Flux controllers will also impersonate two Groups.
+The first is `flux:users` which is a generic group, analogous to kubernetes' `system:authenticated` -- binding against this group allows all reconcilers to access particular resources via RBAC.
+The second is `flux:users:{{namespace}}`, a namespace specific group which enables Namespace Owners to rolebind for reconcilers from another Namespace without mandating a particular username.
+Neither of these groups are optional or overrideable -- they are a given property of each reconciler's Namespace and identity.
+Note `users` is plural in the Group names, but not for a single User -- this matches kubernetes SA's.
+It could cause typos, but not matching k8s plural semantics could also be confusing. We could be silly and add both without notable consequence. 🚲🏠
+
+Impersonation of Users and Groups is done via API Headers that are permitted for the controller ServiceAccounts.
+- **Con**: controller SA's can impersonate any user or group, not just flux:user namespaced ones is explicitly specified via a ClusterRole (impractical,bad-ux)
```
squaremo, line 93: This sentence doesn't quite scan, for me.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+Using Namespaced Users and Groups provides a security boundary in the same way the ServiceAccount User names do.
+It's never allowed for a reconcile to impersonate a non-namespaced User or user from outside the Namespace of the particular reconcile resource.
+If a reconcile needs access to resources from another Namespace, the other Namespace can have a RoleBinding targeting the reconciler's differently namespaced User name.
+Alternatively, a ClusterRoleBinding on the reconciler's namespaced User name can grant access to all Namespaces.
+
+#### serviceAccount impersonation
+`spec.serviceAccountName` will remain supported, but impersonating Users is preferable because it reduces the attack surface within a Namespace.
+`ServiceAccounts` produce privileged token Secrets in the same Namespace often for use within `Pods`.
+It's possible to impersonate a `ServiceAccount` User name even when they don't exist, but it's possible to create the resource after the fact creating generated tokens without a change in the reconciler behavior.
```
squaremo, line 114:
```suggestion
It's possible to impersonate a `ServiceAccount` User name even when they don't exist; but it's possible to create the resource after the fact, creating generated tokens without a change in the reconciler behavior.
```

(I'm not sure I've grasped where the emphasis should go in this sentence, so my punctuation suggestion is a guess)

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+This doesn't happen with impersonated `Users` -- `Users` don't have any object representation. This prevents `Pods` in same Namespace as the `Kustomization`/`HelmRelease` from assuming the role of the privileged reconciler which would typically be an inappropriate privilege escalation.
+
+Impersonation of reconciler ServiceAccounts is done via API Headers that are permitted for the controller ServiceAccounts.
+The FluxUser Groups can still be added to requests for impersonated ServiceAccounts.
```
squaremo, line 118: Can or should? (if only "can", who decides?)

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+The token-fetching impersonation method currently implemented within kustomize-controller and helm-controller should be removed or flagged off.
```
squaremo, line 119: A sentence about what the "token-fetching impersonation" does would be welcome here, just to set up the points that follow.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+- *Pro*: No per-tenant crypto/random-number dependency due to SA-tokens
+- *Pro*: Controller SA's no longer require cluster-wide access to Secrets
+- *Pro*: Possible to configure controller to only have ServiceAccount impersonation via RBAC -- disable arbitrary User and Group Impersonation if controller code is untrusted (prevent use of `flux:user:*`)
+- **Con**: Single namespace installs of Flux controllers may depend on ClusterRoleBindings to flux-sa-impersonator which requires cluster-admin setup permissions.
+       Cannot(?) restrict ServiceAccount User impersonation to specific namespaces.
+       Compare to SA's and token Secrets which a Namespace admin has full access to *only within their Namespace*.
+       SA Token Impersonation allows single namespace installs to have less dependency on the cluster-admin (as long as Flux CRD's are already registered) while still keeping reconciler specific users.
+       For this reason, it may be nice to keep `--sa-token-impersonation` as a controller flag and support it with the single-namespace installation option, despite the faults /w using SA's.
```
squaremo, line 129:
```suggestion
       For this reason, it may be nice to keep `--sa-token-impersonation` as a controller flag and support it with the single-namespace installation option, despite the faults with using SA's.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+       It's possible to do some heuristic fallback here, but it could complicate things.
+
+If SA Token impersonation is enabled, impersonated FluxUser Groups are omitted from the request.
+ServiceAccounts still have a Group that you can rolebind against: `system:serviceaccounts:{{namespace}}`.
+
+#### defaulting
+Behavior must be clarified when both `user` and `serviceAccountName` are provided.
+It should probably be a validation or runtime error condition.
```
squaremo, line 137:
```suggestion
It should be a validation or runtime error condition.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+These filepaths need to be resolved to absolute paths before they are checked to prevent subversive path traversal via symlinks.
+
+KubeConfigs that set `users[].user.exec.command` or `users[].user.auth-provider.config.cmd-path` should be rejected unless all `[command|cmd-path]` values return
+an [`os.exec.LookPath()`](https://golang.org/pkg/os/exec/#example_LookPath) inside the `/kubeconfig-bin/` of the controller Pod filesystem.
+While executing the kubeConfig, the `$PATH` likely needs to be preserved in case the auth-helper needs to execute other POSIX binary dependencies.
+Restricting `[command|cmd-path]` to `/kubeconfig-bin/` allows the administrator an opt-in directory where they can add tools they trust/accept like `gcloud` and `aws-iam-authenticator` 
+while disallowing access to more dangerous binaries like `kubectl`. This mechanism is somewhat inspired by `/sbin` and `setuid` binaries.
+
+Using exec helpers is not an approach used by Cluster API.
+In addition to being insecure, it's not possible to distrubite controllers containing all of the necessary binaries to support all kubernetes platforms.
```
squaremo, line 168:
```suggestion
In addition to being insecure, it's not possible to distribute controllers containing all of the necessary binaries to support all kubernetes platforms.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+ClusterRoles should exist for the Source controller API's `flux-source-viewer`, `flux-gitrepo-viewer`, `flux-helmrepo-viewer`, `flux-bucket-viewer`, etc.
+FluxUsers should be bound explicitly against these ClusterRoles, either with RoleBindings (for single Namespaces) or ClusterRoleBindings.
+Cross-Namespace access of Sources can be enabled per Namespace or across the whole cluster via RoleBindings/ClusterRoleBindings to a FluxUser, Flux Group, or All FluxUsers.
+Kinds of sources can be differentiated, and individual Sources can be permitted per-User/Group through `resourceNames` within Role rules.
+The current API objects that use sourceRefs are `Kustomization` and `HelmChart`.
+`Kustomization` and `HelmRelease` have the `serviceAccountName` proposed `user` fields for impersonation.
+Both of these impersonation fields should also be added to `HelmChart` because reconciling one requires accessing other Sources.
+source-controller will need impersonation behavior implemented /w additional RBAC and any supporting flags/options.
```
squaremo, line 184:
```suggestion
source-controller will need impersonation behavior implemented with additional RBAC and any supporting flags/options.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+Alternatively, `HelmChart` could be moved into the helm-controller control loops instead of source-controller.
+It's not sufficient to gate the creation of `HelmCharts` based off of a `sourceRef` permission check in `HelmRelease`;
+RBAC rules can change at any time, so the source-controller API client should have the constraints of the impersonated User.
+
+If a kubeConfig is specified for a resource that specifies a sourceRef, a separate client matching the same User or SA from the remote cluster is used in the source cluster.
+This is sensible because the User for the remote cluster must match the source cluster's reconciling Namespace.
+If no User is specified for the remote cluster, the default FluxUser is impersonated to fetch the sourceRef from the source cluster.  
+
+
```
squaremo, line 193: Is it worth talking about transitive access to secrets via sources, here? E.g., should an impersonated user need access to a secret referenced by a source. If it's a non-issue, it'd be good to at least say why it's a non-issue, to cover that off.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+Kinds of sources can be differentiated, and individual Sources can be permitted per-User/Group through `resourceNames` within Role rules.
+The current API objects that use sourceRefs are `Kustomization` and `HelmChart`.
+`Kustomization` and `HelmRelease` have the `serviceAccountName` proposed `user` fields for impersonation.
+Both of these impersonation fields should also be added to `HelmChart` because reconciling one requires accessing other Sources.
+source-controller will need impersonation behavior implemented /w additional RBAC and any supporting flags/options.
+Alternatively, `HelmChart` could be moved into the helm-controller control loops instead of source-controller.
+It's not sufficient to gate the creation of `HelmCharts` based off of a `sourceRef` permission check in `HelmRelease`;
+RBAC rules can change at any time, so the source-controller API client should have the constraints of the impersonated User.
+
+If a kubeConfig is specified for a resource that specifies a sourceRef, a separate client matching the same User or SA from the remote cluster is used in the source cluster.
```
squaremo, line 189: If I understand this correctly, I can see it causing confusion -- you need the same user to be defined in both clusters, with quite different permissions (in the source cluster, it needs to see all the sources; in the remote cluster, it needs to create things in the destination namespace).

An alternative might be to name the remote user to impersonate (and change the structure of the CRD to make this obvious). But: additional complexity, certainly.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+Using exec helpers is not an approach used by Cluster API.
+In addition to being insecure, it's not possible to distrubite controllers containing all of the necessary binaries to support all kubernetes platforms.
+Cluster API providers instead schedule regular updates to kubeConfig Secret data which is more tenant friendly and dependency-free.
+
+Regardless, we must sanitize the kubeConfig, so it's easy to add a restricted, PATH-based allow-list that the flux-system owner can control if they choose to manage remote clusters in this way.
+
+
+#### sourceRefs
+Controllers should impersonate the reconciler User when fetching sourceRefs.
```
squaremo, line 175: Did you consider the effect of removing cross-namespace references? It would still be necessary to impersonate; but perhaps easier to reason about.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+Some reconciler ServiceAccounts may need `flux-source-viewer` bound to them.
+Others may prefer to update their serviceAccountNames to user.
+Migration/Upgrade documentation will be necessary.
+It's possible to build a tool-assisted migration that modifies resources in a git repo. (`flux migrate <git-repo-path>` ?)
+
+#### performance
+Flux controllers will be using many more Kubernetes client Objects, informers, and caches.
+Storing a map of Flux Users to clients/caches might be a good first step.
+kubeConfig client/caches are not stored currently, which was an explicit decision to support refreshing clients,
+this this doesn't have to be true for all kubeConfigs.
```
squaremo, line 251:
```suggestion
but this doesn't have to be true for all kubeConfigs.
```

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+  kind: ClusterRole
+  name: view
+subjects:
+  - kind: Group
+    name: flux:users
+```
+
+
+#### opening Source policies with RBAC
+Accessing Sources across Namespaces is denied by defaultl.
```
squaremo, line 454:
```suggestion
Accessing Sources across Namespaces is denied by default.
```

squaremo (commented):
This is very thorough and well thought through*, excellent work Leigh.

I've given suggestions for tightening it up; I don't have substantial disagreement with anything that's suggested here. I'll give my thoughts on the particular decisions in the main thread.

(*not a poet)

stefanprodan:
How are we going to address multi-tenancy when connecting to Git, container registries and key management services using IAM roles binding? If we bing the IAM roles of all tenants to the Flux controllers pods then a tenant can access another tenant's KMS keys, registries, etc.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+The Flux username format should be `flux:user:{{namespace}}:{{spec.user}}`.
+Alternatively flux could use the reserved prefix `system:flux:user:<...>` -- this would break the rules, but make it less likely for other systems to accidentally or maliciously use the FluxUser namespace when authenticating with kube-apiserver.
+The default `spec.user` can be "reconciler".
+A user field will be added to the Kustomization and HelmRelease spec.
+It enables a partial override for the namespaced username the controller uses for that particular `Kustomization`/`HelmRelease`.
+Namespace Owners can use the user override to optionally specify flux `Users` with different `Roles`/`RoleBindings` for managing each workload,
+similar to Pods specifying non-default serviceAccounts from the same Namespace.
+- **Con**: RoleBinding against a FluxUser requires specifying the namespace in the username string which can be hard to do in a generic way with kustomize -- would need to document a kustomize example
+
+Flux controllers will also impersonate two Groups.
```
jonathan-innis, line 85: I'm a bit unclear on who is adding the ClusterRoleBinding on the `flux:users` and `flux:users:{{namespace}}` groups. Is the onus on the end-user to create this ClusterRoleBinding after the controllers are deployed or is there some proposed automated process to create these bindings?

stefanprodan: Please see https://github.com/fluxcd/flux2-multi-tenancy#onboard-tenants

TLDR: there is a `flux create tenant` that creates the user or serviceaccount to be impersonated. At install time, the `flux bootstrap` command will setup the service account and RBAC for the controller pods to enable impersonation.

jonathan-innis:
💯 From the Microsoft Team. Looks good to us, thanks for the discussion @stealthybox

squaremo:
I would love to see this implemented. Without wanting to erect more barriers in the way of that happening, I think there are some lingering worries over how much RBAC manipulation it would push onto users. Is it possible to show some worked examples, e.g., here is how the RBAC would need to be set up for multi-tenancy (and each tenant), here is how it works if you don't care about multi-tenancy. @stefanprodan I think you had a Helm-related scenario you weren't sure about, too?

stealthybox:
> here is how the RBAC would need to be set up for multi-tenancy (and each tenant)

The proposal examples attempt to cover this lightly.
The gist of it is that whoever creates/administers a Namespace RoleBinds any tenant FluxUser from any other Namespace to `admin` or whatever Role they please.

> here is how it works if you don't care about multi-tenancy.

Give every FluxUser access to all Sources:
```yaml
---
kind: ClusterRoleBinding
metadata:
  name: flux-users-view-sources
roleRef:
  kind: ClusterRole
  name: flux-source-viewer
subjects:
  - kind: Group
    name: flux:users
```

Going even further...

Give every FluxUser `cluster-admin`:
```yaml
---
kind: ClusterRoleBinding
metadata:
  name: flux-users-cluster-admin
roleRef:
  kind: ClusterRole
  name: cluster-admin
subjects:
  - kind: Group
    name: flux:users
```

stealthybox:
The unfortunate bit is that Namespace FluxUsers have to be templated into their RoleBinding so that their namespace matches, so you can't just copy the same exact RoleBinding into every Namespace without some tool capable of doing that.

( RoleBindings have this built-in as a special case for ServiceAccounts )

https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/
Is able to copy the same object down to all child Namespaces, but it's not available by default in just any cluster.

From a practical sense, whoever sets up the tenants will likely just extend a Kustomize base that has the RoleBinding baked in.
They just patch the GitRepository and Kustomization path to start a new tenant.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+#### controller serviceAccounts
+The controller ServiceAccounts are far overprivileged for Flux API operations.
+They should have a ClusterRoleBinding (or RoleBinding for single Namespace installs) for necessary Flux API Kinds and other needed objects such as ConfigMaps/Secrets + a separate ClusterRoleBinding for User Impersonation.
+
+#### user/group names
+All usernames impersonated by kustomize-controller and helm-controller for managing resources will be Namespaced relative to the `Kustomization`/`HelmRelease` object.
+This is similar to the mechanism Kubernetes uses for [ServiceAccount User names](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-subjects)).
+Kubernetes uses `system:serviceaccount:{{namespace}}:{{metadata.name}}` for ServiceAccounts.
+The Flux username format should be `flux:user:{{namespace}}:{{spec.user}}`.
```
simongottschlag, line 76: What use cases do we see for making the name configurable? I can't see anything other than making it possible to lower the impact of something by decision of the creator.

If tenant-A grants tenant-B access (edit role) to the whole namespace tenant-A for reconciler but then grants another role (pod-deleter) to the user pod-deleter, then it would be up to tenant-B to know and decide on which one to use.

If there isn't a strong enough case for these scenarios - maybe the user spec isn't needed and a default one is always used?

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+  verbs: ["impersonate"]
+```
+
+Using Namespaced Users and Groups provides a security boundary in the same way the ServiceAccount User names do.
+It's never allowed for a reconcile to impersonate a non-namespaced User or user from outside the Namespace of the particular reconcile resource.
+If a reconcile needs access to resources from another Namespace, the other Namespace can have a RoleBinding targeting the reconciler's differently namespaced User name.
+Alternatively, a ClusterRoleBinding on the reconciler's namespaced User name can grant access to all Namespaces.
+
+#### serviceAccount impersonation
+`spec.serviceAccountName` will remain supported, but impersonating Users is preferable because it reduces the attack surface within a Namespace.
```
simongottschlag, line 112: Could it be an option to remove it?

If the controller starts up without impersonation permissions, it can fall back to using the pod service account ("namespaced mode" or cluster admin, up to how the SA permissions are configured).

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+       Compare to SA's and token Secrets which a Namespace admin has full access to *only within their Namespace*.
+       SA Token Impersonation allows single namespace installs to have less dependency on the cluster-admin (as long as Flux CRD's are already registered) while still keeping reconciler specific users.
+       For this reason, it may be nice to keep `--sa-token-impersonation` as a controller flag and support it with the single-namespace installation option, despite the faults /w using SA's.
+       It's possible to do some heuristic fallback here, but it could complicate things.
+
+If SA Token impersonation is enabled, impersonated FluxUser Groups are omitted from the request.
+ServiceAccounts still have a Group that you can rolebind against: `system:serviceaccounts:{{namespace}}`.
+
+#### defaulting
+Behavior must be clarified when both `user` and `serviceAccountName` are provided.
```
simongottschlag, line 136: If we don't need different usernames per namespace and remove the ability to impersonate SAs but use the pod SA - then maybe this isn't even needed?

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+# example singleton style API
+spec:
+  user:
+    kind: ServiceAccount  # defaults to User
+    name: reconciler-svc
+```
+If `user` and `serviceAccountName` are unset/empty-strings, the default FluxUser (proposed: `reconciler`) is used.
+Alternatively, we could use [CRD defaulting](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting), but this puts other considerations on the design with detecting empty values.
+
+#### kubeConfig
```
simongottschlag, line 149: Would it be possible to do a kubeConfigTemplate the same way we have constraintTemplates with OPA-Gatekeeper?

Then maybe we could limit what is actually run inside the pod in a safer manner?

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+Regardless, we must sanitize the kubeConfig, so it's easy to add a restricted, PATH-based allow-list that the flux-system owner can control if they choose to manage remote clusters in this way.
+
+
+#### sourceRefs
+Controllers should impersonate the reconciler User when fetching sourceRefs.
+This allows for users to open up Source access policies via RBAC, rather than relying on a Gatekeeper or Kyverno installation to restrict default Flux behavior.
+ClusterRoles should exist for the Source controller API's `flux-source-viewer`, `flux-gitrepo-viewer`, `flux-helmrepo-viewer`, `flux-bucket-viewer`, etc.
+FluxUsers should be bound explicitly against these ClusterRoles, either with RoleBindings (for single Namespaces) or ClusterRoleBindings.
+Cross-Namespace access of Sources can be enabled per Namespace or across the whole cluster via RoleBindings/ClusterRoleBindings to a FluxUser, Flux Group, or All FluxUsers.
```
simongottschlag, line 179: Could there be a point to not have cross-namespace references at all and instead allow users to utilize tools like kubed: https://github.com/appscode/kubed

I still think that Cluster-scoped resources should be used instead of doing cross-namespace refs for namespaced resources: https://github.com/fluxcd/flux2/discussions/497

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+kubeConfig client/caches are not stored currently, which was an explicit decision to support refreshing clients,
+this this doesn't have to be true for all kubeConfigs.
+
+We have done some performance testing in the past, but it would be good to have a strategy/tools to help us
+understand performance regressions.
+
+## Decisions
+Direction should be decided on the following:
+
+- [ ] `flux:user` vs. `system:flux:user` User prefix
```
simongottschlag, line 259: flux:user

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+- [ ] `flux:users` vs. `flux:user` Group prefix
```
simongottschlag, line 260: flux:users

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+- [ ] default FluxUser name (proposed: "reconciler")
```
simongottschlag, line 261: reconciler or maybe system? Should have a status that writes out the full name.

simongottschlag (commented):
A few thoughts from me. I am starting to like the impersonation implementation.

Great write up! 😁👍

docs/proposals/secure-impersonation.md
stefanprodan: This is not an option IMO, many users are running Flux in production with this setting, not to mention that flux1 multi- tenancy was based on SAs, we’ll be cutting off all these users which I find unacceptable.

docs/proposals/secure-impersonation.md
stefanprodan: Not being able to share sources like HelmRepositories has serious implications to resource usage and cost. If users have to register e.g. Bitnami on each namespace, then source-controller will need many GB of memory, not to mention all the egress traffic people have to pay for to pull the same artifacts into each namespace.

squaremo: Is the resource cost because of the model, or because of the implementation? (or to ask another way: is it possible to rewrite the controllers in such a way that HelmRepositories are not duplicated in memory, without changing the visible behaviour?)

stefanprodan: @squaremo I don’t think this is doable, we use the Helm getter package that loads the index into memory. We’ll need some kind of locking between parallel reconciliations of different charts and that seems like a controller-runtime anti pattern, also if you have multiple sources pointing at the same helm repo, each will run on its own interval. Maybe @hiddeco knows a way a around this.

hiddeco: I have been summoned.

The indexes are put into memory mostly when a `HelmChart` is reconciled. This loads the index into memory so the latest chart can be looked up based on the SemVer selector, and I see very little roam to work around this in a sane way. I do however think that the resource consumption will not increase much, as there will always be at most 8 repository indexes in memory, given we have a worker limit of 4 per reconciler.

My personal concern around resource consumption is more with the in-memory actions we perform on Git repositories, people have started to put their manifests next to their applications, applications which in some cases include e.g. training models or other large files, causing the source-controller to OOM kill.

> I still think that Cluster-scoped resources should be used instead of doing cross-namespace refs for namespaced resources

Have suggested this to Stefan before, and think this is indeed the way to deal with it given how Kubernetes wants you to work, but there were some concerns which I can not recall at the moment.

stefanprodan: Secrets are namespaced so a cluster scope source will need a secretRef.namespace. I’m not against having all the sources kinds duplicated as ClusterGitRepository/ClusterHelmRepository/ClusterBucket but I don’t see how it improves the multi-tenancy story, you would need a cluster role binding for the sources and role binding to read the secrets from different namespaces, doesn’t make it worse?

docs/proposals/secure-impersonation.md
squaremo: or `flux:user:default`

stealthybox: My only concern /w `default` is users conflating it with the default SA in the same NS.
Otherwise it's a pretty good name.

docs/proposals/secure-impersonation.md
simongottschlag: @stefanprodan @stealthybox a lot of the changes will already need a migration path - if this makes the product easier and more secure, wouldn't it be worth it? We won't be able to do it after the fact.

Do we see a reason maintaining two ways of doing multi-tenancy? Seems like a lot of work keeping track of it just because of compatibility.

docs/proposals/secure-impersonation.md
simongottschlag: My opinion is that it's better to reference a namespaced resource from a cluster scoped resource instead of breaking how namespaced resources are supposed to work.

The resource consumption, if it becomes a problem, can be solved by creating a Cluster scoped resource (but then maybe up to the fleet admin to do it).

We want to make it easy to do the right / secure thing.

docs/proposals/secure-impersonation.md
phillebaba: I think there should be a deprecation path but not removed immediately. Removing serviceaccountname immediately will make it difficult for existing users to migrate. One solution could be to look into adding some deprecation notice to convince users to migrate over to users.

simongottschlag:
@stealthybox @stefanprodan @squaremo @phillebaba 

Suggestion from yesterday:

# Source Reflection

## Custom Resource Definitions

### Source

```YAML
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: example-repo
  namespace: tenant-a-namespace
spec:
  interval: 1m
  url: https://github.com/example/repo
  ref:
    branch: main
  reflections:
  - kind: SourceReflection
    namespace: tenant-b-namespace
```

### Source Reflection

```YAML
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: SourceReflection
metadata:
  name: tenant-a-namespace-example-repo
  namespace: tenant-b-namespace
spec:
  sourceName: example-repo
  sourceNamespace: tenant-a-namespace
```

### Kustomization

```YAML
kind: Kustomization
metadata:
  name: example-kustomization
  namespace: tenant-b-namespace
spec:
  sourceRef:
    kind: SourceReflection
    name: tenant-a-namespace-example-repo
``` 

## Role Based Access Control

### Cluster Role: View Source Reflection

More information about [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).

```YAML
kind: ClusterRole
metadata:
  name: flux-source-reflection-view
  labels:
    rbac.authorization.k8s.io/aggregate-to-view: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
rules:
- apiGroups: ["source.toolkit.fluxcd.io"]
  resources: ["sourcereflection"]
  verbs: ["get", "list", "watch"]
```

stefanprodan:
@simongottschlag how would source-controller create the SourceReflection objects in other namespaces? If we give it cluster access how would you prevent a tenant from creating reflections and "DDOS-ing" other tenants namespaces with bogus objects?

simongottschlag:
> @simongottschlag how would source-controller create the SourceReflection objects in other namespaces? If we give it cluster access how would you prevent a tenant from creating reflections and "DDOS-ing" other tenants namespaces with bogus objects?

My thought was a `ClusterRole` and `ClusterRoleBinding` giving it access to all verbs in all namespaces. What's the difference in a security context of "DDOS-ing" your own namespace with 100s or 1000s of GitRepositories compared to creating references of them in other namespaces? I don't see the DDoS issue since it will affect the same Kubernetes API and the same source-controllers.

stefanprodan:
>  I don't see the DDoS issue since it will affect the same Kubernetes API and the same source-controllers.

So as a tenant, you don't see anything wrong with objects popping up in your namespace made by other tenants?

simongottschlag:
> >  I don't see the DDoS issue since it will affect the same Kubernetes API and the same source-controllers.
> 
> 
> 
> So as a tenant, you don't see anything wrong with objects popping up in your namespace made by other tenants?

@stefanprodan I actually don't see a problem with it.

If the resource is explicitly used for this, then it would be expected that it's being populated from the outside. If a cluster has problems with tenants abusing something, then it's most likely not something we should have a technical solution for but most likely and organizational one.

If we think about the security, and DDoS is the biggest we can come up with - then it should be fine there as well since we don't introduce something new (or at least I think so) since it's a shared control plane and DOS of ones own namespace will cause just as much issues as DOS in another.

By using reflections, we also make what can be used visually compared to the RBAC case. You will easily see what is being shared to you.

We could also consider using something like ClusterSourceReflection to share something to the whole cluster. Haven't thought that through completely yet though.

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+
+If a kubeConfig is specified for a resource that specifies a sourceRef, a separate client matching the same User or SA from the remote cluster is used in the source cluster.
+This is sensible because the User for the remote cluster must match the source cluster's reconciling Namespace.
+If no User is specified for the remote cluster, the default FluxUser is impersonated to fetch the sourceRef from the source cluster.  
+
+
+#### notifications
+notification-controller needs to be able to impersonate FluxUsers and ServiceAccounts.
+`Recievers` specify `resources` which are Sources -- `Alerts` specify `eventSources` which can be Sources or Applies.
+Both impersonation fields should also be added to `Recievers` and `Alerts` to allow for the same kind of access polices possible with sourceRefs.
```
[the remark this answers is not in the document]
stealthybox: providerRefs are in the same Namespace, so this seems okay to me?
There isn't any reason we couldn't also guard same-Namespace refs wit the same SA though.
It's purely a user-experience/granular-permissions thing.

stealthybox:
Just noting from our discussion yesterday that I had an idea of combining Source Access-Lists and the RBAC Impersonation solutions for Source Policy.

If source-controller has permission to create RoleBindings, it can convert the in-Source Namespace-Selector/Access-List into RoleBindings for the appropriate FluxGroups.
These RoleBindings could use a pre-created ClusterRole, and would be managed as sub-objects of the Source.

You get the benefits of both systems and more:
- easy API embedding within Source obj
- high-level dynamic behavior like label selectors
- source-controller implements the security constraints with native k8s policy,  kustomize/helm-controller don't have to implement source-specific accessLists client-side.
- k8s audit log shows granular access /w errors  (no immediate burden to have access log in Source status/Events)
- behavior is completely opt-in, deny-all by default
- platform-admins can still implement their own cluster-wide policies with RBAC RoleBindings/ClusterRoleBindings directly without interfering with the mechanism (can easily allow-all for the whole NS or cluster)
- future improvements to RBAC improve Source Policy

I think it's the right approach to use RBAC as the primitive for access within the client controllers like helm/kustomize-controller.
If we agree the higher-level behavior is something we want, it can be implemented using the Source spec and some automated RBAC.

Similarly, default, per-Namespace RoleBindings for Kustomizations and HelmReleases could really alleviate the UX and migration issues we're all worried about for this change while still giving platform-admins complete control to lock everything down.

stefanprodan:
We could consider implementing an access list on the source side such as:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: podinfo
  namespace: flux-system
spec:
  interval: 1m
  url: https://github.com/stefanprodan/podinfo
  accessFrom:
    - namespaceSelector:
        matchLabels:
          tenant: one
    - namespaceSelector:
        matchLabels:
          tenant: two
``` 

All Flux reconcilers such as kustomize-controller, helm-controller and image-automation-controller would need a RO cluster role to query the API for namespaces to enforce the policy on the called side.

Another option is for source-controller to generate RBAC and add role bindings for the Flux user in each namespace to grant access to specific sources by name.

squaremo:
> We could consider implementing an access list on the source side [...]

I like the approach of granting access in the source object itself. I think this would fit well with the impersonation scheme: the source-controller sorts out the RBAC according to the shared access as declared, and so long as controllers use impersonation, they will enforce sharing correctly.

stealthybox:
Since the Source specifies the subject list, you could add more than just `FluxUser`'s and `FluxGroup`'s.
Listing arbitrary users would allow you to add `dev-team1@example-company.com` to the Source access list.
This would grant read access to that User for this Source in the same spirit as the reconciler workloads.

When `dev-team1@example-company.com` does a `flux get source git -A`, they will see all of the GitRepos they have access to.
This is not the same as showing exactly what a reconciler has access to, but it's a decent way for an admin to show what a reconciler has access to.

An additive solution is to give `dev-team1@example.com` a ClusterRoleBinding to a ClusterRole that allows them to only impersonate their specific `FluxGroup` (`flux:users:dev-team1`) for administrative purposes.

If the platform-admin has good discipline (and we have good examples), they will actually split up the FluxUser RoleBindings into `read-write` + `read-only`, so it would be possible to give `dev-team1@example.com` read-only access.

We could build this into the proposal so that every FluxUser actually has a read-only component as well, but it adds another Subject that folks need to consider when writing their RBAC.

Unfortunately, I'm not aware of a way to impersonate a User but tell the APIServer to drop certain verbs.
Maybe that feature exists? If not, it's a great KEP topic.

docs/proposals/secure-impersonation.md
maxgio92: Is this supposed to be the default impersonated user the specific controller impersonate when reconciling? Or I am misunderstood something?
Thanks

docs/proposals/secure-impersonation.md
```diff
@@ -0,0 +1,636 @@
...
+## Goals
+
+Flux should remain easy to install but promote secure practices in its default configuration.
+Flux's security model should allow a cluster administrator to delegate self-service management of
+Sources, Applies, Webhooks, Automations, and Notifications to Namespace owners/users without having to review every detail for
+compliance to security policies.
+Tenants or Namespace owners should be able to act freely without fear that their changes will overreach their permissions.
+
+Flux APIs must structurally prevent privilege escalation and arbitrary code execution.
+Flux APIs should be congruent with Kubernetes core API's security model.
```
pjbgf, line 53: ** Assuming the proposal is geared towards the Namespace-as-a-Service [tenancy model](https://kubernetes.io/blog/2021/04/15/three-tenancy-models-for-kubernetes/). **

The challenge here from a security perspective is providing _privilege escalation and arbitrary code execution_ prevention guarantees to flux's users. 
The proposed changes focuses in providing isolation and least privileges at control plane level only.
Privesc routes from Pod to Host, would not be contained. In the same way as lateral movement through to other tenants would be open for abuse by a malicious tenant.  

At the same time, trying to enforce those levels of isolation on behalf of flux's users may become disruptive and break compatibility with existing workloads. That would entail becoming more opinionated around some security primitives (i.e. Pod Security Standards, workload scheduling, etc), which feels beyond flux's scope.
So I am not sure it is worth pursuing _privilege escalation and arbitrary code execution_ prevention as a goal.

In my opinion the proposal could highlight as a goal making flux _optimised_ for multi-tenancy, focusing on a least privileged security model at *control plane* and better resource utilisation instead.

stefanprodan: Good point @pjbgf, I think we should limit the scope to “resources reconciliation” and be explicit about what this means. The model proposed here is about giving cluster admins a way to define polices that restrict what a tenant source (Git, Helm, Bucket, etc) can contain and thus limiting the operations that tenants can perform on a cluster. Flux can ensure that a tenant can deploy applications in certain namespaces and only in those. Flux can’t ensure that the applications deployed are themselves secure, this is in scope for network policy, PsP, seccomp profiles, etc.

stefanprodan:
Closing this in favour of:
- https://github.com/fluxcd/flux2/pull/2092
- https://github.com/fluxcd/flux2/pull/2093
- https://github.com/fluxcd/flux2/pull/2086

Starting with Flux 0.26 we allow platform admins to enforce tenant isolation as described in the [multi-tenancy lockdown documentation](https://fluxcd.io/docs/installation/#multi-tenancy-lockdown).

Thanks @stealthybox for all your proposals and to everyone that contributed with ideas. Please let's continue the discussions started here on the RFC PRs.

PR #595, opened 2020-12-12
Add cluster-domain option for bootstrap command

L3o-pold:
Since https://github.com/fluxcd/source-controller/pull/233 the bootstrap command will use `--storage-adv-addr=source-controller.$(RUNTIME_NAMESPACE).svc.cluster.local.` as default option for the source controller.

But for some users, the cluster domain is not `cluster.local` and this will result in breaking their source controller each time they run the `bootstrap` command (for upgrading their flux instance for example).


stefanprodan (approved):
LGTM

Thanks @L3o-pold

PR #599, opened 2020-12-14
Add image automation guide

stefanprodan:


docs/guides/image-update.md
```diff
@@ -0,0 +1,341 @@
...
+* CI: build and push a container image tagged as `registry.domain/org/app:v1.0.1`
+* CD: pull the latest image metadata from the app registry (Flux image scanning)
+* CD: update the image tag in the app manifest to `v1.0.1` (Flux cluster to Git reconciliation)
+* CD: deploy `v1.0.1` to production clusters (Flux Git to cluster reconciliation)
+
+## Prerequisites
+
+You will need a Kubernetes cluster version 1.16 or newer and kubectl version 1.18.
+For a quick local test, you can use [Kubernetes kind](https://kind.sigs.k8s.io/docs/user/quick-start/).
+Any other Kubernetes setup will work as well though.
```
hiddeco, line 32:
```suggestion
Any other Kubernetes setup will work as well, though.
```

docs/guides/image-update.md
```diff
@@ -0,0 +1,341 @@
...
+
+In order to follow the guide you'll need a GitHub account and a
```
hiddeco, line 34: I would add a block below this with something like:

```suggestion
!!! hint "GitLab and other Git platforms"

The `--components-extra` flag used below is available to the `flux install` and `flux bootstrap gitlab` commands as well, see [...]
```

to guide people towards using it without GitHub.

docs/guides/image-update.md
```diff
@@ -0,0 +1,341 @@
...
+  --owner=$GITHUB_USER \
+  --repository=flux-image-updates \
+  --branch=main \
+  --path=clusters/my-cluster \
+  --token-auth \
+  --personal
+```
+
+The bootstrap command creates a repository if one doesn't exist, and commits the manifests for the
+Flux components to the default branch at the specified path. Then it configures the target cluster to
```
hiddeco, line 61:
```suggestion
Flux components to the default branch at the specified path. It then configures the target cluster to
```

docs/guides/image-update.md
```diff
@@ -0,0 +1,341 @@
+# Automate image updates to Git
+
```
hiddeco, line 2: Would maybe add an alpha version notice/warning here.

hiddeco (approved):
Awesome, thank you Stefan :grapes:

docs/guides/image-update.md
stefanprodan: I've added the `--components-extra` flag to the install guide.

Issue #591, opened 2020-12-11
can't connect to azure devops repo


Edblakejitsuin:
Trying to hook this up with Azure Devops repo:

Docs say use --git-implementation=libgit2 to connect to azure devops repo.
This flag doesnt seem to exist... :
unknown flag: --git-implementation

attempting witout the flag fails:

flux create source git flux-system \
  --url=ssh://ssh.dev.azure.com/<project>/<repo> \
  --branch=master \
  --interval=1m

► applying secret with keys
✔ authentication configured
✚ generating GitRepository source
► applying GitRepository source
✗ no matches for kind "GitRepository" in version "source.toolkit.fluxcd.io/v1beta1"

stefanprodan:
You need to use the latest version v0.5.0 and follow the install docs, your current installation will not work, so you need to rerun install then continue with the setup.

Edblakejitsuin:
thanks, I've upgraded to 0.5.2 and able to use flag --git-implementation=libgit2.

However, still getting errors:

► collecting preferred public key from SSH server
✔ collected public key from SSH server:
ssh.dev.azure.com ssh-rsa 
► applying secret with keys
✔ authentication configured
✚ generating GitRepository source
► applying GitRepository source
✗ no matches for kind "GitRepository" in version "source.toolkit.fluxcd.io/v1beta1"

documents say use; '--ssh-key-algorithm=ecdsa' but this doesn't work as Azure Devops expects ssh-rsa, so I am using these options:

flux create source git flux-system \
  --url=ssh://ssh.dev.azure.com/<project>/<repo> \
  --branch=master \
  --interval=1m \
  --git-implementation=libgit2

(error from Azure Devops: Invalid key: Valid keys will start with "ssh-rsa".)

Any ideas?

stefanprodan:
`no matches for kind "GitRepository" in version "source.toolkit.fluxcd.io/v1beta1"` means that you haven't run `flux install` and there is no flux controller running on your cluster.

Edblakejitsuin:
sorry yes, I realised that shortly after posting. here's my next issue after running the below:

flux create source git flux-system \
  --url=ssh://ssh.dev.azure.com/org/project/repo/ \
  --branch=master \
  --interval=1m \
  --git-implementation=libgit2

► collecting preferred public key from SSH server
✔ collected public key from SSH server:
► applying secret with keys
✔ authentication configured
✚ generating GitRepository source
► applying GitRepository source
[...]

PR #603, opened 2020-12-14
Fix Azure DevOps SSH URL in docs

stefanprodan:
Fix: #591

### DIFF
--- a/docs/guides/installation.md
+++ b/docs/guides/installation.md
@@ -292,12 +292,15 @@ Create a `GitRepository` object on your cluster by specifying the SSH address of
 
 ```sh
 flux create source git flux-system \
-  --url= ssh://<host>/<org>/my-repository \
+  --url=ssh://git@<host>/<org>/<repository> \
   --ssh-key-algorithm=ecdsa \
   --ssh-ecdsa-curve=p521 \
   --branch=master \
   --interval=1m
 ```
+
+You will be prompted to add a deploy key to your repository.
+If you don't specify the SSH algorithm, then `flux` will generate an RSA 2048 bits key.
 
 !!! hint "Azure DevOps"
     Azure DevOps requires a non-default Git implementation (`libgit2`) to be enabled, so that the Git v2 protocol is supported.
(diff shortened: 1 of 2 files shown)


docs/guides/installation.md
```diff
@@ -292,34 +292,34 @@ Create a `GitRepository` object on your cluster by specifying the SSH address of
...
-  --ssh-key-algorithm=ecdsa \
-  --ssh-ecdsa-curve=p521 \
-  --branch=master \
-  --interval=1m \
-  --git-implementation=libgit2
-```
-
-You will be prompted to add a deploy key to your repository.
-If you don't specify the SSH algorithm, then `flux` will generate an RSA 2048 bits key.
+    If you are using Azure DevOps you need to specify a different git implementation than the default:
```
hiddeco, line 312:
```suggestion
    If you are using Azure DevOps you need to specify a different Git implementation than the default:
```

(Git is a name)

docs/guides/installation.md
```diff
@@ -292,34 +292,34 @@ Create a `GitRepository` object on your cluster by specifying the SSH address of
...
+    
+    ```sh
+    flux create source git flux-system \
+      --git-implementation=libgit2 \
+      --url=git@ssh.dev.azure.com/v3/org/project/repository \
+      --branch=master \
+      --interval=1m
+    ```
+    
+    Note that the SSH URL can't contain `:`, instead of `ssh.dev.azure.com:v3` use `ssh.dev.azure.com/v3`.
```
hiddeco, line 322:
```suggestion
	Note that unlike `git`, Flux does not support the  ["shorter" scp-like syntax for the SSH protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_ssh_protocol) (e.g. `ssh.dev.azure.com:v3`). Use the [RFC 3986 compatible syntax](https://tools.ietf.org/html/rfc3986#section-3) instead: `ssh.dev.azure.com/v3`.
```

docs/guides/installation.md
```diff
@@ -292,34 +292,37 @@ Create a `GitRepository` object on your cluster by specifying the SSH address of
...
+      --git-implementation=libgit2 \
+      --url=git@ssh.dev.azure.com/v3/org/project/repository \
+      --branch=master \
+      --interval=1m
+    ```
+
+    Note that unlike `git`, Flux does not support the
+    ["shorter" scp-like syntax for the SSH protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_ssh_protocol)
+    (e.g. `ssh.dev.azure.com:v3`).
+    Use the [RFC 3986 compatible syntax](https://tools.ietf.org/html/rfc3986#section-3) instead: `ssh.dev.azure.com/v3`.
```
hiddeco, line 325: Is the new indent intended? If so, does this render as expected?

stefanprodan: The indentation is need because I want it to in the Azure box.

hiddeco (approved):

PR #604, opened 2020-12-14
Add tests for CLI flags

hiddeco:
This includes various bug fixes, especially around the area of missing
names for `<kind>/<name>` formats.


stefanprodan (approved):
LGTM 💯

PR #619, opened 2020-12-17
Add note about deploy key linked to pat

phillebaba:
Adds a hint regarding the fact that the deploy key created by the bootstrap command is linked to the personal access token used to authenticate. This is important as users must be warned so that they do not just remove their personal access tokens.

### DIFF
--- a/docs/guides/installation.md
+++ b/docs/guides/installation.md
@@ -118,6 +118,12 @@ flux bootstrap github \
   --personal
 ```
 
+!!! hint "Deploy Key"
+    The bootstrap command creates a ssh key which it stores as a secret in the
+    Kubernetes cluster. The key is also used to create a deploy key in the GitHub
+    repository. The new deploy key will be linked to the personal access token used
+    to authenticate. Removing the personal access token will remove the deploy key.
+
 Run the bootstrap for a repository owned by a GitHub organization:
 
 ```sh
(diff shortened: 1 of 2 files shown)


stefanprodan:
@phillebaba please move this in the installation guide under the GitHub section.

phillebaba:
Done and rebased.

docs/guides/installation.md
```diff
@@ -108,6 +108,12 @@ Export your GitHub personal access token as an environment variable:
 export GITHUB_TOKEN=<your-token>
 ```
 
+!!! hint "Deploy Key"
+    The bootstrap command creates a ssh key which it stores as a secret in the
+    Kubernetes cluster. The key is also used to create a deploy key in the GitHub
+    repository. The new deploy key will be linked to the personal access token used
+    to authenticate. Removing the personal access token will remove the deploy key.
+
 Run the bootstrap for a repository on your personal GitHub account:
```
stefanprodan, line 117: Please move after the first bootstrap command

stefanprodan (changes requested):

docs/guides/installation.md
phillebaba: done

stefanprodan (approved):
LGTM

Thanks @phillebaba 🍰

PR #641, opened 2021-01-01
Link docs how to get zsh/fish/ps completion to work in getting-started

jonaskello:
I tried to make completions work in zsh by just adding the same code as for the bash example (switching bash for zsh) but it did not work. When I googled and dug deeper I finally found the answer was already documented in the deeper docs here:

https://github.com/fluxcd/flux2/blob/main/docs/cmd/flux_completion_zsh.md

The command in there works if I add it to my .zshrc file.

I think linking to these deeper docs may prevent others from just assuming it will work the same in zsh.


hiddeco (commented):
Thank you for submitting this PR, and sorry it took so long to get to it.

Can you please sign-off your commit? Without it we aren't able to merge.

jonaskello:
@hiddeco Seems good to go now?

stefanprodan (approved):
LGTM

Thanks @jonaskello

PR #647, opened 2021-01-04
Validates components set

### LINKED ISSUE #644: Validate flux {install,bootstrap} flags --components and --components-extra
The components available are a known set, so the values supplied to both of these flags could be checked.
(first mentioned in https://github.com/fluxcd/flux2/pull/580)

SomtochiAma:
Fixes #644 

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -104,6 +104,11 @@ func bootstrapValidate() error {
 			return fmt.Errorf("component %s is required", component)
 		}
 	}
+
+	if err := utils.ValidateComponents(components); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/cmd/flux/install.go
+++ b/cmd/flux/install.go
@@ -110,6 +110,10 @@ func installCmdRun(cmd *cobra.Command, args []string) error {
 
 	components := append(installDefaultComponents, installExtraComponents...)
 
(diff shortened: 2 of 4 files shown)


cmd/flux/bootstrap.go
```diff
@@ -58,6 +58,7 @@ var (
 	bootstrapArch               = flags.Arch(defaults.Arch)
 	bootstrapLogLevel           = flags.LogLevel(defaults.LogLevel)
 	bootstrapRequiredComponents = []string{"source-controller", "kustomize-controller"}
+	bootstrapAllComponents      = []string{"source-controller", "kustomize-controller", "helm-controller", "notification-controller", "image-reflector-controller", "image-automation-controller"}
```
stefanprodan, line 61: Please see https://github.com/fluxcd/flux2/blob/main/pkg/manifestgen/install/options.go#L44 we could reuse the array from that package instead of duplicating the components name.

SomtochiAma: It seems the array is incomplete and it seems like updating it will install the extra components by default.

stefanprodan: You could add the extra components as field in manifestgen then merge the 2 array both here and in the install cmd.

stefanprodan (approved):
LGTM 💯 

Please rebase with the main branch

PR #654, opened 2021-01-06
Update helm release guide

phillebaba:
This changes the documentation regarding how to configure HelmRepositories with cloud storage services.

The old documentation recommended using the Bucket resource, which has the downside that source-controller will download the whole bucket. Chartmuseum seems like a  better option as it is a helm project that has added support for a bunch of different backends. This is a huge benefit as it may reduce the amount of feature requests for source-controller to support different storage backends.

Fixes fluxcd/source-controller/issues/249
Fixes fluxcd/source-controller/issues/206


stefanprodan (approved):
LGTM

Thanks @phillebaba 🍰

Issue #631, opened 2020-12-21
kustomization path not found: stat /tmp/flux-system022185384/clusters\staging: no such file or directory


xiemeilong:
fluxcd version: v0.5.7
os: window 10

```
flux bootstrap gitlab --cluster-domain xxx.local --repository git-ops --hostname code.xxxx.com --token-auth %GITLAB_TOKEN% --branch master --owner devops --path clusters/staging
```

```
► connecting to code.xxxxx.com
✔ repository cloned
✚ generating manifests
✔ components are up to date
► installing components in flux-system namespace
namespace/flux-system unchanged
customresourcedefinition.apiextensions.k8s.io/alerts.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/buckets.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/gitrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmcharts.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmreleases.helm.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/kustomizations.kustomize.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/providers.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/receivers.notification.toolkit.fluxcd.io configured
role.rbac.authorization.k8s.io/crd-controller-flux-system unchanged
rolebinding.rbac.authorization.k8s.io/crd-controller-flux-system unchanged
clusterrolebinding.rbac.authorization.k8s.io/cluster-reconciler-flux-system unchanged
service/notification-controller unchanged
service/source-controller unchanged
service/webhook-receiver unchanged
deployment.apps/helm-controller configured
deployment.apps/kustomize-controller configured
deployment.apps/notification-controller configured
deployment.apps/source-controller configured
networkpolicy.networking.k8s.io/allow-scraping unchanged
networkpolicy.networking.k8s.io/allow-webhooks unchanged
networkpolicy.networking.k8s.io/deny-ingress unchanged
deployment "source-controller" successfully rolled out
deployment "kustomize-controller" successfully rolled out
deployment "helm-controller" successfully rolled out
[...]

Issue #645, opened 2021-01-04
ManifestFile path error on Windows


bibAtdev:
## Issue description
On Windows 10 from [sync.go](https://github.com/fluxcd/flux2/blob/main/pkg/manifestgen/sync/sync.go) generated ManifestFiles like gotk-sync.yaml are containing ' \ ' instead of ' / '.
This most likely is caused by the usage of the  `path/filepath` instead of `path.Join` package, as it is OS aware.

## Technical Setup
|Artifact|Version|
|--|--|
|Windows 10| Windows 10 Build 19041.685 |
|Terraform | v0.14.3 |
|Terraform Flux Provider| v0.0.8 |

### Prerequisits

- Terraform installed
- [Terraform Flux Provider install module](https://github.com/fluxcd/terraform-provider-flux/tree/main/examples/install) successfully executed

### Commands

1.  Clone the Github [example ](https://github.com/fluxcd/terraform-provider-flux/tree/main/examples/github) from the Terraform Flux Provider repository to a Windows based computer.
2. Open a command line and navigate into the cloned repository
3. Set your Github credentials and token via environment variables, i.e. TF_VAR_github_owner AND TF_VAR_github_token
4. Execute `terraform init`
4. Execute `terraform apply -auto-approve`

## What's the expected result?

- RepoName/TargetPath/flux-system/gotk-sync.yaml
- RepoName/TargetPath/flux-system/kustomization.yaml

## What's the actual result?

- RepoName\TargetPath\flux-system\gotk-sync.yaml
- RepoName\TargetPath\flux-system\kustomization.yaml

## Suggested solution

- replace `path/filepath` with `path.Join` in [sync.go line 96](https://github.com/fluxcd/flux2/blob/main/pkg/manifestgen/sync/sync.go#L96) or replace ' \ ' with ' / ' before committing files to Git if `path/filepath` package is intended to be used.

stefanprodan:
> replace path/filepath with path.Join in sync.go line 96 or replace ' \ ' with ' / ' before committing files to Git if path/filepath package is intended to be used.

This will work on your local Windows machine, then when it gets to Flux it will fail since Flux runs in Alpine not Windows.

bibAtdev:
Thanks for your quick response.
Based on your answer I assume you would suggest to execute the build on a Linux based OS, or is there a known workaround for this issue?

PR #658, opened 2021-01-07
Coverts backward slash to forward slash in path flag

SomtochiAma:
Fixes: #631 


stefanprodan:
I don't see how would this fix #645 that issue is about TF and minifestgen, your fix is for the CLI only.

stefanprodan:
This only fixes #631

stefanprodan (approved):
LGTM

PR #660, opened 2021-01-07
Checks if bootstrap path differs

### LINKED ISSUE #659: Warn users when cluster has already been bootstrapped to a different path.
A user could forget to switch the Kubernetes context and try bootstrapping the cluster to a different path again.
We would want to tell the user and then quit the bootstrap process as you cannot bootstrap the cluster to two different paths.

SomtochiAma:
Fixes #659 

### DIFF
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -123,10 +123,10 @@ func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	bootstrapPathDiffers := checkIfBootstrapPathDiffers(ctx, kubeClient, namespace, filepath.ToSlash(ghPath.String()))
+	usedPath, bootstrapPathDiffers := checkIfBootstrapPathDiffers(ctx, kubeClient, namespace, filepath.ToSlash(ghPath.String()))
 
 	if bootstrapPathDiffers {
-		return fmt.Errorf("cluster already bootstrapped to a different path")
+		return fmt.Errorf("cluster already bootstrapped to a %v path", usedPath)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/flux/bootstrap_github.go
```diff
@@ -115,6 +115,20 @@ func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
...
+
+	kubeClient, err := utils.KubeClient(kubeconfig, kubecontext)
+	if err != nil {
+		return err
+	}
+
+	bootstrapPathDiffers := checkIfBootstrapPathDiffers(ctx, kubeClient, namespace, filepath.ToSlash(ghPath.String()))
+
+	if bootstrapPathDiffers {
+		return fmt.Errorf("cluster already bootstrapped to a different path")
```
stefanprodan, line 129: It would be good if we tell the user what path is being used.

cmd/flux/bootstrap_gitlab.go
```diff
@@ -123,10 +123,10 @@ func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	bootstrapPathDiffers := checkIfBootstrapPathDiffers(ctx, kubeClient, namespace, filepath.ToSlash(glPath.String()))
+	usedPath, bootstrapPathDiffers := checkIfBootstrapPathDiffers(ctx, kubeClient, namespace, filepath.ToSlash(glPath.String()))
 
 	if bootstrapPathDiffers {
-		return fmt.Errorf("cluster already bootstrapped to a different path")
+		return fmt.Errorf("cluster already bootstrapped to a %v path", usedPath)
```
stefanprodan, line 129:
```suggestion
		return fmt.Errorf("cluster already bootstrapped to '%v' path", usedPath)
```

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🍰

PR #663, opened 2021-01-07
Cross off tasks in the image automation roadmap

squaremo:
Specifically:
 - integration into `flux` CLI
 - guide to how to use the controllers


stefanprodan (approved):
LGTM

PR #683, opened 2021-01-11
Fix Azure DevOps URL in docs

stefanprodan:
Fix: https://github.com/fluxcd/source-controller/issues/253


andrewDoing (changes requested):
This new link will NOT work, I was trying to say that in my issue

hiddeco (approved):

hiddeco:
@andrewDoing pretty sure that should just work, as recently (publicly) confirmed in https://github.com/fluxcd/flux2/issues/591#issuecomment-744407260.

andrewDoing:
Yes, I just got it working, my apologies, and thanks for the prompt fix.

Issue #493, opened 2020-11-23
Create architecture telemetry in the gotk


anthr76:
Evening,

I'm inquiring for the feasibility of adding telemetry to your images hosted on ghcr. I came across this [discussion](https://github.com/fluxcd/flux2/discussions/303) earlier today. It seems as though multi-arch images are not being pushed due to the ghcr registries metrics. To combat this perhaps some kind of anonymous telemetry can phone-home so the maintainers are getting an idea of what users are running as their architecture. I'm currently using a 3rd party image on fluxv1 for flux and the helm-controller to support my multi-arch cluster. It would be greatly appreciated to see proper multi-arch images for AMD64 and AARCH64. Most popular images these days are supporting multi-arch builds and getting useful metrics through anonymous telemetry. Is there any interest in adding this since we're in active development?

Marx2:
It's really weird to make intentionally bad architectural decision (not supporting all architecture at once in multiarch build) at so early stage of development. I really like Flux, but being so rigid and not hearing users is a bad sign. If you really need telemetry, just do it correctly - built it properly,  make it clear for every user and allow opt-out. This is correct behavior in open source world.

cerebrate:
Actually, this is breaking the telemetry. I have a multi-arch cluster and whichever I install (amd64 or arm) hides the other from the telemetry when I would be perfectly happy to let the Flux deployments roam between nodes irrespective of architecture. Effectively, I'm a would-be ARM user hidden by this choice of data collection.

(I don't suppose anyone knows of a way to merge individual single-arch images into a multi-arch image?)

anthr76:
@cerebrate Yes this is the considerable issue. Currently the raspbernetes image repository offers a multi-arch build of fluxv1 with about 540k worth of pulls that the flux maintainers are missing out on, on their metrics. These builds are multi-arch and can be ran on ARM64 or AMD64.

You can find them here https://github.com/raspbernetes/multi-arch-images

Of course if the flux maintainers opted to instead maintain a multi-arch build and depend on metrics from telemetry rather than GHCR metrics they would have a wider scope of their users.

danielsand:
also a good reason for this is the last 12+ month:

Nvidia aquires ARM
https://nvidianews.nvidia.com/news/nvidia-to-acquire-arm-for-40-billion-creating-worlds-premier-computing-company-for-the-age-of-ai

Apple Silicon and M1
https://en.wikipedia.org/wiki/Apple_M1
[...]

PR #684, opened 2021-01-11
Deprecate arch flags in favor of multi-arch images

hiddeco:
This commit deprecates the architecture flag (`--arch`) for the install
and bootstrap commands, in favor of the bundled multi-arch images that
will be available for the next MINOR range of GOTK controller releases.

Summary of changes:

* `*Arch` variables have been marked as deprecated for both commands.
* `-arm64` suffix is no longer selectively added to the image definition
  of a component's `Deployment`.
* `kubernetes.io/arch` node selector with the defined value has been
  removed from the components' `Deployment`s.
* `Arch` has been removed from the available `Options` in
  `manifestgen/install`.
* Documentation references have been changed to highlight existence
  of multi-arch images and supported architectures.

Fixes: https://github.com/fluxcd/flux2/issues/493

### DIFF
--- a/docs/guides/installation.md
+++ b/docs/guides/installation.md
@@ -62,10 +62,10 @@ flux bootstrap <GIT-PROVIDER> \
   --version=latest
 ```
 
-!!! hint "ARM"
-    When deploying to a Kubernetes cluster with ARM architecture,
-    you can use `--arch=arm` for ARMv7 32-bit container images
-    and `--arch=arm64` for ARMv8 64-bit container images.
+!!! hint "Multi-arch images"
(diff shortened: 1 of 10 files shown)


docs/guides/installation.md
```diff
@@ -62,11 +62,6 @@ flux bootstrap <GIT-PROVIDER> \
   --version=latest
 ```
 
-!!! hint "ARM"
```
stefanprodan, line 65: Let's add here a multi-arch hint and list the supported arches

stefanprodan (approved):
LGTM

anthr76:
Thank you!

PR #694, opened 2021-01-13
Remove deprecated source behaviour

stefanprodan:


phillebaba (approved):
I guess the first half still holds true even though there is nothing enforcing it. Either way.

LGTM

PR #696, opened 2021-01-13
Update to controller-runtime 0.7.0

squaremo:
controller-runtime methods now accept `client.Object` and `client.ObjectList` rather than `runtime.Object`. This means the adapter interfaces need to change signature, but happily, little else.

Since the list adapter is now distinct to the object adapter, `len()` can go there instead of the command-specific interfaces.


stefanprodan (approved):
LGTM 💯

Issue #43, opened 2020-06-18
Platform agnostic bootstrapper


hiddeco:
A simple and straight forward way to make the toolkit work with any platform is to provide a bootstrapper that just assumes it should connect to some Git server while making use of the local SSH agent.

Command:

```
tk bootstrap git \
--repository=<GitHub repository name> \
--version=<toolkit version> \
--components=<toolkit controllers> \
```

Workflow:

- clone the repository
- generate install manifests
- commit and push the install manifests to git using the local SSH agent (override the files if they exist)
- apply the manifests on the target cluster (on install)
- wait for controllers rollout
- generate SSH keys and hosts (on install)
- print generated public key (on install)
- create a source manifest for the specified Git repository (on install)
- create a kustomization manifest for reconciling the manifests (on install)
- commit and push the source/ks manifests to git (on install)
- apply the manifests on the target cluster (on install)
- wait for the kustomization rollout

Bonus points:

Some Git servers allow new repositories to be created on push, we could provide an additional flag (e.g. `--init`) which creates a new Git repository locally, with the remote set to the provided `--repository` URL, instead of cloning it. Cloning should however always be attempted first to make the command idempotent.

stealthybox:
Related #1165

stefanprodan:
I would close this and open a new issue for implementing a bootstrap that shells out to the Git CLI (if we still want to do that, IMO this is very low prio).

PR #702, opened 2021-01-13
Add git implementation to generate sync options

phillebaba:
Add git implementations to options. I don't think it is pertinent to add this as a flag to the GitHub or GitLab bootstrap commands as they will probably not be used. I guess we should reconsider if we add self signed cert support which would only work in libgit2.

### DIFF
--- a/pkg/manifestgen/sync/options.go
+++ b/pkg/manifestgen/sync/options.go
@@ -19,23 +19,25 @@ package sync
 import "time"
 
 type Options struct {
-	Interval     time.Duration
-	URL          string
-	Name         string
-	Namespace    string
-	Branch       string
-	TargetPath   string
-	ManifestFile string
+	Interval          time.Duration
+	URL               string
+	Name              string
+	Namespace         string
+	Branch            string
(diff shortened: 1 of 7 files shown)


pkg/manifestgen/sync/options.go
```diff
@@ -19,23 +19,25 @@ package sync
...
-		ManifestFile: "gotk-sync.yaml",
-		TargetPath:   "",
+		Interval:          1 * time.Minute,
+		URL:               "",
+		Name:              "flux-system",
+		Namespace:         "flux-system",
+		Branch:            "main",
+		ManifestFile:      "gotk-sync.yaml",
+		TargetPath:        "",
+		GitImplementation: "go-git",
```
hiddeco, line 41: Maybe better to use the available `const` here from the package?

hiddeco:
> I don't think it is pertinent to add this as a flag to the GitHub or GitLab bootstrap commands as they will probably not be used. 

We should probably include this once we get to #43

stefanprodan (approved):

Issue #703, opened 2021-01-14
AKS: Azure network policy addon blocks source-controller ingress


mazen-bassiouny:
Dears, 

I'm trying to bootstrap flux2 in a new azure  AKS without any network policies defined.
After all CRDs are installed and the GitHub account is create, the bootstrap finishes with (time exceeded). The four controller pods are up and running. 
- The cluster is synchronized with the last commit correctly as below:

**#kubectl get gitrepositories.source.toolkit.fluxcd.io -A**
NAMESPACE     NAME          URL                                           READY   STATUS                                                            AGE
flux-system   flux-system   https://github.com/name/tanyflux   True    Fetched revision: main/6554ea6324d70caf0f2dfa200e137fd9c2aecc8a   59m

- But the Kustomization has the below error : 
**#kubectl get kustomizations.kustomize.toolkit.fluxcd -A**
failed to download artifact from http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/6554ea6324d70caf0f2dfa200e137fd9c2aecc8a.tar.gz, error: Get "http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/6554ea6324d70caf0f2dfa200e137fd9c2aecc8a.tar.gz": dial tcp 10.0.165.86:80: i/o timeout

- Same error exists while checking the Kustomization pod as below:
**#kubectl logs kustomize-controller-7f5455cd78-wwxhk -n flux-system**  
"level":"error","ts":"2021-01-14T09:04:52.524Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"kustomize.toolkit.fluxcd.io","reconcilerKind":"Kustomization","controller":"kustomization","name":"flux-system","namespace":"flux-system","error":"failed to download artifact from http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/6554ea6324d70caf0f2dfa200e137fd9c2aecc8a.tar.gz, error: Get \"http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/6554ea6324d70caf0f2dfa200e137fd9c2aecc8a.tar.gz\": dial tcp 10.0.165.86:80: i/o timeout"}

Thanks for any helpful advice.

stefanprodan:
Duplicate of: #690 and #697

frank-bee:
@stefanprodan 

with the latest cli I still get these errors

```
[...]

Issue #697, opened 2021-01-13
"flux bootstrap github" fails due to timeout on AKS cluster using repository on a GitHub Enterprise Server


abrokwasoft:
I'm trying to setup flux2 on an Azure Kubernetes Service cluster following https://toolkit.fluxcd.io/get-started/.

For this purpose I use a Git repository on a GitHub Enterprise Server. 

The bootstrap procedure fails though - it shows following error:
`✗ failed to download artifact from http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/2cb1b23bafe6aaa82a00562b0135dd7f9b5e57ca.tar.gz, error: Get "http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/2cb1b23bafe6aaa82a00562b0135dd7f9b5e57ca.tar.gz": dial tcp 10.2.126.78:80: i/o timeout`

Do you have any advices how to fix this?

Here more details about my setup:

```
$ flux --version
flux version 0.5.9
```

```
$ flux check
► checking prerequisites
✔ kubectl 1.19.4 >=1.18.0
✔ Kubernetes 1.18.10 >=1.16.0
► checking controllers
✔ source-controller is healthy
► ghcr.io/fluxcd/source-controller:v0.5.6
✔ kustomize-controller is healthy
► ghcr.io/fluxcd/kustomize-controller:v0.5.3
✔ helm-controller is healthy
► ghcr.io/fluxcd/helm-controller:v0.4.4
✔ notification-controller is healthy
► ghcr.io/fluxcd/notification-controller:v0.5.0
✔ all checks passed
```

```
$ flux bootstrap github \
>   --token-auth \
>   --hostname=$HOSTNAME \
>   --owner=$OWNER \
>   --repository=$REPO \
>   --branch=master \
>   --path=clusters/my-cluster \
>   --log-level debug
► connecting to github....com
✔ repository cloned
✚ generating manifests
✔ components manifests pushed
► installing components in flux-system namespace
namespace/flux-system configured
customresourcedefinition.apiextensions.k8s.io/alerts.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/buckets.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/gitrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmcharts.source.toolkit.fluxcd.io configured
[...]

Issue #690, opened 2021-01-12
i/o timeouts in kustomize-controller -> source-controller


ckittel:
I've been trying to troubleshoot a reconilation problem in my AKS cluster.  This is using the latest released flux CLI and also container images (tried with the freshly minted 0.6.0 as well)

```bash
flux --version
flux version 0.5.9

kubectl get service,deployment,pods,replicaset -n flux-system
NAME                        TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/source-controller   ClusterIP   172.16.34.92   <none>        80/TCP    14m

NAME                                   READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/kustomize-controller   1/1     1            1           14m
deployment.apps/source-controller      1/1     1            1           14m

NAME                                        READY   STATUS    RESTARTS   AGE
pod/kustomize-controller-5d4b86b547-kqr2q   1/1     Running   0          14m
pod/source-controller-6d976bcc57-cqgqp      1/1     Running   0          14m

NAME                                              DESIRED   CURRENT   READY   AGE
replicaset.apps/kustomize-controller-5d4b86b547   1         1         1       14m
replicaset.apps/source-controller-6d976bcc57      1         1         1       14m
```

Things seem to be deployed and running.  The source-controller logs are healthy, pumping out notifications on fetch.  The kustomize controller though is producing Reconciler errors...

```output
{"level":"debug","ts":"2021-01-12T21:18:22.086Z","logger":"events","msg":"Normal","object":{"kind":"Kustomization","namespace":"flux-system","name":"flux-system","uid":"3980146d-daf8-422b-961a-415f25aae09b","apiVersion":"kustomize.toolkit.fluxcd.io/v1beta1","resourceVersion":"1235134"},"reason":"error","message":"failed to download artifact from http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/680b5cb44a0f5b3de5804b19e03b2a8ba704ff08.tar.gz, error: Get \"http://source-controller.flux-system.svc.cluster.local./gitrepository/flux-system/flux-system/680b5cb44a0f5b3de5804b19e03b2a8ba704ff08.tar.gz\": dial tcp 172.16.34.92:80: i/o timeout"}
[...]

PR #708, opened 2021-01-14
Allow egress traffic for controller pods

stefanprodan:
Seems that AKS network policy addon blocks pod-to-pod communication by default, even if the pods are in the same namespace, this PR enabled egress traffic for Flux controller pods allowing kustomize/helm controllers to fetch artifacts from source-controller.

Fix: #703, #697, #690


phillebaba (approved):
LGTM

y-peter:
FYI: Linode Clusters (LKE) seem to have the same problem.

PR #773, opened 2021-01-25
Add permissions for controller leader election.

nairb774:
The v0.8.0 version of the controller-runtime uses both config maps and
leases to perform leader election. These permissions seem to be in the
individual controller repos, but not here. For example
https://github.com/fluxcd/kustomize-controller/blob/2d38de8779b68bde3366489a0b7657488c747fdb/config/rbac/leader_election_role.yaml#L33-L44

### DIFF
--- a/manifests/rbac/role.yaml
+++ b/manifests/rbac/role.yaml
@@ -28,6 +28,18 @@ rules:
   verbs:
   - create
   - patch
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @nairb774

Issue #243, opened 2020-09-18
Assign a dedicated cluster role per controller


paulcarlton-ww:
Running the installer generates yaml that installs the source-controller with cluster admin role. As I understand it, the recommended setup is for all *.source.toolkit.fluxcd.io custom resources to be in the gitops-system namespace so the source controller should not need access to anything outside this namespace? Also, for security conscious customers it might be worth defining different roles for each toolkit component, i.e. source-controller only needs access to the  *.source.toolkit.fluxcd.io resources not *.toolkit.fluxcd.io?

paulcarlton-ww:
As I understand it, if the source-controller is set to monitor gitops-system namespaced custom resources only (as per the default) the source-controller with use a role not a clusterrole?

stefanprodan:
No longer true, see https://github.com/fluxcd/toolkit/pull/227

william-salt-cko:
Has anyone got any more hints at potential permissions the components need? I was going to enable an audit policy to monitor the components and start making roles and rules and create service accounts, role bindings and roles with rules in a PR.

stefanprodan:
We generate the required RBAC from code, see the config/rbac in each controller repo, e.g. https://github.com/fluxcd/source-controller/tree/main/config/rbac

stefanprodan:
@william-salt-cko just be clear, we don’t want to duplicate any rbac rules in this repo, when we switch to dedicated service accounts, the rbac must come as a kustomize remote base from each controller repo based on the kubebuilder generated rules. The kubebuilder rbac doesn’t suffice right now, we need to add readonly access to secrets as most Flux objects have a secretRef field.

PR #777, opened 2021-01-25
Add a dedicated service account per controller

stefanprodan:
This PR assigns each controller a service account named the same as the controller. 

At install/upgrade time, `flux bootstrap` and `flux install` will create two cluster roles and bindings prefixed with the `--namespace` value. This change will also apply to Flux Terraform provider.

### crd-controller-flux-system 

This `ClusterRoleBinding` grands cluster wide RW access to the `toolkit.fluxcd.io` custom resources (required by controller runtime), RO access to Kubernetes secrets (required by `secretRef`) and RW access to Kubernetes leases (required by leader election). All the controller service accounts are bind to this cluster role.

### cluster-reconciler-flux-system

This `ClusterRoleBinding` grands `cluster-admin` access to kustomize-controller and helm-controller using their dedicate service account.

Fix: #243

### DIFF
--- a/docs/guides/installation.md
+++ b/docs/guides/installation.md
@@ -242,9 +242,7 @@ mkdir -p ./clusters/my-cluster/flux-system
 Generate the Flux manifests with:
 
 ```sh
-# on ARM64/AARCH64 clusters use --arch=arm64
 flux install --version=latest \
-  --arch=amd64 \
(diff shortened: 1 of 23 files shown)


hiddeco (approved):
Tested the upgrade path and confirmed it to be working as expected.

Awesome work @stefanprodan :cocktail:

PR #788, opened 2021-01-26
Create secret for TLS command

squaremo:
The image-reflector controller now accepts a secret containing a
client certificate and key, and/or a CA certificate; so it's useful to
have a command for creating them.

`flux create secret helm` is close, but accepts username/password
(which would be ignored), and has the wrong name of course. Happily
though, much can be shared between the implementations.

```bash
$ flux create secret tls --help

The create secret tls command generates a Kubernetes secret with certificates for use with TLS.

Usage:
  flux create secret tls [name] [flags]

Examples:

  # Create a TLS secret on disk and encrypt it with Mozilla SOPS.
  # Files are expected to be PEM-encoded.
  flux create secret tls certs \
    --namespace=my-namespace \
    --cert-file=./client.crt \
    --key-file=./client.key \
    --export > certs.yaml

    --in-place certs.yaml


Flags:
      --ca-file string     TLS authentication CA file path
      --cert-file string   TLS authentication cert file path
  -h, --help               help for tls
      --key-file string    TLS authentication key file path

Global Flags:
      --export              export in YAML format to stdout
      --interval duration   source sync interval (default 1m0s)
      --kubeconfig string   path to the kubeconfig file (default "/Users/mikeb/.kube/config")
      --label strings       set labels on the resource (can specify multiple labels with commas: label1=value1,label2=value2)
      --verbose             print generated objects
```

### DIFF
--- a/cmd/flux/create_secret.go
+++ b/cmd/flux/create_secret.go
@@ -39,6 +39,21 @@ func init() {
 	createCmd.AddCommand(createSecretCmd)
 }
 
+func makeSecret(name string) (corev1.Secret, error) {
+	secretLabels, err := parseLabels()
+	if err != nil {
+		return corev1.Secret{}, err
+	}
+
+	return corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: rootArgs.namespace,
+			Labels:    secretLabels,
+		},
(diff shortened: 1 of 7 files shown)


stefanprodan (approved):
LGTM

Thanks @squaremo

PR #790, opened 2021-01-26
Give image repository a cert-secret-ref flag

squaremo:
ImageRepository objects can now refer to a secret containing
certificates to use for TLS. This adds the flag

    flux create image repository --cert-secret-ref

for naming a secret to use. You can create such a secret with

    flux create secret tls

TODO:

 - [x] update the image-reflector API when there's a version which includes the CertSecretRef field
 - [x] give examples of creating objects that use cert secrets

### DIFF
--- a/cmd/flux/create_image_repository.go
+++ b/cmd/flux/create_image_repository.go
@@ -38,9 +38,10 @@ An ImageRepository object specifies an image repository to scan.`,
 }
 
 type imageRepoFlags struct {
-	image     string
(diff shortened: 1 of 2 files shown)


stefanprodan (commented):
It would be great to add an Example section to the command, to give users a hint for `create secret tls`.

stefanprodan (approved):
LGTM

Thanks @squaremo ☕

Issue #736, opened 2021-01-18
Flux v1 automation migration guide


squaremo:
There's a guide for setting image automation up from scratch. As a companion, to give people migrating from Flux v1 extra help, we should have a guide to how to set v2 automation up to work in the same way as v1 automation.

This needs to cover:

 - how to construct the equivalent to automation annotations
 - the equivalents to ignore (actually a sync annotation, but people use it in conjunction with automation), deautomate, and locking
 - general advice on the process for changing a Flux v1 system over to Flux v2

stefanprodan:
@squaremo isn't this done in #806 ?

squaremo:
Fixed by #806.

PR #806, opened 2021-01-27
Image automation migration guide

squaremo:
Ref: #736

TODO:

 - [ ] Fix up some of the links (apparently relative paths don't work)
 - [ ] Decide whether to mention migrating on a branch, or explain it in full, or ignore the possibility.
 - [ ] Operating the automation like Flux v1:
    - how to pause an automation
    - how to lock an image where it is

### DIFF
--- a/docs/guides/flux-v1-automation-migration.md
+++ b/docs/guides/flux-v1-automation-migration.md
@@ -1,0 +1,723 @@
+<!-- -*- fill-column: 100 -*- -->
+# Migrating image update automation to Flux v2
+
+"Image Update Automation" is a process in which Flux makes commits to your Git repository when it
+detects that there is a new image to be used in a workload (e.g., a Deployment). In Flux v2 this
+works quite differently to how it worked in Flux v1. This guide explains the differences and how to
+port your cluster configuration from v1 to v2. There is also a [tutorial for using image update
+automation with a new cluster][image-update-tute].
+
+## Overview of changes between v1 and v2
+
+In Flux v1, image update automation (from here, just "automation") was built into the Flux daemon,
+which scanned everything it found in the cluster and updated the Git repository it was syncing.
+
+In Flux v2,
+
+ - automation is controlled with custom resources, not annotations
+ - ordering images by build time is not supported (there is [a section
+   below](#how-to-migrate-annotations-to-image-policies) explaining what to do instead)
+ - the fields to update in files are marked explicitly, rather than inferred from annotations.
+
+#### Automation is now controlled by custom resources
+
+Flux v2 breaks down the functions in Flux v1's daemon into controllers, with each having a specific
+area of concern. Automation is now done by two controllers: one which scans image repositories to
+find the latest images, and one which uses that information to commit changes to git
+repositories. These are in turn separate to the syncing controllers.
+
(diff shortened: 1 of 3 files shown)


docs/guides/flux-v1-automation-migration.md
```diff
@@ -0,0 +1,507 @@
...
+In Flux v2,
+
+ - automation is controlled with custom resources, not annotations
+ - ordering images by build time is not supported (see
+   [below](#how-to-use-timestamp-in-image-tags) for what to do
+   instead)
+ - the fields to update are marked explicitly, rather than inferred
+   from annotations.
+
+**Automation is now controlled by custom resources**
```
hiddeco, line 25: If you make this proper `###`, they show up in the right sidebar on the documentation website under "table of contents".

docs/guides/sortable-image-tags.md
```diff
@@ -0,0 +1,150 @@
...
+$ # date and time (remember ':' is not allowed in a tag)
+$ date +%F.%H%M%S
+2021-01-28.133158
+```
+
+Alternatively, you can use a stable serial number as part of the tag.  Some CI platforms will
+provide a build number in an environment variable, but that may not be reliable to use as a serial
+number -- check the platform documentation.
+
+A commit count can be a reasonable stand-in for a serial number, if you build an image per commit,
```
stefanprodan, line 33: I don't think this will work with alphabetic ordering, you need to use semver e.g. `1.0.$REV_COUNT`. I would remove this from the docs as it's error prone.

squaremo: Hmm yes, you would need to pad it to a particular number of characters, for it to sort correctly.

squaremo: Easier to remove it than to make it workable.

docs/guides/flux-v1-automation-migration.md
```diff
@@ -0,0 +1,769 @@
...
+  name: my-app-policy
+  namespace: flux-system
+spec:
+  imageRepositoryRef:
+    name: podinfo-image
+  filterTags:
+    pattern: '^main-[a-f0-9]+-(?P<ts>[0-9]+)'
+    extract: '$ts'
+  policy:
+    alphabetical:
```
stefanprodan, line 540:
```suggestion
    numerical:
```

docs/guides/flux-v1-automation-migration.md
```diff
@@ -0,0 +1,769 @@
...
+      order: asc
+```
+
+The `.spec.pattern` field gives a regular expression that a tag must match to be included. The
+`.spec.extract` field gives a replacement pattern that can refer back to capture groups in the
+filter pattern. The extracted values are sorted to find the selected image tag. In this case, the
+timestamp part of the tag will be extracted and sorted alphabetically in ascending order. See [the
```
stefanprodan, line 547:
```suggestion
timestamp part of the tag will be extracted and sorted numerically in ascending order. See [the
```

stefanprodan (approved):
LGTM

Awesome guide @squaremo I added a suggestion on using numerical but it's no blocker.

stefanprodan (approved):

PR #812, opened 2021-01-28
Enable network policy in e2e

### LINKED ISSUE #800: Enable network policy in KIND for e2e testing
As a Flux developer,
I'd like to test Flux on a cluster that enabled network policy,
So that I can make sure all Flux features are working fine in that environment.

To do so, we'll use Calico with KIND instead of the default KIND-net.

chanwit:
This PR disables the default KIND-net, and uses Calico CNI instead to enable network policy for the KIND cluster in e2e testing.

fixes #800


stefanprodan (approved):
LGTM

Thanks @chanwit 🥇

PR #813, opened 2021-01-28
Update docs ImageUpdateAutomation example

relu:
The API changed in v0.7 and the example from the guide is outdated.


stefanprodan (approved):
LGTM

PR #821, opened 2021-01-29
Add HelmChart not ready to FAQ

stefanprodan:
Changes:
- rename "general questions" to "Flux v1 vs v2"
- add How to debug "not ready" errors? to Helm FAQ

### DIFF
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -1,98 +1,5 @@
 # Frequently asked questions
 
-## General questions
-
-### What does Flux v2 mean for Flux?
-
-Flux v1 is a monolithic do-it-all operator; Flux v2 separates the functionalities into specialized controllers, collectively called the GitOps Toolkit.
-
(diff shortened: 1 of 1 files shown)


docs/faq/index.md
```diff
@@ -242,3 +149,126 @@ The kustomize-controller creates `kustomization.yaml` files similar to:
...
+In order to get to the root cause, first make sure the source e.g. the `HelmRepository`
+is configured properly and has access to the remote `index.yaml`:
+
+```console
+$ flux get sources helm --all-namespaces 
+NAMESPACE  	NAME   	READY	MESSAGE
+default   	podinfo	False	failed to fetch https://stefanprodan.github.io/podinfo2/index.yaml : 404 Not Found
+```
+
+If the source is ok, then the error must be caused by an invalid chart name or a non-existing version:
```
hiddeco, line 175:
```suggestion
If the source is `Ready`, then the error must be caused by the chart, for example due to an invalid chart name or non-existing version:
```

hiddeco (approved):

PR #824, opened 2021-01-29
Highlight PATCH versions can be used to upgrade

hiddeco:


docs/guides/installation.md
```diff
@@ -512,6 +512,10 @@ flux create helmrelease nginx \
 
 ## Upgrade
```
hiddeco, line 514: Open to better title suggestions.

docs/guides/installation.md
```diff
@@ -512,6 +512,10 @@ flux create helmrelease nginx \
...
 
+!!! note "Versions"
```
hiddeco, line 515: Open to better title suggestions.

docs/guides/installation.md
```diff
@@ -512,6 +512,10 @@ flux create helmrelease nginx \
...
+Any PATCH version can be used to upgrade to a new MINOR version. It is for
```
stefanprodan, line 516: I would encourage the user of the latest patch release, something like:

Please use the latest PATCH version when upgrade to a new MINOR version.

hiddeco: Changed the wordings a bit.

stefanprodan (approved):

docs/guides/installation.md
```diff
@@ -512,6 +512,10 @@ flux create helmrelease nginx \
 
 ## Upgrade
 
+!!! note "Patch versions"
+It is safe and advised to use the latest PATCH version when upgrading to a new
```
stefanprodan, line 516: I think this needs to be indented to be included in the note.

PR #840, opened 2021-02-01
Improve image update CronJob examples

relu:


stefanprodan (approved):
LGTM

Thanks @relu

docs/guides/image-update.md
```diff
@@ -558,6 +559,16 @@ you can manually create an init job using the following command:
 $ kubectl create job --from=cronjob/ecr-credentials-sync -n flux-system ecr-credentials-sync-init
 ```
 
+After the job runs, a secret named `ecr-credentials` should be created. Use this
+name in your ECR ImageRepository resource manifest as the value for
+`.spec.secretRef.name`.
+
+```
```
hiddeco, line 566: Maybe add `yaml` to the code block for pretty highlighting?

PR #843, opened 2021-02-01
Correct spelling of repository in error

squaremo:


hiddeco (approved):
Repoi-story would have been a nice word too.

squaremo:
> Repoi-story would have been a nice word too.

Let's keep that one in a drawer somewhere, It has a nice mouthfeel to it.

PR #848, opened 2021-02-02
Exclude deleted resources on prometheus alerting query

ViBiOh:
If a resource is deleted, the alert will fire indefinitely.

I multiply the "Deleted" count by 2 to have a distinct case.
* If the reconciliation fails, left part is valued 1.
* If the resource is deleted, right part is valued 2.

### DIFF
--- a/docs/guides/monitoring.md
+++ b/docs/guides/monitoring.md
@@ -86,7 +86,7 @@ groups:
 - name: GitOpsToolkit
   rules:
   - alert: ReconciliationFailure
-    expr: gotk_reconcile_condition{type="Ready",status="False"} == 1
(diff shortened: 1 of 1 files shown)


docs/guides/monitoring.md
```diff
@@ -86,7 +86,7 @@ groups:
 - name: GitOpsToolkit
   rules:
   - alert: ReconciliationFailure
-    expr: gotk_reconcile_condition{type="Ready",status="False"} == 1
+    expr: (gotk_reconcile_condition{status="False",type="Ready"} + on(exported_namespace, name, kind) (gotk_reconcile_condition{status="Deleted"} * 2)) == 1
```
stefanprodan, line 89: Have this been tested? There is no such label in our exporters `exported_namespace`

ViBiOh: Oh indeed, it maybe on my prometheus relabelling 👍 I'll try with a raw scraper

stefanprodan (approved):
LGTM

@ViBiOh can you rebase with main, thanks!

PR #854, opened 2021-02-03
Move migration sub-menu to top-menu

hiddeco:


stefanprodan:
Can we move it after Guides?

hiddeco:
The idea behind the current order was that you either want to "get started" or "migrate", and then want to further dive into guides.

stefanprodan (approved):

PR #857, opened 2021-02-03
Update toolkit components

fluxcdbot:
- source-controller to v0.7.4
  https://github.com/fluxcd/source-controller/blob/v0.7.4/CHANGELOG.md


hiddeco:
@stefanprodan I am wondering if we should tweak the CHANGELOG introduced in #853 from

```
- source-controller to [v0.7.4](fluxcd/source-controller@v0.7.4/CHANGELOG.md)
```

to

```
- source-controller to v0.7.4
  fluxcd/source-controller@v0.7.4/CHANGELOG.md
```

I think it would improve readability in the commit message, and would provide more context within the PR.

hiddeco (approved):

PR #860, opened 2021-02-04
Adding --select-alpha and --extract to create image policy

### LINKED ISSUE #822: Support alphabetical policy and filter `.extract` in flux create image policy
You can create an ImagePolicy with a semver range, and filter by regex, but not give a replacement pattern (`.spec.filter.extract` field) or use an alphabetical (asc or desc) ordering.

Suggested UI:

```bash
$ ./bin/flux create image policy --help
The create image policy command generates an ImagePolicy resource.
An ImagePolicy object calculates a "latest image" given an image
repository and a policy, e.g., semver.

The image that sorts highest according to the policy is recorded in
the status of the object.

Usage:
  flux create image policy <name> [flags]

Flags:
      --alpha string           use alphabetical sorting to select image; either "asc" meaning select the last, or "desc" meaning select the first
      --filter-regex string    regular expression pattern used to filter the image tags
      --filter-extract string  replacement pattern (using capture groups from --filter-regex) to use for sorting
  -h, --help                   help for policy
      --image-ref string       the name of an image repository object
      --semver string          use a semver range to select an image; e.g., '1.x'

Global Flags:
      --export              export in YAML format to stdout
      --interval duration   source sync interval (default 1m0s)
      --kubeconfig string   path to the kubeconfig file (default "/Users/mikeb/.kube/config")
      --label strings       set labels on the resource (can specify multiple labels with commas: label1=value1,label2=value2)
      --verbose             print generated objects
```

Also worth considering putting `--alpha` and `--semver` next to each other by prefixing them, e.g., `--select-semver` and `--select-alpha` or similar.

jonathan-innis:
Closes #822

### DIFF
--- a/cmd/flux/create_image_policy.go
+++ b/cmd/flux/create_image_policy.go
@@ -18,6 +18,10 @@ package main
 
 import (
 	"fmt"
+	"regexp/syntax"
+	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -94,6 +98,8 @@ func createImagePolicyRun(cmd *cobra.Command, args []string) error {
 	}
 
 	switch {
+	case imagePolicyArgs.semver != "" && imagePolicyArgs.alpha != "":
+		return fmt.Errorf("policy cannot be specified with both --select-semver and --select-alpha")
 	case imagePolicyArgs.semver != "":
 		policy.Spec.Policy.SemVer = &imagev1.SemVerPolicy{
 			Range: imagePolicyArgs.semver,
@@ -110,11 +116,18 @@ func createImagePolicyRun(cmd *cobra.Command, args []string) error {
 	}
 
 	if imagePolicyArgs.filterRegex != "" {
+		exp, err := syntax.Parse(imagePolicyArgs.filterRegex, syntax.Perl)
+		if err != nil {
+			return fmt.Errorf("--filter-regex is an invalid regex pattern")
+		}
 		policy.Spec.FilterTags = &imagev1.TagFilter{
 			Pattern: imagePolicyArgs.filterRegex,
 		}
 
 		if imagePolicyArgs.filterExtract != "" {
+			if err := validateExtractStr(imagePolicyArgs.filterExtract, exp.CapNames()); err != nil {
+				return err
+			}
 			policy.Spec.FilterTags.Extract = imagePolicyArgs.filterExtract
 		}
 	} else if imagePolicyArgs.filterExtract != "" {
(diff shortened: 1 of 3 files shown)


cmd/flux/create_image_policy.go
```diff
@@ -94,14 +98,27 @@ func createImagePolicyRun(cmd *cobra.Command, args []string) error {
 		policy.Spec.Policy.SemVer = &imagev1.SemVerPolicy{
 			Range: imagePolicyArgs.semver,
 		}
+	case imagePolicyArgs.alpha != "":
```
squaremo, line 101: Should also check (perhaps earlier) `alpha XOR semver` -- i.e., the user supplied _one_ of `--select-alpha`, `--select-semver`.

cmd/flux/create_image_policy.go
```diff
@@ -94,14 +98,27 @@ func createImagePolicyRun(cmd *cobra.Command, args []string) error {
...
-		return fmt.Errorf("a policy must be provided with --semver")
+		return fmt.Errorf("a policy must be provided with either --select-semver or --select-alpha")
 	}
 
 	if imagePolicyArgs.filterRegex != "" {
 		policy.Spec.FilterTags = &imagev1.TagFilter{
 			Pattern: imagePolicyArgs.filterRegex,
 		}
+
+		if imagePolicyArgs.filterExtract != "" {
```
squaremo, line 117: Stretch goal: can the extract pattern be validated against the regex, e.g., to make sure it doesn't mention capture groups that aren't present?

squaremo (changes requested):
Neat and tidy, thanks Jonathan 🍏  I'm always happy to see concise commit messages too :-)
One minor request (also in a comment): could it check that the user didn't supply _both_ semver and alpha, please.

cmd/flux/create_image_policy.go
jonathan-innis: Added this validation

cmd/flux/create_image_policy.go
jonathan-innis: Added validation, though the amount of code to do this is a bit humorous

stefanprodan:
@jonathan-innis can you please rebase with main, the e2e tests will fail after that, you need to change `./bin/flux get image policy podinfo | grep '5.0.3'` to match your policies.

jonathan-innis:
> @jonathan-innis can you please rebase with main, the e2e tests will fail after that, you need to change `./bin/flux get image policy podinfo | grep '5.0.3'` to match your policies.

Rebased

stefanprodan (approved):
LGTM

Thanks @jonathan-innis 🏅

squaremo (approved):

cmd/flux/create_image_policy.go
squaremo: Oof, yep. If extract is taken verbatim from `<go stdlib>/regexp/regexp.go`, it's worth noting this I think.

cmd/flux/create_image_policy.go
```diff
@@ -117,3 +147,94 @@ func createImagePolicyRun(cmd *cobra.Command, args []string) error {
...
+		}
+		name, num, rest, ok := extract(template)
+		if !ok {
+			// Malformed extract string, assume user didn't want this
+			template = template[1:]
+			return fmt.Errorf("--filter-extract is malformed")
+		}
+		template = rest
+		if num >= 0 {
+			// we won't worry about numbers as we can't validate these
```
squaremo, line 171: You could check it against `exp.MaxCap()` (exp from the caller).

Issue #674, opened 2021-01-08
Unable to bootstrap when GitLab Subgroup is used


MonolithProjects:
Hi team, for some reason the `flux bootstrap` is failing when GitLab Subgroup is used. If the repository is located directly in the parent group, everything works fine.

```
❯ flux bootstrap gitlab \
  --owner devops_sre/test_apps \
  --token-auth \
  --hostname "gitlab.tools.in.pan-net.eu" \
  --repository "ghost-db-k8s" \
  --path "clusters/alien" \
  --namespace "ghost" \
  --branch "master" \
  --cluster-domain "alien.caas.in.pan-net.eu." \
  --verbose
► connecting to gitlab.tools.in.pan-net.eu
✗ failed to list projects, error: failed to find group named 'devops_sre/test_apps'
```
```
❯ flux --version
flux version 0.5.8
```
PAT has the full permissions. The group and subgroup names are 100% correct. So, i am wondering what can be the problem :thinking: 
```
curl --request GET --header "PRIVATE-TOKEN: $GITLAB_TOKEN" --header "Content-Type: application/json" "https://gitlab.tools.in.pan-net.eu/api/v4/groups?search=test_apps"
[{"id":1096,"web_url":"https://gitlab.tools.in.pan-net.eu/groups/devops_sre/test_apps","name":"test_apps","path":"test_apps",...
```

stefanprodan:
Looks like a bug in https://github.com/fluxcd/pkg/blob/main/git/provider_gitlab.go I guess we don't support sub-groups cc @relu @SomtochiAma

totycro:
Hi,

I still hit this bug with later version. I've now tried 0.6.0, 0.7.0 and 0.7.5, and everytime I get this:
```
$ flux bootstrap gitlab --branch main --repository=eoxhub-flux-system --hostname gitlab.eox.at --owner "eox/hub"  --verbose --token-auth
✗ failed to list projects, error: failed to find group named 'eox/hub'
```

The git repo is available under `gitlab.eox.at:eox/hub/eoxhub-flux-system.git`.

I've also tried with `--owner "/eox/hub"` and `--owner "eox/hub/"` but that didn't help. Am I calling this the wrong way or is this still the same bug?

MonolithProjects:
Hmm... it works for me just fine
```
~/Documents/NFVI/flux2 master
❯ flux --version
flux version 0.7.6

~/Documents/NFVI/flux2 master
[...]

PR #863, opened 2021-02-04
Fix GitLab bootstrap by using (sub)group paths

hiddeco:
This incorporates the changes made to the GitLab provider here: https://github.com/fluxcd/pkg/pull/80.

This means that we no longer rely on UI names, but rather use the unique path identifier (the elements you see in your address bar when looking at e.g. a group in your GitLab environment).

Binary for testing is available at: https://dl.huge.horse/flux

Addresses https://github.com/fluxcd/flux2/issues/674


stefanprodan:
@hiddeco can you please update the install doc (GitLab section) with the path requirement.

stefanprodan (approved):
LGTM

PR #877, opened 2021-02-08
Add ACR auth to Image Updates examples

stealthybox:
AWS/GCP/Azure all had similar approaches so I factored the 3 approaches onto
a single kustomize base for a Deployment and CronJob option.

The cloud specific reconcile() shell functions are interpolated into the command
array using the Pod API so that we can patch it separately from any shell library funcs.

The KUBE_SECRET is user configurable, and as long as they are using kustomize,
it's updated on the Role as the only allowed resourceName, so that the RBAC is safer.

A similar vars trick is used to keep the AAD Pod label in sync.

The documented example uses a remote base since kustomize does not have a standard vendoring approach.
I can update the GCP and AWS examples to reference kustomize in a follow-up.

- Document ACR / AKS Image Update Considerations
- Add registry cred Deployments/CronJobs for aws/gcp/azure via kustomize

### DIFF
--- a/docs/guides/image-update.md
+++ b/docs/guides/image-update.md
@@ -657,7 +657,7 @@ spec:
 
 #### Using access token [short-lived]
 
-!!!note "Workload Identity"
+!!! note "Workload Identity"
     Please ensure that you enable workload identity for your cluster, create a GCP service account that has
     access to the container registry and create an IAM policy binding between the GCP service account and
     the Kubernetes service account so that the pods created by the cronjob can access GCP APIs and get the token.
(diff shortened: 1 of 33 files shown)


docs/guides/image-update.md
```diff
@@ -792,4 +792,87 @@ or [Sealed Secrets](sealed-secrets.md) , commit and push the encypted file to gi
...
+Once we have AAD Pod Identity installed, we can create a Deployment that frequently refreshes an image pull secret into
+our desired Namespace.
+
+Create a directory in your control repository and save this `kustomization.yaml`:
+```yaml
+# kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- github.com/fluxcd/flux2/manifests/integrations/registry-credentials-sync/azure
```
stefanprodan, line 822: This will not work since Kustomize defaults to `master` branch and we use `main`.

stealthybox: ty -- moving to the ZIP archive on main

docs/guides/image-update.md
```diff
@@ -792,4 +792,87 @@ or [Sealed Secrets](sealed-secrets.md) , commit and push the encypted file to gi
...
+Once we have AAD Pod Identity installed, we can create a Deployment that frequently refreshes an image pull secret into
+our desired Namespace.
+
+Create a directory in your control repository and save this `kustomization.yaml`:
+```yaml
+# kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- https://github.com/fluxcd/flux2/archive/main.zip//manifests/integrations/registry-credentials-sync/azure
```
stealthybox, line 822: Updated to zip archive @ `main`.
We don't have a `stable` tag that denotes the latest release -- that could be useful in the future.

docs/guides/image-update.md
```diff
@@ -792,4 +792,87 @@ or [Sealed Secrets](sealed-secrets.md) , commit and push the encypted file to gi
...
+  resourceID: /subscriptions/8c69185e-55f9-4d00-8e71-a1b1bb1386a1/resourcegroups/stealthybox/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acr-sync
+  type: 0  # user-managed identity
+```
+
+Verify that `kustomize build .` works, then commit the directory to you control repo.
+Flux will apply the Deployment and it will use the AAD managed identity for that Pod to regularly fetch ACR tokens into your configured `KUBE_SECRET` name.
+Reference the `KUBE_SECRET` value from any `ImageRepository` objects for that ACR registry.
+
+This example uses the `fluxcd/flux2` github archive as a remote base, but you may copy the [./manifests/integrations/registry-credentials-sync/azure](github.com/fluxcd/flux2/tree/main/manifests/integrations/registry-credentials-sync/azure)
+folder into your own repository or use a git submodule to vendor it if preferred.
```
stealthybox, line 862: Updated this link -- I realized it was broken before

stefanprodan (approved):
LGTM

Thanks @stealthybox 🍔

Issue #873, opened 2021-02-07
Enable pprof endpoints and write guide


hiddeco:
A helper was added to `pkg/runtime` to enable the pprof endpoints on the metrics server of controller-runtime: https://github.com/fluxcd/pkg/pull/81.

We should enable the endpoints on every Toolkit controller, and write a guide to explain to users how to port-forward and dump the CPU and memory profiles.

PR #878, opened 2021-02-08
Add guide for pprof endpoints

hiddeco:
Fix: #873

Should not be merged before the controllers have seen a release.


stefanprodan (approved):
LGTM

Please add this to mkdoc index.

Issue #811, opened 2021-01-28
flux uninstall deletes HelmReleases with old apiVersions


mrkwtz:
Hey everyone,

so I just wanted to switch from helm-operator to helm-controller, because the former one is not maintained anymore and a chart we want to use needs a newer helm version. 

So I installed the helm-controller via _flux install_ but with a bugged version (the bug where the CRB always refers to the SA in the flux-system namespace, but we use gitops-toolkit as namespace). So I thought: alrighty, then I'll just use _flux uninstall --namespace gitops-toolkit_ and we're good to go. The CLI also tells me that it'll remove the namespace itself, and that's fine for me.

But then this happend (just xxx'ed the sensitive information)

`xxx@xxx-TUXEDO:~$ flux uninstall --namespace gitops-toolkit
? Are you sure you want to delete the gitops-toolkit namespace? [y/N] y█

► uninstalling custom resources
No resources found
helmrepository.source.toolkit.fluxcd.io "grafana" deleted
No resources found
No resources found
helmrelease.helm.fluxcd.io "argo-events" deleted
helmrelease.helm.fluxcd.io "argo-workflows" deleted
helmrelease.helm.fluxcd.io "flux" deleted
helmrelease.helm.fluxcd.io "cert-manager" deleted
helmrelease.helm.fluxcd.io "cluster-autoscaler" deleted
helmrelease.helm.fluxcd.io "flux" deleted
helmrelease.helm.fluxcd.io "xxx-staging" deleted
helmrelease.helm.fluxcd.io "external-dns" deleted
helmrelease.helm.fluxcd.io "flux" deleted
helmrelease.helm.fluxcd.io "xxx-alpha" deleted
helmrelease.helm.fluxcd.io "xxx-backup" deleted
helmrelease.helm.fluxcd.io "xxx-bufix-capitalize" deleted
helmrelease.helm.fluxcd.io "xxx-dev-1278" deleted
helmrelease.helm.fluxcd.io "xxx-dev-1386" deleted
helmrelease.helm.fluxcd.io "xxx-dev-1413" deleted
helmrelease.helm.fluxcd.io "xxx-dev-537" deleted
helmrelease.helm.fluxcd.io "xxx-dev-644" deleted
helmrelease.helm.fluxcd.io "xxx-dev-665" deleted
helmrelease.helm.fluxcd.io "xxx-dev-748" deleted
helmrelease.helm.fluxcd.io "xxx-dev-918" deleted
[...]

PR #884, opened 2021-02-08
Make `flux uninstall` warning verbiage much stronger

kingdonb:
I am nearly sure it is not intended to delete helmreleases.helm.fluxcd.io HelmReleases when users run `flux uninstall`, but the effect is there currently, it does delete helm releases from the old API version. So until a behavior change can be implemented (if one makes sense), a more strongly worded warning might be appropriate such as this. (#811 first reported this.)

I would expect `flux uninstall` to only remove resources that come from Flux v2 CRDs, not Helm Operator HRs, since a person may try to upgrade their cluster to Flux v2, run into some issue, and then abort, thinking it's safe to just run `flux uninstall`.

I think the risk may be slightly overblown, so far only developer clusters have been impacted by this that we know of, at least. But the worst case scenario modes of failure are pretty bleak. Now that we both know it does this, we won't make this mistake again! It was discussed at another Flux team meeting that some migrations might not take place at once, so we should take some extra care around scenarios that may impact users who have only partially migrated their cluster from Flux v1 to Flux v2.

Does it make sense to merge in a change to the warning message such as this?

Is it possible to change the uninstall routine to refer to a blessed list of CRDs (maybe one exists) or at least regex filter the fully qualified names so that `helm.fluxcd.io` HelmReleases are not uninstalled too when this code runs? I think we have other CRDs with names that overlap other well-known types that could be affected by this, like Kustomization.

### DIFF
--- a/cmd/flux/uninstall.go
+++ b/cmd/flux/uninstall.go
@@ -78,7 +78,7 @@ func uninstallCmdRun(cmd *cobra.Command, args []string) error {
 
 	if !uninstallArgs.dryRun && !uninstallArgs.silent {
 		prompt := promptui.Prompt{
-			Label:     fmt.Sprintf("Are you sure you want to delete the %s namespace", rootArgs.namespace),
+			Label:     fmt.Sprintf("Are you sure you want to delete the %s namespace (including all Flux CRDs and resources, as well as Flux v1 helm.fluxcd.io HelmReleases)", rootArgs.namespace),
 			IsConfirm: true,
 		}
 		if _, err := prompt.Run(); err != nil {


hiddeco:
Wouldn't it be a better solution to use the FQDN of the Custom Resources here: https://github.com/fluxcd/flux2/blob/42011d028e09112440ae26be3d3b32a9e2f1a6de/cmd/flux/uninstall.go#L93-L97 to simply solve the whole issue of removing the wrong resources?

yebyen:
Yes, absolutely. I wasn't sure how to suggest that myself, but I agree, if this is not an intentional thing then it should be fixed.

I still think a text change is warranted because the default value `--resources true` will remove resources that are not mentioned in the warning (all the custom resources, which could be in any namespace) but the third clause in my suggested text can go as long as this is an error and can be fixed.

stefanprodan:
Closing in favour of #891

PR #898, opened 2021-02-11
docs: improve Kustomize behavior FAQ

hiddeco:


stefanprodan:
Wanted to do this next :) Please add a link to the validation script https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/scripts/validate.sh to encourage people to run this locally or in CI before pushing/merging changes to branches that get synced.

stefanprodan (approved):
LGTM

Thanks @hiddeco ☕

Issue #723, opened 2021-01-15
Make `flux bootstrap` commit messages descriptive


hiddeco:
At present, running `flux bootstrap` will always commit with the same commit message: `Add manifests`. This does not tell the user much about the change in the commit when they are inspecting the Git repository using e.g. `git log`.

The commit message should be changed to something that includes the version, or maybe even include if it was an install or an upgrade.

In addition, we may want to list the versions of all (changed) controllers in the long commit message.

kingdonb:
Do you think #904 went far enough, or should we go further and add the versions of which controllers have been changed to the commit message?

kingdonb:
I think there is another enhancement that would be nice, somewhat unrelated, (but if I'm picking this issue up I can maybe handle both) -- `flux version` tells you what version of CLI you are running and what version of controllers are on the cluster, but it does not tell you about any diff. It should at least tell users if `flux bootstrap` to upgrade would provide new versions of any controllers.

stefanprodan:
> `flux version` tells you what version of CLI you are running and what version of controllers are on the cluster, but it does not tell you about any diff. 

Nor does kubectl or any other CLI that I've used so far. The `flux check` command is where we show a "diff", currently only for the CLI version.

PR #904, opened 2021-02-12
Add flux version to bootstrap commit messages

stefanprodan:
This PR changes the commit messages of `flux bootstrap` to include the actual flux version even when using latest.

Example:

```
flux bootstrap github \
--owner=stefanprodan \
--repository=test \
--branch=main \
--path=./clusters/my-cluster \
--personal
```

Screenshot 2021-02-12 at 10 33 39

Ref: #723


hiddeco (approved):
I think this will be much appreciated by users as they can now distinguish commits, thank you @stefanprodan :beers:

Issue #920, opened 2021-02-15
Bootstrap version warning


gdziwoki:
### Describe the bug

When using `flux bootstrap` with an incorrect version (for example `--version=0.7.4`) version defaults to latest/
It would be nice to either throw an error, or a least get a log entry stating that version defaulted to latest.

### To Reproduce

Steps to reproduce the behaviour:

`flux bootstrap --version=$nonexistingversion`

### Expected behavior

Error or warning in logs, that used version is incorrect.

### Additional context

Below please provide the output of the following commands:

```cli
flux --version
flux version 0.7.7

flux check
► checking prerequisites
✔ kubectl 1.20.2 >=1.18.0
✔ Kubernetes 1.18.9-eks-d1db3c >=1.16.0
► checking controllers
► ghcr.io/fluxcd/source-controller:v0.7.4
✔ kustomize-controller: healthy
► ghcr.io/fluxcd/kustomize-controller:v0.7.4
✔ helm-controller: healthy
► ghcr.io/fluxcd/helm-controller:v0.6.1
► ghcr.io/fluxcd/notification-controller:v0.7.1

```

PR #922, opened 2021-02-15
Improve version checks for `check`, `bootstrap` and `install`

hiddeco:
Addresses: #920

- [x] Use label selector to check components in `flux check`
- [x] Check for new Flux binary version in `flux check`
- [x] Check if targeted version does exist for `flux bootstrap|install`
- [x] Check if targeted version is supported by current version of `flux`


stefanprodan (approved):
LGTM 

Thanks @hiddeco

PR #923, opened 2021-02-15
Use `MergeMaps` from pkg/runtime v0.8.2

arbourd:
pkg/runtime v0.8.2 introduces runtime/transform which includes `MergeMaps`: https://github.com/fluxcd/pkg/pull/85


hiddeco (approved):
Thank you @arbourd :1st_place_medal:

PR #929, opened 2021-02-16
Refactor Flux GitHub Action

stefanprodan:
Use composite mode to download and add the `flux` binary to the GitHub runner bin dir.


action/action.yml
```diff
@@ -1,15 +1,38 @@
...
+        tar -C /tmp/flux/ -zxvf /tmp/flux.tar.gz
+    - name: "Add flux binary to /usr/local/bin"
+      shell: bash
+      run: |
+        sudo cp /tmp/flux/flux /usr/local/bin
+    - name: "Cleanup tmp"
+      shell: bash
+      run: |
+        rm -rf /tmp/flux/ /tmp/flux.tar.gz
+    - name: "Verify binary"
```
hiddeco, line 35:
```suggestion
    - name: "Print flux binary version"
```

maybe? I would expect "verify" to run e.g. `flux check --pre`.

stefanprodan: The check command requires kubectl and a Kubernetes cluster. By calling -v we make sure the binary can be run on the runner.

hiddeco: I am well aware, that is why my suggestion changed the name of the run, and not the command.

hiddeco:
```suggestion
    - name: "Verify correct installation of binary"
```

would be another option. It is more that"verify" is now pretty broad. Anyway, nitpick and non-blocking.

hiddeco (approved):

action/action.yml
stefanprodan: Updated as you suggested ✔️ 

Just noticed that these steps don't show up in GH UI 🙃

hiddeco (approved):

PR #932, opened 2021-02-16
Snyk GitHub Action added

idvoretskyi:
Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/.github/workflows/snyk.yaml
+++ b/.github/workflows/snyk.yaml
@@ -1,0 +1,15 @@
+name: Snyk 
+on: [push]
+jobs:
+  security:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Run Snyk to check for vulnerabilities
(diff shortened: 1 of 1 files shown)


idvoretskyi:
cc @caniszczyk 

We'll have to add the SNYK_TOKEN as a secret to the GitHub repo here.

caniszczyk:
LGTM, token added

hiddeco:
Possible that the action fails because the token is not accessible from a fork?

idvoretskyi:
@hiddeco yep, we've just figured it out

caniszczyk:
fix the DCO @idvoretskyi :)

hiddeco:
It would also be awesome if the commits could be smashed, given this is a relative small change / addition.

idvoretskyi:
@hiddeco yep, fixed and squashed

hiddeco (approved):
Thank you @idvoretskyi and @caniszczyk :100:

Issue #919, opened 2021-02-14
[Docs] Flux Sealed Secrets / kubeseal --controller-namespace not mentioned


Sijoma:
Hi guys,

it seems the documentation on sealed secrets is missing a simple flag:

The docs state [Sealed Secret Doc](https://toolkit.fluxcd.io/guides/sealed-secrets/):
```
kubeseal --format=yaml --cert=pub-sealed-secrets.pem \
< basic-auth.yaml > basic-auth-sealed.yaml
```

I believe the **--controller-namespace** argument is required (mentioned in bitnami docs & decryption did not work for me)
```
kubeseal --controller-namespace flux-system --format=yaml --cert=pub-sealed-secrets.pem \
< basic-auth.yaml > basic-auth-sealed.yaml
```

adrian:
@Sijoma The `--controller-namespace` and `--controller-name` parameters are only needed when the `--cert` parameter isn't used. In that case kubeseal pulls down the public key from the controller to encrypt the secret. If the public key is available locally then only the `--cert` parameter is required.

If you're following along with [the docs](https://toolkit.fluxcd.io/guides/sealed-secrets/) then you'll have downloaded the public key with this command,
```
kubeseal --fetch-cert \
--controller-name=sealed-secrets \
--controller-namespace=flux-system \
> pub-sealed-secrets.pem
```

stefanprodan:
> decryption did not work for me

My guess is that you haven't fetched the `pub-sealed-secrets.pem` from your cluster. I'm closing this, but feel free to comment here.

Sijoma:
I fetched it from my Cluster. However, the cluster was not able to decrypt the secrets. 

The flag is mentioned by kubeseal docs.

Sijoma:
I followed the docs. I also have downloaded the cert to encrypt my secrets. However the cluster was not able to decrypt them. 

I copy & pasted everything. I did not even try to write the commands myself.

PR #933, opened 2021-02-16
A few minor updates to sealed secrets docs

adrian:
- Update to sealed secrets 1.13 helm chart. 1.10 is no longer available.
- Use `--dry-run=client` with kubectl. `--dry-run` on its own is deprecated.

### DIFF
--- a/docs/guides/sealed-secrets.md
+++ b/docs/guides/sealed-secrets.md
@@ -52,10 +52,10 @@ flux create helmrelease sealed-secrets \
 --target-namespace=flux-system \
 --source=HelmRepository/sealed-secrets \
 --chart=sealed-secrets \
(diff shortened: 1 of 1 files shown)


hiddeco:
Am I understanding correctly that this kind of solves #919?

adrian:
As far as I can tell the problem reported in #919 is down to a misunderstanding. I’ve left a note there to explain.

This PR doesn’t fix a reported issue. It’s just some tidying up.

adrian:
Although anyone following along with the docs will hit a problem since the 1.10 sealed secrets helm chart is no longer in the bitnami helm repo. So this PR does fix that.

hiddeco (approved):
Thank you for keeping our documentation up to date :pray:

PR #944, opened 2021-02-18
Split scan workflow into seperate CodeQL, FOSSA, Snyk jobs

hiddeco:
So that it also runs on the `main` branch and can properly compare.

Fixes:

> Warning: Code scanning cannot determine the alerts introduced or fixed by this pull request, because 1 analysis was not found.

https://github.com/fluxcd/flux2/runs/1926439471


hiddeco:
@stefanprodan this does not seem to work as

```
Error: Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job
Error: Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job
```

I think we need to split them again.

stefanprodan:
> only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job

😢 but way... each task has an unique id 

Ok let's pull Snyk out.

hiddeco:
I think it would be better to have them all split in that case, as there is little FOSSA and CodeQL have in common (while Snyk and CodeQL do).

.github/workflows/bootstrap.yaml
```diff
@@ -2,8 +2,9 @@ name: bootstrap
 
 on:
   push:
-    branches:
-      - '*'
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
stefanprodan, line 7: Please undo this, we can't share our GH admin token with contributor.

stefanprodan (changes requested):

stefanprodan (approved):

PR #974, opened 2021-02-23
Add example podMonitor for prometheus

tarioch:


stefanprodan (changes requested):
I propose we list a single object e.g.

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: source-controller
  namespace: flux-system
spec:
  namespaceSelector:
    matchNames:
      - flux-system
  selector:
    matchLabels:
      app: source-controller
  podMetricsEndpoints:
  - port: http-prom
```

And we tell people to create one for each controller they've installed.

tarioch:
good idea, I adjusted it

stefanprodan:
@tarioch please use my example yaml

stefanprodan (approved):
LGTM

Thanks @tarioch 🏅

PR #984, opened 2021-02-24
Publish flux binary for Apple Silicon

stefanprodan:
Changes:
- update Go to v1.16 in go.mod and CI
- add darwin/arm64 build to goreleaser config


hiddeco (approved):
Assuming you have confirmed locally that it compiles as expected.

hiddeco:
```
flux2 on  main [$] via 🐹 v1.16 
❯ CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build -o ./bin/flux-darwin-arm64 ./cmd/flux
flux2 on  main [$] via 🐹 v1.16 took 25s 
```

:heavy_check_mark:, awesome @stefanprodan :beers:

stefanprodan:
I verified it with `$ goreleaser release --snapshot` but idea if it actually runs as I don't have a M1.

PR #987, opened 2021-02-25
Embed the Kubernetes manifests in flux binary

stefanprodan:
This PR allows using the flux `install` and `bootstrap` commands in environments without access to GitHub releases (ref: https://github.com/fluxcd/flux2/discussions/913). 

Changes:
- add make target for generating the manifests using kustomize
- embed the generated manifests in flux binary
- use the embedded manifests by default in install and bootstrap commands
- download the manifests from GitHub only if the install/bootstrap `--version` arg is set to `latest` or to a value that's different to the CLI version

### DIFF
--- a/manifests/scripts/bundle.sh
+++ b/manifests/scripts/bundle.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Flux authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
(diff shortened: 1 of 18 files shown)


manifests/scripts/bundle.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# build policies
+policies_path="${REPO_ROOT}/manifests/policies"
+policies_output_path="${OUT_PATH}/policies.yaml"
+build $policies_path $policies_output_path
+files+=" $(basename $policies_output_path)"
+
+# create tarball
+if [[ -n $TAR ]];then
+    info "archiving $TAR"
```
hiddeco, line 68:
```suggestion
  info "archiving $TAR"
```

pkg/manifestgen/install/install.go
```diff
@@ -35,17 +35,13 @@ import (
 // Generate returns the install manifests as a multi-doc YAML.
 // The manifests are built from a GitHub release or from a
 // Kustomize overlay if the supplied Options.BaseURL is a local path.
-func Generate(options Options) (*manifestgen.Manifest, error) {
+func Generate(options Options, manifestsBase string) (*manifestgen.Manifest, error) {
```
hiddeco, line 38: Any specific reason everything is an option except for the base?

stefanprodan: Because the base can be created by the CLI only, since you can't embed files in a module and TF can't deal with it.

stefanprodan: Let me put this way, the install.Options are meant to be mapped in TF/eksctl/etc as user-supplied inputs, the `manifestsBase` doesn't fit into this category.

hiddeco (approved):
LGTM, awesome work @stefanprodan :sparkles:

PR #1008, opened 2021-02-27
faq: Can I use Flux HelmReleases without GitOps?

stefanprodan:
This PRs adds an entry to the FAQ for users who want to manage Helm releases with kubectl instead of git.


hiddeco (commented):
Can you add a link to this in the upgrade guide as well?

hiddeco (approved):

Issue #1006, opened 2021-02-26
Error on Cli command in Windows


andreracz:
### Describe the bug

I'm following the installations documents using windows cli at https://github.com/fluxcd/flux2/blob/main/docs/guides/installation.md, and when trying to create a customization with the command ```flux create kustomization flux-system   --source=flux-system   --path="./clusters/dev"   --prune=true   --interval=10m``` the following error occurs:

✗ kustomization path not found: stat /tmp/flux-system046498242/clusters\dev: no such file or directory

### To Reproduce

Steps to reproduce the behaviour:

1. In windows, run command: ``flux create kustomization flux-system   --source=flux-system   --path="./clusters/dev"   --prune=true   --interval=10m```

### Expected behavior

Kustomization should be created

### Additional context

- Kubernetes version: 1.18.10
- Git provider: Azure Devops
- Container registry provider: Docker Hub

Below please provide the output of the following commands:

```cli
> flux --version
flux version 0.9.0

> flux check
► checking prerequisites
✔ kubectl 1.19.3 >=1.18.0-0
✔ Kubernetes 1.18.10 >=1.16.0-0
► checking controllers
✔ helm-controller: healthy
► ghcr.io/fluxcd/helm-controller:v0.8.0
✔ kustomize-controller: healthy
► ghcr.io/fluxcd/kustomize-controller:v0.9.1
✔ notification-controller: healthy
► ghcr.io/fluxcd/notification-controller:v0.9.0
✔ source-controller: healthy
► ghcr.io/fluxcd/source-controller:v0.9.0
✔ all checks passed

> kubectl -n flux-system get all
NAME                                           READY   STATUS    RESTARTS   AGE
pod/helm-controller-657d76475b-jqxjx           1/1     Running   0          3h4m
pod/kustomize-controller-54fdf84b79-v9d5s      1/1     Running   0          3h4m
pod/notification-controller-769876bb9f-lfmxq   1/1     Running   0          3h4m
pod/source-controller-c55db769d-bbf69          1/1     Running   0          3h4m

NAME                              TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/notification-controller   ClusterIP   10.0.91.249    <none>        80/TCP    3h4m
service/source-controller         ClusterIP   10.0.169.143   <none>        80/TCP    3h4m
service/webhook-receiver          ClusterIP   10.0.204.209   <none>        80/TCP    3h4m

[...]

PR #1022, opened 2021-03-02
Use path with '/' slashes in created Kustomization

hiddeco:
This fixes a bug on Windows where the safe relative path would contain
'\' slashes, which are not compatible with the controller.

Fixes: #1006

### DIFF
--- a/cmd/flux/create_kustomization.go
+++ b/cmd/flux/create_kustomization.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


stefanprodan:
Doesn't this affect HelmReleases? e,g. `flux create hr --chart=./charts/podinfo`

hiddeco:
No, as the `spec.chart` value can be both a string (name) and path there is no complex safe path type: https://github.com/fluxcd/flux2/blob/main/cmd/flux/create_helmrelease.go#L108

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1028, opened 2021-03-03
Update installation guide for Azure DevOps GitRepository Source

maxgio92:
Update installation guide updating `username` to be empty inside tip for Azure DevOps with HTTP basic authentication, in order to reflect the documentation on Azure DevOps about using Personal Access Token for HTTP authentication:
https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page#use-a-pat


hiddeco (commented):
From that same page:

> **Git interactions require a username**, which can be anything except the empty string.

stefanprodan:
I highly doubt this change is needed, we have several users that are running Flux on Azure DevOps.

maxgio92:
I'm very sorry for this

PR #1035, opened 2021-03-04
don't default to prune=yes in migration documents. 

pmatheson-greenphire:
When following these instructions migrating an existing cluster you can accidentally prune resources if you're not paying attention. This could cause a catastrophic loss of cluster resources and ruin some ones day.

### DIFF
--- a/docs/guides/flux-v1-migration.md
+++ b/docs/guides/flux-v1-migration.md
@@ -169,7 +169,7 @@ Configure the reconciliation of the `deploy` dir on your cluster:
 $ flux create kustomization app \
   --source=app \
   --path="./deploy" \
-  --prune=true \
+  --prune=false \
   --interval=10m
 ✚ generating Kustomization
 ► applying Kustomization
@@ -244,7 +244,7 @@ Configure the reconciliation of the `prod` overlay on your cluster:
 flux create kustomization app \
   --source=GitRepository/app \
   --path="./overlays/prod" \
-  --prune=true \
+  --prune=false \
   --interval=10m
 ```
 


stealthybox:
@pmatheson-greenphire good catch here; we should make sure this is called out since it was a more experimental feature with Flux v1.

> you can accidentally prune resources if you're not paying attention.

Considering pruning only occurs when you remove objects from a repo, do you still think this could easily happen by accident during a migration?

Perhaps we could instruct the person doing the migration to enable prune afterward?
I think a lot of folks have found it to be an option with pretty sensible behavior.

pmatheson-greenphire:
The way I used the garbage collect feature is Flux v1 was to only have it enabled for some namespaces (i.e. staging or testing overlays). If you're using a cluster wide configuration repository prune=yes means that *everything*  on the cluster must be in git. This doesn't work well for a development environments because you need to be able test some workloads prior to them being committed. Also, because it's development the convenience of turning off garbage collection outweighs the need to eventually clean up some workloads. We can just delete the namespace and have flux rebuild it when it gets too messy. 

When I copied/pasted  the upgrade instructions I didn't notice "prune=yes" and I had a typo in the path. This pointed Flux at an empty directory and completely wiped out all resources on the cluster.

yebyen:
What documents describe a cluster wide configuration repository @pmatheson-greenphire ? I'm not familiar with this configuration as you described it, and I don't recognize it from any FluxCD.io docs.

pmatheson-greenphire:
> What documents describe a cluster wide configuration repository @pmatheson-greenphire ? I'm not familiar with this configuration as you described it, and I don't recognize it from any FluxCD.io docs.

https://github.com/fluxcd/multi-tenancy

kingdonb:
I was referring to this line from your earlier comment:

> If you're using a cluster wide configuration repository prune=yes means that everything on the cluster must be in git.

I don't think this is true, and I'm not sure where you got that idea. Clusters generally come pre-installed with some services like kubedns or cluster network, or even etcd, that are installed ahead of gitops configuration inside the cluster. Even clusters that are fully gitops managed like WKS don't always keep every resource in git, some resources are baseline and others are managed by operators. It is not expected that you will record every last resource in a Flux repo, this would never work with operators. Otherwise some resources are managed by `kubeadm` and they get upgraded when cluster nodes and control plane are upgraded.

So in a strict sense, those resources may be represented by some ClusterAPI resource as a side-effect, but their details usually will not be enumerated in the repo, they are managed by some operator and they will not be deleted by prune. Only resources which are created by Flux are pruned or garbage collected as they are removed from the repository source.

I'm not sure where you got a different idea. The repo you linked doesn't appear to say any such thing as far as I can tell. It is also archived, as it is now outdated, that repo only describes multi-tenancy on Flux v1. Current multi-tenancy docs are at https://github.com/fluxcd/flux2-multi-tenancy

pmatheson-greenphire:
> Only resources which are created by Flux are pruned or garbage collected as they are removed from the repository source.

I see my confusion - it wasn't that flux didn't know about the resources and deleted them. It just saw a bunch of resources (created by flux v1) that didn't exist in my typo'd kustomzation and got rid of them.

stefanprodan:
Docs have moved to the [website](https://github.com/fluxcd/website) repo, please open a PR there.

PR #1048, opened 2021-03-05
Restore default key algorithm flag create source

hiddeco:
This was removed by accident in the PR that introduced the new
`manifestgen` packages, and now restored in full glory.

### DIFF
--- a/cmd/flux/create_source_git.go
+++ b/cmd/flux/create_source_git.go
@@ -122,6 +122,7 @@ func init() {
 
 func newSourceGitFlags() sourceGitFlags {
 	return sourceGitFlags{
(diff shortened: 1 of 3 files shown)


cmd/flux/create_source_git.go
```diff
@@ -122,6 +122,7 @@ func init() {
 
 func newSourceGitFlags() sourceGitFlags {
 	return sourceGitFlags{
+		keyAlgorithm:  "rsa",
```
stefanprodan, line 125: Can't we use here the RSAPrivateKeyAlgorithm from the manifestgen package? this way we don't hardcode it twice

stefanprodan (approved):
LGTM

Thanks @hiddeco 🍻

PR #1052, opened 2021-03-05
Add detail to SOPS / Azure guide

apeschel:
Provide a more in-depth guide on how to set the environment variables
required by SOPS to access an Azure Key Vault.

### DIFF
--- a/docs/guides/mozilla-sops.md
+++ b/docs/guides/mozilla-sops.md
@@ -141,6 +141,55 @@ When using Azure Key Vault you need to authenticate the kustomize controller eit
 [Service Principal credentials as environment variables](https://github.com/mozilla/sops#encrypting-using-azure-key-vault)
 or with [add-pod-identity](https://github.com/Azure/aad-pod-identity).
 
+There are several authentication methods available in SOPS for connecting to an
+Azure Key Vault. SOPS looks for specific environment variables to determine
+which method to use, and then uses the credentials in those environment
+variables. Please refer to the SOPS documentation to determine which
+environment variables you will need to set for your preferred authentication
+method.
+
+For example, to use a service principal for authentication, you would need to
+have these environment variables set for SOPS:
+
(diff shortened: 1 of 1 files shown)


docs/guides/mozilla-sops.md
```diff
@@ -141,6 +141,55 @@ When using Azure Key Vault you need to authenticate the kustomize controller eit
...
+  --from-literal=AZURE_TENANT_ID="XXX" \
+  --from-literal=AZURE_TENANT_ID="XXX" \
+  --from-literal=AZURE_TENANT_ID="XXX"
+```
+
+You'll need a separate process from Flux for bootstrapping this specific secret
+before you bootstrap Flux, or you'll end up with a dependency cycle.
+
+Finally, update your kustomize controller deployment definition in
+`flux-system/gotk-components.yaml` to mount the secret data as environment
```
stefanprodan, line 177: Please see the docs on how to do this without altering the bootstrap files https://toolkit.fluxcd.io/guides/installation/#customize-flux-manifests. Manual edits to the components manifests means that users can no longer upgrade Flux automatically.

stefanprodan (changes requested):

stealthybox:
@apeschel I'm sorry I didn't see this patch before, or I would have worked with you on it.

I just PR'd some overlapping docs on SOPS + Flux in #1064

docs/guides/mozilla-sops.md
apeschel: @stefanprodan Thanks, this makes a lot of sense! I was wondering if there was a way to do this, but hadn't found the guide.

apeschel:
@stealthybox Your PR looks good, I'd be willing to coordinate with you still, if you're willing. I should have some time in the next few days.

stefanprodan:
Closing this is in favour of #1064

PR #1062, opened 2021-03-08
Improve build process embedded manifests

hiddeco:
This commit changes the way the build of manifests is triggered by
making smarter use of the capabilities of Make. The result should be
that the manifests are only regenerated if:

1. There is no `cmd/flux/manifests/` directory.
2. There have been made changes to the YAML files in the `manifests/`
   directory that are newer than the files in `cmd/flux/manifests/`.


stefanprodan (approved):
LGTM 

Thanks @hiddeco

PR #1065, opened 2021-03-08
Add Go 1.16 to prerequisites (contributing doc)

stefanprodan:
Changes to contributing doc:
- add image automation repos to GOTK list
- add prerequisites for running the tests locally

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,11 +48,13 @@ you might want to take a look at the [introductory talk and demo](https://www.yo
 
 This project is composed of:
 
-- [/f/flux2](https://github.com/fluxcd/flux2): The Flux CLI
-- [/f/source-manager](https://github.com/fluxcd/source-controller): Kubernetes operator for managing sources
-- [/f/kustomize-controller](https://github.com/fluxcd/kustomize-controller): Kubernetes operator for building GitOps pipelines with Kustomize
-- [/f/helm-controller](https://github.com/fluxcd/helm-controller): Kubernetes operator for building GitOps pipelines with Helm
-- [/f/notification-controller](https://github.com/fluxcd/notification-controller): Kubernetes operator for handling inbound and outbound events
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,10 +26,10 @@ build: $(EMBEDDED_MANIFESTS_TARGET)
 install:
 	go install cmd/flux
 
+install-dev: $(EMBEDDED_MANIFESTS_TARGET)
+	CGO_ENABLED=0 go build -o /usr/local/bin ./cmd/flux
```
hiddeco, line 30: This has never worked on my system, as `/usr/local/bin` is protected and I do not run my `make` commands with `sudo`:

```console
$ make install-dev
CGO_ENABLED=0 go build -o /usr/local/bin ./cmd/flux
go build github.com/fluxcd/flux2/cmd/flux: copying /tmp/go-build3338692359/b001/exe/a.out: open /usr/local/bin/flux: permission denied
make: *** [Makefile:35: install-dev] Error 1
```

If you change it to something like this, it will prompt for a password, while (hopefully) not creating any problems on MacOS:

```suggestion
install-dev: $(EMBEDDED_MANIFESTS_TARGET)
	CGO_ENABLED=0 go build -o /tmp/flux-dev ./cmd/flux
	sudo install -Dm755 /tmp/flux-dev /usr/local/bin/flux
```

stefanprodan: Seems that we are in a deadlock here, you don't want sudo make and I don't want make sudo.

hiddeco: `sudo make` would make Go run as root, which is not something you want as all Go Modules are cached for the normal user.

My suggestion makes this target work on any system, and the issue you have with it for MacOS can be overcome by running `brew` with the `--overwrite` flag. Given all this, I see much stronger arguments in favor of my suggestion, than against.

stefanprodan: I've undone the changes to Makefile

hiddeco (approved):

PR #1066, opened 2021-03-08
Make manifests dir `bundle.sh` configurable

hiddeco:
There was an assumption in this script that it is always executed in Git
repository/directory, this is however not always true, for example when
one downloads the `.tar.gz` that is made available for every release
by GitHub (and used in one of our AUR packages).

This commit changes this, and makes the first argument of `bundle.sh`
configurable, so a custom manifests directory can always be defined
_without_ relying on Git.

Omitting it, or passing an empty string, will still fall back to the
previous behavior of using `git rev-parse --show-toplevel`.

### DIFF
--- a/.github/aur/flux-go/PKGBUILD.template
+++ b/.github/aur/flux-go/PKGBUILD.template
@@ -12,7 +12,7 @@ provides=("flux-bin")
 conflicts=("flux-bin")
 replaces=("flux-cli")
 depends=("glibc")
-makedepends=("go")
+makedepends=('go>=1.16', 'kustomize>=3.0')
 optdepends=('kubectl: for apply actions on the Kubernetes cluster',
 'bash-completion: auto-completion for flux in Bash',
 'zsh-completions: auto-completion for flux in ZSH')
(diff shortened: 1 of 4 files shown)


.github/aur/flux-go/PKGBUILD.template
```diff
@@ -12,7 +12,7 @@ provides=("flux-bin")
 conflicts=("flux-bin")
 replaces=("flux-cli")
 depends=("glibc")
-makedepends=("go")
+makedepends=('go>=1.16', 'kustomize>=3.9', 'kustomize<4.0.0')
```
hiddeco, line 15: @stefanprodan if versions lower than `3.9.x` work fine, please say so and I will make the range less strict.

stefanprodan: The manifests can be built with kustomize v4 since flux 0.9. The minimum supported version is 3.0 as we use `transformers`.

hiddeco: Thank you, updated the build constraints.

relu (approved):
LGTM :heavy_check_mark:

stefanprodan (approved):

PR #1083, opened 2021-03-11
WIP: GitHub actions use case (manifest generation)

kingdonb:
At the time of this writing, the new document I've added called "GitHub Actions Manifest Generation" has most of the content for the first four examples; it is still in need of work before I'll call for reviewers. WIP

The working examples that I've developed are also live and in context in the test repo at https://github.com/kingdonb/any_old_app/tree/main/.github/workflows – at this point you may still need additional context from this repo that I haven't provided in the document just yet. After I think I've provided all the context needed inline, I will begin testing first by myself, and then when it seems self-contained, call for others to try my examples and test them too.

Resolves fluxcd/flux#3433

### DIFF
--- a/docs/guides/gh-actions-manifest-generation.md
+++ b/docs/guides/gh-actions-manifest-generation.md
@@ -1,0 +1,544 @@
+# GitHub Actions Manifest Generation
+
+This is an example implementation of "build-time" manifest generation on GitHub Actions.
+
+We first use third-party tools to prepare and update YAML manifests in a CI job. Then changes are committed to a deploy branch from CI and finally pushed, where `kustomize-controller` can apply them to a cluster.
+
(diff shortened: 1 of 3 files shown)


kingdonb:
[kingdonb/any-old-app](https://github.com/kingdonb/any_old_app/) is mentioned in here a couple of times because that is the repo where I worked on and developed this strategy of workflow.

Also big thanks to Paul on the CNCF #flux slack channel that got me started with an example that he developed first, for `kubecfg` (which I believe is also based on jsonnet.) His example was at [PaulRudin/xamaral](https://github.com/PaulRudin/xamaral/blob/main/.github/workflows/gen-k8s.yaml)

kingdonb:
I will close this out in favor of #1200

PR #1091, opened 2021-03-12
Fix deployment name in image update guide 

joebowbeer:
The filename is `podinfo-deployment.yaml`


stefanprodan (approved):
LGTM

Thanks @joebowbeer good catch!

PR #1116, opened 2021-03-18
Implement get all for sources and images

stefanprodan:
This PR partially addresses https://github.com/fluxcd/flux2/discussions/1114

Changes:
- add `flux get sources all` command
- add `flux get images all` command

Examples:

```console
$ flux get images all
NAME                   	READY	MESSAGE                       	LAST SCAN                	SUSPENDED 
imagerepository/podinfo	True 	successful scan, found 16 tags	2021-03-18T12:38:31+02:00	False    	

NAME               	READY	MESSAGE                                                               	LATEST IMAGE                       
imagepolicy/podinfo	True 	Latest image tag for 'ghcr.io/stefanprodan/podinfo' resolved to: 5.0.3	ghcr.io/stefanprodan/podinfo:5.0.3
```

```console
$ flux get sources all --all-namespaces
NAMESPACE  	NAME                 	READY	MESSAGE                                                         	REVISION                                      	SUSPENDED 
flux-system	gitrepository/podinfo	True 	Fetched revision: 5.1.4/e43ebfa5bf4b87c46f2e1db495eb571cd398e2f7	5.1.4/e43ebfa5bf4b87c46f2e1db495eb571cd398e2f7	False    	

NAMESPACE  	NAME                  	READY	MESSAGE                                                   	REVISION                                	SUSPENDED 
flux-system	helmrepository/bitnami	True 	Fetched revision: 93687fb96ed93edd4dad9e6702fd9b2b73d8f3fb	93687fb96ed93edd4dad9e6702fd9b2b73d8f3fb	False    


```


cmd/flux/get.go
```diff
@@ -100,17 +108,26 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	getAll := cmd.Use == "all"
+
 	if get.list.len() == 0 {
-		logger.Failuref("no %s objects found in %s namespace", get.kind, rootArgs.namespace)
+		if !getAll {
+			logger.Failuref("no %s objects found in %s namespace", get.kind, rootArgs.namespace)
```
hiddeco, line 115: Wondering if it would be better to use `%q` instead of `%s` to highlight what the variables are.

cmd/flux/get.go
```diff
@@ -100,17 +108,26 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
...
+		row := get.list.summariseItem(i, getArgs.allNamespaces, getAll)
 		rows = append(rows, row)
 	}
 	utils.PrintTable(os.Stdout, header, rows)
+
+	if getAll {
+		fmt.Println()
+	}
 	return nil
+
```
hiddeco, line 132:
```suggestion
```

cmd/flux/get.go
stefanprodan: The kind is not an arg only the namespace is, I think the `%q` change has to be done all over the place not only here.

hiddeco: Yes, agreed. I started doing it for the bootstrap commands, but we may indeed want to do one big sweep to settle it everywhere.

hiddeco (approved):

PR #1138, opened 2021-03-22
Add source namespace arg to create commands

stefanprodan:
This PR adds an optional arg `--source-namespace` to `flux create kustomization` and `flux create helmrelease` commands.


hiddeco (commented):
I do not recall what the precise string format is, but I think there is one that looks something like `namespace:kind/name` for FQ resource IDs.

Wondering if it would be better to extend the `source` argument to support this format instead of adding a new flag?

PR #1141, opened 2021-03-22
Add frontmatter to command documentation

### LINKED ISSUE #1123: Add frontmatter to cli reference
For the new website we are going to need frontmatter added to the cli reference. @hiddeco already found this possible to implement using https://github.com/spf13/cobra/blob/master/doc/md_docs.md#customize-the-output

Frontmatter entries should roughly look like this:

https://github.com/fluxcd/website/pull/98/files#diff-2d810b85b1bef829d62f2e710c8c804d66d876243ca6e23fbd19513fd63a1b07

hiddeco:
Fixes #1123

### DIFF
--- a/docs/cmd/flux.md
+++ b/docs/cmd/flux.md
@@ -1,3 +1,6 @@
+---
+title: "flux command"
+---
 ## flux
 
 Command line utility for assembling Kubernetes CD pipelines
@@ -69,7 +72,7 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.
 ```
       --context string      kubernetes context to use
   -h, --help                help for flux
-      --kubeconfig string   path to the kubeconfig file (default "~/.kube/config")
+      --kubeconfig string   absolute path to the kubeconfig file
   -n, --namespace string    the namespace scope for this operation (default "flux-system")
       --timeout duration    timeout for this operation (default 5m0s)
       --verbose             print generated objects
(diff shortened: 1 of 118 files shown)


dholbach:
I just built the docs from this and `cp -r` them into my local fluxcd/website#110 branch and things look fine. We'll need a `_index.md` file too, but we can add that in the website branch I guess.

dholbach (approved):
Tested this, works well AFAICS.

hiddeco:
Requesting an additional review from Stefan as I moved some code logic around.

docs/cmd/flux.md
```diff
@@ -69,25 +72,25 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.
...
-* [flux delete](flux_delete.md)	 - Delete sources and resources
-* [flux export](flux_export.md)	 - Export resources in YAML format
-* [flux get](flux_get.md)	 - Get the resources and their status
-* [flux install](flux_install.md)	 - Install or upgrade Flux
-* [flux logs](flux_logs.md)	 - Display formatted logs for Flux components
-* [flux reconcile](flux_reconcile.md)	 - Reconcile sources and resources
-* [flux resume](flux_resume.md)	 - Resume suspended resources
-* [flux suspend](flux_suspend.md)	 - Suspend resources
-* [flux uninstall](flux_uninstall.md)	 - Uninstall Flux and its custom resource definitions
+* [flux bootstrap](/cmd/flux/flux_bootstrap/)	 - Bootstrap toolkit components
```
stefanprodan, line 83: Doesn't this break the links in mkdocs?

stefanprodan: All links are broken when you click on them from GitHub UI, I guess mkdocs fails the same.

hiddeco: I tested this with `mkdocs serve` and it works fine, Hugo is unable to parse `.md` links without them being wrapped in [`ref` or `relref`](https://gohugo.io/content-management/cross-references/).

dholbach: Yes, how about `[flux bootstrap](/cmd/flux_bootstrap/)`? That should work both in `mkdocs` and `hugo` land.

dholbach: https://www.docsy.dev/docs/best-practices/site-guidance/#linking

@alisondy and I talked about this as well.

hiddeco: Ah, that is what it was supposed to be. Let me push a fix in a couple minutes.

stefanprodan: @dholbach @hiddeco do you think is ok to have broken links when people read the docs with GitHub UI? I expect users to read the docs on our website, so maybe this is acceptable.

hiddeco: I see no other way at the moment, we may opt to use the `ref` thing once everything has been moved over so that they are more obviously broken.

stefanprodan (approved):

PR #1157, opened 2021-03-25
Allow supplying PK from file for Git source/secret

hiddeco:
This commit adds support for supplying a path to an existing private
key file to both the `flux create secret git` and `flux create source
git`
commands.

If a path is given, any private key generation configuration options
are ignored by the manifest generator. The SSH host will however still
be scanned for server keys.

### DIFF
--- a/cmd/flux/create_secret_git.go
+++ b/cmd/flux/create_secret_git.go
@@ -34,8 +34,7 @@ import (
 var createSecretGitCmd = &cobra.Command{
 	Use:   "git [name]",
 	Short: "Create or update a Kubernetes secret for Git authentication",
-	Long: `
-The create secret git command generates a Kubernetes secret with Git credentials.
+	Long: `The create secret git command generates a Kubernetes secret with Git credentials.
 For Git over SSH, the host and SSH keys are automatically generated and stored in the secret.
 For Git over HTTP/S, the provided basic authentication credentials are stored in the secret.`,
 	Example: `  # Create a Git SSH authentication secret using an ECDSA P-521 curve public key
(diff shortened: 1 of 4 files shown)


stefanprodan:
> The SSH host will however still be scanned for server keys.

We could add an example to both commands and mention the key scan there.

hiddeco:
Always forget there are examples embedded into the commands, will update it later tonight, thanks for the suggestion @stefanprodan :+1:

cmd/flux/create_secret_git.go
```diff
@@ -45,6 +45,12 @@ For Git over HTTP/S, the provided basic authentication credentials are stored in
     --ssh-key-algorithm=ecdsa \
     --ssh-ecdsa-curve=p521
 
+  # Create a Git SSH authentication secret with a private key from file
+  # The public SSH host key will still be gathered from the host
+  flux create secret git podinfo-auth \
+    --url=ssh://git@github.com/stefanprodan/podinfo \
+    --private-key-file=./key
```
stefanprodan, line 52: Is the key a dir? Can we say more about what should be in there?

stefanprodan: I would use key.pem also mention that it must a key without password?

hiddeco: PEM files are OpenSSL public keys, or key chains (and are defined in an RFC), so this wouldn’t be right.

Closest would probable be .key which may contain just a private key.

Note that the key you normally would generate with defaults (still `id_rsa`) is  extensionless.

stefanprodan: Ok so maybe id_rsa is a better example

stefanprodan (approved):
LGTM

Thanks @hiddeco ☕

PR #1163, opened 2021-03-26
Fix cmd links

dholbach:
For the move to the `hugo` site, I noticed that links in the CLI docs didn't quite work. This was due the new URL having an additional `/docs/` in there. To keep CLI docs working for both sites, moving to relative links.

Fixed a link in the components overview doc as well while I was at it.

### DIFF
--- a/cmd/flux/docgen.go
+++ b/cmd/flux/docgen.go
@@ -32,8 +32,7 @@ title: "%s"
 `
 
 var (
-	cmdDocPath    string
-	cmdDocURLPath string
+	cmdDocPath string
 )
 
 var docgenCmd = &cobra.Command{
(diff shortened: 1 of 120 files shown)


cmd/flux/docgen.go
```diff
@@ -67,5 +67,5 @@ func frontmatterPrepender(filename string) string {
 
 func linkHandler(name string) string {
 	base := strings.TrimSuffix(name, path.Ext(name))
-	return cmdDocURLPath + strings.ToLower(base) + "/"
+	return "../" + strings.ToLower(base) + "/"
```
hiddeco, line 70: `cmdDocURLPath` is now no longer used.

dholbach: Well-spotted! :1st_place_medal:

hiddeco (approved):
Can you please smash (or group) your commits together? Other than that LGTM, thanks @dholbach :pie:

hiddeco:
You need to rerun `make test` and amend your commit.

PR #1171, opened 2021-03-28
Updated automation migration guide filter tags crd

Legion2:
The text description in the automation migration guide uses a wrong crd schema.


stefanprodan (approved):
LGTM

Thanks @Legion2

Issue #1125, opened 2021-03-19
`flux create source git` sshkeyscan struggles with split DNS


stealthybox:
As seen on https://tgik.io/149, if a user is using an in-cluster repo, the outside and inside DNS names may not match.
In the demo the Windows System32 `hosts` file could have been used to make the port-forwarded DNS name match the in-cluster Service but this wasn't done.

It seems like the `flux create source git` command generates the known_hosts file for the in-cluster Secret from the host (which is the easiest way to do it).

It would be possible to exec into one of the flux controller pods, to output the `known_hosts`, and use that instead? This means the `create source git`'s generation features would rely on that particular controller running, and we'd have to consider supporting a non-default controller-namespace or even Deployment name.

We could use a Job instead, but that would have similar concerns and necessitate that an image is pullable for air-gapped env's. (could just use a flux controller image)

Alternatively, an `--extra-known-hosts` flag that could duplicate the remote public key for multiple host-names to support. This would allow the kind of create->edit hostname workflow in the demo to work since the hosts file wouldn't be messed up.
This is a very incomplete solution that I don't think we should do, and it doesn't answer for why you wouldn't just edit the hosts field in the Secret yourself aside from how inconvenient that is.

stealthybox:
I was able to reproduce this.
You can create a source and generate a deploy key using the local DNS name, but the in cluster names for the `GitRepository` and `known_hosts` secret data field need to be fixed manually afterward.

I can confirm this workaround fixes the problem:
```shell
# install gitea in the default NS -- https://docs.gitea.io/en-us/install-on-kubernetes/
helm repo add gitea-charts https://dl.gitea.io/charts/
helm install gitea gitea-charts/gitea

# forward the web and ssh ports
kubectl port-forward sts/gitea 3000 2222:22

# create an account @ localhost:3000, then configure personal SSH/GPG keys
# create a new repo and initialize/clone it
```
```shell
# here we make the broken GitRepository related to this issue

# this fails to keyscan
flux create source git flux-system --url ssh://gitea-ssh.default/stealthybox/fleet --ssh-key-algorithm ed25519 --ssh-ecdsa-curve p521 --branch main

# this succeeds with keyscan, creating the GitRepo and Secret, but fails to fetch the repo naturally
flux create source git flux-system --url ssh://127.0.0.1:2222/stealthybox/fleet --ssh-key-algorithm ed25519 --ssh-ecdsa-curve p521 --branch main

[...]

PR #1179, opened 2021-03-30
Publish AMD64, ARM64, ARMv7 images for binary

hiddeco:
This PR adds the bits and pieces to enable the build of Docker images (and a multi-arch) image for goreleaser, publishing them to both Docker Hub and GHCR.

It offers a workaround for #1125 as described in https://github.com/fluxcd/flux2/issues/1125#issuecomment-806482915.

Release process was tested locally against a fork, see:

- https://hub.docker.com/r/hiddeco/flux-cli/tags
- https://github.com/users/hiddeco/packages/container/package/flux-cli

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,6 +16,26 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.x
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco ☕ 

PS. After the next release, we could add the a section to docs on how people can run bootstrap using a pod in cluster.

PR #1215, opened 2021-04-05
update sortable image tag guide with github.run_number

Frederik-Baetens:
Signed-off-by: Frederik Baetens <baetens.fr@gmail.com>

This makes the example a bit simpler, and I think many people will find github.run_number useful, because it saves them from having to define an additional step.

If you feel it's educational to still keep the branch name in the example, I will add that back in.

### DIFF
--- a/docs/guides/sortable-image-tags.md
+++ b/docs/guides/sortable-image-tags.md
@@ -36,6 +36,7 @@ $ date +%s
 Alternatively, you can use a serial number as part of the tag.  Some CI platforms will provide a
 build number in an environment variable, but that may not be reliable to use as a serial number --
 check the platform documentation.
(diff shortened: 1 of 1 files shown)


docs/guides/sortable-image-tags.md
```diff
@@ -65,8 +66,8 @@ The branch and tag will usually be made available in a CI platform as environmen
 
 ## Example of a build process with timestamp tagging
```
stealthybox, line 67: This is an awesome usage of `run_number` with GitHub actions, but perhaps we would want to add this as a separate example?

The header here already states that the example is specifically showing timestamps as a monotonically increasing number.

Frederik-Baetens: Thanks, I made it into a separate example!

stefanprodan (approved):
LGTM 

Thank you

Issue #1204, opened 2021-04-03
`flux install --export --verbose` modifies active cluster


nhinds:
### Describe the bug

Running `flux install --export --verbose` applies changes to the currently-active cluster in your kubeconfig, rather than exporting manifests to files

### To Reproduce

Steps to reproduce the behaviour:

1. Run `flux install --export --verbose`
2. Observe it modifying your current kubernetes cluster

### Expected behavior

I expected the `--export` flag to only print the manifests, and `--verbose` to maybe give me more information about what `flux` was doing while generating the manifests.

### Additional context

- Kubernetes version: N/A - happens even without a Kubernetes cluster, with `export KUBECONFIG=/dev/null` (get connection refused)
- Git provider: N/A - happens before git provider configured
- Container registry provider: N/A - happens before container registry configured

Below please provide the output of the following commands:

```cli
$ export KUBECONFIG=/dev/null
$ flux --version
flux version 0.11.0
$ flux check
► checking prerequisites
✔ kubectl 1.20.5 >=1.18.0-0
✗ Kubernetes client initialization failed: kubernetes configuration load failed: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
► checking controllers
$ kubectl -n flux-system get all
The connection to the server localhost:8080 was refused - did you specify the right host or port?
$ kubectl -n flux-system logs deploy/source-controller
The connection to the server localhost:8080 was refused - did you specify the right host or port?
$ kubectl -n flux-system logs deploy/kustomize-controller
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

PR #1219, opened 2021-04-06
fix: install command flags logic bug

frankgu968:
Resolves https://github.com/fluxcd/flux2/issues/1204

If the `export` flag exists, print the generated manifest and return from the function; it should prevent any additional steps from being run. 
If the `verbose` flag exists, then print the generated manifest and continue with the function.


hiddeco (approved):
LGTM, thank you @frankgu968 :sun_with_face:

Issue #1220, opened 2021-04-06
`flux create source git` silently drops username/password with HTTP


makkes:
### Describe the bug

When creating a GitRepository with an HTTP URL (not HTTPS) through the CLI, the username/password parameters are silently dropped without further notice. This causes the source creation to fail.

### To Reproduce

Steps to reproduce the behaviour:

```
$ flux create source git local2 --url http://git.gitserver.svc.cluster.local/repo.git
 --username=git --password=git123
✚ generating GitRepository source
► applying GitRepository source
✔ GitRepository source created
◎ waiting for GitRepository source reconciliation
✗ unable to clone 'http://git.gitserver.svc.cluster.local/repo.git', error: authentication required
```

### Expected behavior

I suggest something like this:

```
$ flux create source git local2 --url http://git.gitserver.svc.cluster.local/repo.git
 --username=git --password=git123
✗ WARNING: username/password are dropped for unencrypted HTTP sources

✚ generating GitRepository source
► applying GitRepository source
✔ GitRepository source created
◎ waiting for GitRepository source reconciliation
✗ unable to clone 'http://git.gitserver.svc.cluster.local/repo.git', error: authentication required
```

### Additional context

- Kubernetes version: n/a
- Git provider: n/a
- Container registry provider: n/a

Below please provide the output of the following commands:

```cli
flux --version
flux version 0.11.0
flux check
► checking prerequisites
✔ kubectl 1.21.0-beta.1 >=1.18.0-0
✔ Kubernetes 1.20.2 >=1.16.0-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.9.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.10.0
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.11.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.10.0
✔ all checks passed
```

PR #1222, opened 2021-04-07
Fix create source git auth for non-HTTPS repos

stefanprodan:
Fix: #1220


hiddeco:
I would include a warning that a HTTP URL was given with credentials, this may prevent people from accidentally creating a non-secure configuration.

hiddeco (approved):
Thanks @stefanprodan :100:

makkes:
Thanks @stefanprodan! That was quick. 😌

PR #968, opened 2021-02-23
Refactor bootstrap to make use of go-git-providers

hiddeco:
This pull request changes the implementation used to connect to Git
providers like GitHub from `fluxcd/pkg/git` to `fluxcd/go-git-providers`,
eliminating quite some bugs that were present in the GitLab integration.

In addition to this, the bootstrap logic has been factored out into an
(for now) internal package with two bootstrap "runners" `PlainGitBootstrapper`
and `GitProviderBootstrapper`. The `GitProviderBootstrapper` is a superset
of the `PlainGitBootstrapper` that besides `Reconciler` also implements the
`RepositoryReconciler`.

The Git actions rely on an interface, making it easier to support other
implementations than `go-git` at a later moment, to for example support
bootstrapping to Git servers that only support the v2 protocol.

By doing this, it was possible to introduce a new `flux bootstrap git`
command that bootstraps Flux to a plain Git repository and applies the
resources to the cluster, prompting the user to upsert the public key
by hand if a private key was generated.

As a finishing touch, almost all options are now configurable like they
already were for the `flux <install|create>` commands, making it possible
to have much finer grain configurations like e.g. the Git author that
makes the commit, the type of key that is generated, adding a `[skip ci]`
appendix to the commit message, and many more.

### Future enhancement possibilities:

- Assign different rights than `maintainer` to GitHub and GitLab teams
- Provide another Git implementation so bootstrapping to e.g. Azure DevOps is possible
- ...

### Help testing this PR :superhero:

1. Clone the repository and checkout the branch:

   ```console
   $ git clone https://github.com/fluxcd/flux2.git
   $ cd flux2
   $ git checkout -b go-git-providers-bootstrap origin/go-git-providers-bootstrap
   ```

2. Build the binary (requires Go 1.16 and Kustomize):

   ```console
   $ make build
   ```

3. Run one of the bootstrap commands:

   ```console
   $ ./bin/flux bootstrap git --help
   $ ./bin/flux bootstrap gitlab --help
   $ ./bin/flux bootstrap github --help
   ```

### Ref list

Fixes #799 
Fixes #835 
Fixes #916
Fixes #924
Fixes #1021
Fixes #1129
Fixes #1143

Addresses part of #43

PR #1231, opened 2021-04-08
Take private key from file into account in Git bootstrap

hiddeco:
Follow up on #968


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1242, opened 2021-04-09
Always report components health in bootstrap

hiddeco:


internal/bootstrap/bootstrap_plain_git.go
```diff
@@ -283,22 +282,33 @@ func (b *PlainGitBootstrapper) ReconcileSyncConfig(ctx context.Context, options
...
 	if err := wait.PollImmediate(pollInterval, timeout, kustomizationReconciled(
 		ctx, b.kube, client.ObjectKey{Name: options.Name, Namespace: options.Namespace}, &k, expectRevision),
 	); err != nil {
-		return fmt.Errorf("failed waiting for Kustomization: %w", err)
+		b.logger.Failuref(err.Error())
+		return err
 	}
 
-	b.logger.Successf("reconciled sync configuration")
+	b.logger.Successf("Kustomization reconciled")
```
stefanprodan, line 307:
```suggestion
	b.logger.Successf("Kustomization/%s reconciled", options.Name)
```

hiddeco: Added it to the first log message so that it is also logged when things go south.

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1255, opened 2021-04-12
Add Helm use case intro page

scottrigby:
What is the simplest way to explain how Flux works with Helm to existing Helm users?

There's a lot more we can say, but is there anything critical we think should go into an intro that I'm missing?

See https://github.com/fluxcd/website/issues/120

### DIFF
--- a/docs/use-cases/helm.md
+++ b/docs/use-cases/helm.md
@@ -91,6 +91,116 @@ Why is this important?
 If you or your team has ever collaborated with multiple engineers on one or more apps, and/or in more than one namespace or cluster, you probably have a good idea of how declarative, automatic reconciliation can help solve common problems.
 If not, or either way, you may want to check out this [short introduction to GitOps](https://youtu.be/r-upyR-cfDY).
 
+## Customizing Your Release
+
+While Helm charts are usually installable using default configurations, users will often customize charts with their preferred configuration by [overriding the default values](https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing).
+The Helm client allows this by imperatively specifying override values with `--set` on the command line, and in additional `--values` files. For example:
+
+```console
+helm install my-traefik traefik/traefik --set service.type=ClusterIP
+```
+
+and
+
+```console
+helm install my-traefik traefik/traefik --values ci/kind-values.yaml
+```
+
+where `ci/kind-values.yaml` contains:
+
+```yaml
+service:
+  type: ClusterIP
+```
+
+Flux Helm Controller allows these same YAML values overrides on the `HelmRelease` CRD.
+These can be declared directly in `spec.values`:
+
+```yaml
+spec:
+  values:
+    service:
+      type: ClusterIP
+```
+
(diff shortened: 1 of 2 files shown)


docs/use-cases/helm.md
```diff
@@ -0,0 +1,98 @@
+# Flux for Helm Users
+
+Welcome Helm users!
+We think Flux's Helm Operator is the best way to do Helm according to GitOps principles, and we're dedicated to doing what we can to help you feel the same way.
```
stefanprodan, line 4:
```suggestion
We think Flux's Helm Controller is the best way to do Helm according to GitOps principles, and we're dedicated to doing what we can to help you feel the same way.
```

stefanprodan (commented):
@scottrigby I think we could add a section on secrets management and explain how valuesFrom work with secrets decrypted by kustomize-controller. Another thing that's missing from here is the automated upgrade feature (semver range) and automated rollback.

scottrigby:
> @scottrigby I think we could add a section on secrets management and explain how valuesFrom work with secrets decrypted by kustomize-controller

Should I just summarize and link to the existing doc pages on that?
- https://toolkit.fluxcd.io/components/helm/helmreleases/#values-overrides
- https://toolkit.fluxcd.io/guides/mozilla-sops/
- also configMap doc: https://toolkit.fluxcd.io/guides/helmreleases/#refer-to-values-in-configmaps-generated-with-kustomize
- and a note on how these refs changed from Flux 1 -> 2: https://toolkit.fluxcd.io/guides/helm-operator-migration/#configmap-and-secret-references

stefanprodan:
> Should I just summarize and link to the existing doc pages on that?

Yes, for each topic we need to explain what's the GitOps way of doing it and why it's useful, then link to each doc.

> a note on how these refs changed from Flux 1 -> 2

This doc is about v2, I would not mention v1 at all.

scottrigby:
> Another thing that's missing from here is the automated upgrade feature (semver range) and automated rollback.

While there's a note about the feature here <https://toolkit.fluxcd.io/components/helm/controller/>, and the API pages have details, the best description for this is currently in the Helm Operator -> Controller migration doc https://toolkit.fluxcd.io/guides/helm-operator-migration/#defining-a-rollback-uninstall-configuration and https://toolkit.fluxcd.io/guides/helm-operator-migration/#automated-rollbacks.

Perhaps the bulk of this should be moved to the helm use case intro page, and that section could be linked to from the operator->controller migration page instead of the info living there.

stefanprodan:
> Perhaps the bulk of this should be moved to the helm use case intro page, and that section could be linked to from the operator->controller migration page instead of the info living there.

Let’s copy and adapt the rollback info in this doc. I want to avoid sending people from the migration guide to other places. The audience for this document is very different to the migration one, so duplicating some info is fine.

scottrigby:
@stefanprodan WDYT? Trying to cover bases with simple examples, while trying to duplicate as little as possible from other areas of the docs. I did follow your suggestion about copying adapt and rollback.

stefanprodan (approved):
LGTM 

Thanks @scottrigby 🏅

hiddeco (approved):
This is awesome, thank you very much @scottrigby :100: :1st_place_medal:

PR #1262, opened 2021-04-13
Put potentially destructive reconcile behind flag

hiddeco:
The behavior introduced during the introduction of go-git-providers
was more strict, and has proven pretty quickly to not be useful to
all users. Therefore, the reconciliation behavior for repository
configuration has been put behind an opt-in flag, so that it does
not overwrite people their configs by accident.

### DIFF
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -80,6 +80,7 @@ type githubFlags struct {
 	path         flags.SafeRelativePath
 	teams        []string
 	readWriteKey bool
+	reconcile    bool
 }
 
 const (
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco

Legion2:
What exactly does change when I opt-in this flag? Can you point to an issue or documentation?

yebyen:
I think this problem was only reported in slack.

This change ensures that if you bootstrap an existing public repo but fail to pass the private=false flag at bootstrap time, the CLI won't make your repo suddenly turn private (potentially shedding all of the public usage stats including stars and forks)

This was new behavior introduced in a recent release, and we had some fairly upset people reporting on this issue in the Slack, (if you have not been following release updates as closely then you may not have seen this behavior in any previous release.)

The release that updated to use go-git-providers was the first to include this (now changed) feature.

Legion2:
@yebyen Thanks for sharing this information.

PR #1224, opened 2021-04-07
Refactor reconcile and resume cmd for alert and receiver

SomtochiAma:
Part of: #291 

Signed-off-by: Somtochi Onyekwere <somtochionyekwere@gmail.com>

PR #1285, opened 2021-04-16
Ensure kustomization/helmrelease is reconciled when source is in a different namespace

### LINKED ISSUE #1283: --with-souce not work if kustomization and source in diffrent namespace.
--with-souce  not work if kustomization and source in diffrent namespace.

reconcile with command: 
`flux reconcile ks test-apps  -n  test --with-source`
output:
`gitrepositories.source.toolkit.fluxcd.io "apps" not found`

here is my config:
```
kind: Kustomization
metadata:
  name: test-apps
  namespace: test
spec:
  interval: 10m0s
  sourceRef:
    kind: GitRepository
    name: apps
    namespace: flux-system
  path: ./apps/test/
  prune: true
  validation: client
```

## version info
flux version 0.12.1
k8s version: v1.19.9

SomtochiAma:
Fixes #1283

I tested these changes and I confirm that I fixed the regression bug introduced in #1224 that prevented users from reconciling a Kustomization or helmrelease when the source is in a different namespace.

### DIFF
--- a/cmd/flux/reconcile_helmrelease.go
+++ b/cmd/flux/reconcile_helmrelease.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/types"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Thanks @SomtochiAma ☕

PR #1286, opened 2021-04-16
Suggest an alternative gradual migration technique

sa-spag:
[As suggested on Slack](https://cloud-native.slack.com/archives/CLAJ40HV3/p1618498889004400), here is a technique based on customizing `RoleBindings` bound to the Helm Operator `ServiceAccount` so it continues to continuously process legacy `HelmReleases` in all namespaces but the ones in which `HelmReleases` are being migrated to the Helm Controller's.


sa-spag:
After a second look, this may not work as expected. We are going to conduct further tests before eventually re-opening this pull request.

PR #1287, opened 2021-04-16
Add get all command

SomtochiAma:

### DIFF
--- a/cmd/flux/get_all.go
+++ b/cmd/flux/get_all.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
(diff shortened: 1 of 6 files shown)


stefanprodan:
I think we should call into get sources all and get images all, so that we have a single table for those.

cmd/flux/get_all.go
```diff
@@ -0,0 +1,91 @@
...
+				apiType: alertType,
+				list:    &alertListAdapter{&notificationv1.AlertList{}},
+			},
+			{
+				apiType: alertProviderType,
+				list:    alertProviderListAdapter{&notificationv1.ProviderList{}},
+			},
+		}
+
+		err := getSourceAllCmd.RunE(cmd, args)
```
stefanprodan, line 63: Please move this first in the list

cmd/flux/get_source_all.go
```diff
@@ -32,36 +34,31 @@ var getSourceAllCmd = &cobra.Command{
...
+			{
+				apiType: gitRepositoryType,
+				list:    &gitRepositoryListAdapter{&sourcev1.GitRepositoryList{}},
+			},
+			{
+				apiType: helmChartType,
+				list:    &helmChartListAdapter{&sourcev1.HelmChartList{}},
+			},
+			{
+				apiType: helmRepositoryType,
```
stefanprodan, line 51: Please move this before helmCharts

cmd/flux/get_all.go
```diff
@@ -0,0 +1,91 @@
...
+			{
+				apiType: receiverType,
+				list:    receiverListAdapter{&notificationv1.ReceiverList{}},
+			},
+			{
+				apiType: alertType,
+				list:    &alertListAdapter{&notificationv1.AlertList{}},
+			},
+			{
+				apiType: alertProviderType,
```
stefanprodan, line 58: Please move the provider before alerts.

stefanprodan (changes requested):

stefanprodan:
@SomtochiAma please add the page to the docs index here: https://github.com/fluxcd/flux2/blob/main/mkdocs.yml#L149 then rebase & squash.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🥇

PR #1288, opened 2021-04-16
Suggest an alternative gradual migration technique

sa-spag:
[As suggested on Slack](https://cloud-native.slack.com/archives/CLAJ40HV3/p1618498889004400), here is a technique based on customizing `RoleBindings` bound to the Helm Operator `ServiceAccount` so it continues to continuously process legacy `HelmReleases` in all namespaces but the ones in which `HelmReleases` are being migrated to the Helm Controller's.

It follows https://github.com/fluxcd/flux2/pull/1286 up, which was not completely accurate. In fact the Helm Operator requires a cluster-wide read-only role, otherwise it's not able to start its reconciliation loop at all. Credits to @ViBiOh for experimenting and validating the technique.

### DIFF
--- a/docs/guides/helm-operator-migration.md
+++ b/docs/guides/helm-operator-migration.md
@@ -5,11 +5,11 @@ This guide will learn you everything you need to know to be able to migrate from
 ## Overview of changes
 
 ### Support for Helm v2 dropped
-   
+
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Neatly written, thank you @sa-spag :sun_with_face:

PR #1290, opened 2021-04-16
Add to FAQ - Why should I upgrade

kingdonb:
The FAQ entries themselves are ready for review and could be merged as-is, but there are more changes needed.

We're pretty sure that the entire section under:

```yaml
## Flux v1 vs v2 questions

### What are the differences between v1 and v2?
```

should receive more prominent placement, at least until Flux v1 is well-and-truly behind us.

It is at the bottom of the FAQ, under `Kustomize questions` and `Helm questions`, I think that makes it harder to find this information and would propose simply moving it to the top of the FAQ.

Open to suggestions about where else this type of "Why Upgrade" content could be placed or how we can raise the prominence, (but I had to put it somewhere, and this seemed like the most consistent space to locate it.)

There is another treatment of "differences between v1 and v2" in the link under the heading that precedes this one. It has coverage of subjects like "What does Flux v2 mean for Flux" and "Why did you rewrite Flux?"

Neither of these angles really support the question "Why should I upgrade" which is what I'm trying to improve with this PR.

### DIFF
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -260,68 +260,6 @@ Flux v2 separates the functionalities into specialized controllers, collectively
 
 You can find a detailed comparison of Flux v1 and v2 features in the [migration FAQ](../guides/faq-migration.md).
 
-### Why should I upgrade
-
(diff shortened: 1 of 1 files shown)


stefanprodan (changes requested):
Please move this to migration/faq

PR #1291, opened 2021-04-18
Fix ownership issue in bash install script

jlengelsen:
When using tar with sudo the extracted files will retain the ownership
of the files in the archive. When using the bash install script the flux
binary is owned by user ID 1001 and group docker after installation.

This commit fixes the ownership by appending the -o option to the tar
command which will extract files with the correct ownership, namely user
root and group root.

Signed-off-by: Julian Lengelsen <julian.lengelsen@th-koeln.de>


hiddeco (approved):
Thanks @jlengelsen :sunflower:

Issue #1258, opened 2021-04-13
Update components version in CRD kustomization


stefanprodan:
The [update job](https://github.com/fluxcd/flux2/blob/main/.github/workflows/update.yaml) should patch https://github.com/fluxcd/flux2/blob/main/manifests/crds/kustomization.yaml

PR #1325, opened 2021-04-22
build: incl CRD version change in component update

hiddeco:
This includes updating the version in the `manifests/crds` directory
for the component thas has a newer latest version.

Fixes: #1258


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1334, opened 2021-04-23
Add upgrade CRDs policy arg to create helmrelease cmd

stefanprodan:
Changes:
- add optional `--crds` arg to `flux create helmrelease` command
- update the sealed-secrets guide with `--crds=CreateReplace`

### DIFF
--- a/cmd/flux/create_helmrelease.go
+++ b/cmd/flux/create_helmrelease.go
@@ -116,6 +116,7 @@ type helmReleaseFlags struct {
 	valuesFiles     []string
 	valuesFrom      flags.HelmReleaseValuesFrom
 	saName          string
+	crds            flags.CRDsPolicy
 }
 
 var helmReleaseArgs helmReleaseFlags
@@ -130,6 +131,7 @@ func init() {
 	createHelmReleaseCmd.Flags().StringVar(&helmReleaseArgs.saName, "service-account", "", "the name of the service account to impersonate when reconciling this HelmRelease")
 	createHelmReleaseCmd.Flags().StringArrayVar(&helmReleaseArgs.valuesFiles, "values", nil, "local path to values.yaml files")
 	createHelmReleaseCmd.Flags().Var(&helmReleaseArgs.valuesFrom, "values-from", helmReleaseArgs.valuesFrom.Description())
+	createHelmReleaseCmd.Flags().Var(&helmReleaseArgs.crds, "crds", helmReleaseArgs.crds.Description())
 	createCmd.AddCommand(createHelmReleaseCmd)
 }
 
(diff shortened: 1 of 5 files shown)


docs/guides/sealed-secrets.md
```diff
@@ -52,11 +52,12 @@ flux create helmrelease sealed-secrets \
...
 --chart=sealed-secrets \
---chart-version="1.13.x"
+--chart-version=">=1.15.0-0" \
+--crds=CreateReplace
 ```
 
-With chart version `1.13.x` we configure helm-controller to automatically upgrade the release
-when a new chart patch version is fetched by source-controller.
+With chart version `>=1.15.0-0` we configure helm-controller to automatically upgrade the release
+when a new chart version is fetched by source-controller.
```
hiddeco, line 60:
```suggestion
With chart version `>=1.15.0-0` we configure helm-controller to automatically upgrade the release
when a new chart version is fetched by source-controller.
```

I would not include the `-0` suffix with a `>=` range selector without telling the user what the suffix does.

cmd/flux/create_helmrelease.go
```diff
@@ -184,6 +186,11 @@ func createHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		helmRelease.Spec.ServiceAccountName = helmReleaseArgs.saName
 	}
 
+	if helmReleaseArgs.crds != "" {
+		helmRelease.Spec.Install = &helmv2.Install{CRDs: helmv2.CRDsPolicy(helmReleaseArgs.crds.String())}
+		helmRelease.Spec.Upgrade = &helmv2.Upgrade{CRDs: helmv2.CRDsPolicy(helmReleaseArgs.crds.String())}
+	}
```
hiddeco, line 192: I would not set the `Install` policy to the same value as upgrade, the best and safest default configuration is to simply use `Create`.

docs/guides/sealed-secrets.md
```diff
@@ -140,11 +141,15 @@ spec:
...
-      version: "1.13.x"
+        name: sealed-secrets
+      version: ">=1.15.0-0"
   interval: 1h0m0s
   releaseName: sealed-secrets
   targetNamespace: flux-system
+  install:
+    crds: CreateReplace
+  upgrade:
+    crds: CreateReplace
```
hiddeco, line 152:
```suggestion
  install:
    crds: Create
  upgrade:
    crds: CreateReplace
```

docs/guides/sealed-secrets.md
stefanprodan: Without `-0` there is no way to install it, they only do pre releases.

hiddeco: Oh, the joy of versioning these days.

cmd/flux/create_helmrelease.go
```diff
@@ -130,6 +131,7 @@ func init() {
 	createHelmReleaseCmd.Flags().StringVar(&helmReleaseArgs.saName, "service-account", "", "the name of the service account to impersonate when reconciling this HelmRelease")
 	createHelmReleaseCmd.Flags().StringArrayVar(&helmReleaseArgs.valuesFiles, "values", nil, "local path to values.yaml files")
 	createHelmReleaseCmd.Flags().Var(&helmReleaseArgs.valuesFrom, "values-from", helmReleaseArgs.valuesFrom.Description())
+	createHelmReleaseCmd.Flags().Var(&helmReleaseArgs.crds, "crds", helmReleaseArgs.crds.Description())
```
hiddeco, line 134:
```suggestion
	createHelmReleaseCmd.Flags().Var(&helmReleaseArgs.crds, "upgrade-crds", helmReleaseArgs.crds.Description())
```

maybe?

stefanprodan: I would leave it like it is, since we also set the install value.

hiddeco (approved):

Issue #1346, opened 2021-04-25
Throttling cause of client-go


rverma-nsl:
### Describe the bug
I am seeing a lot errors from flux below:
```
1 request.go:621] **Throttling request took** 1.047035991s, request: GET:https://10.0.0.1:443/apis/**?timeout=32s
```

### Additional context

- Kubernetes version: eks 1.19
- Git provider: github
- Container registry provider: ecr

While researching found that these messages are generated from client-go in the first place. Relevant issues
- https://github.com/vmware-tanzu/velero/issues/3191
- https://github.com/kubernetes/kubernetes/issues/86149

Can we also introduce the clientqps in flux cli?

hiddeco:
All our controllers should support the following flags to configure client QPS:

```
--kube-api-burst int            The maximum burst queries-per-second of requests sent to the Kubernetes API. (default 50)
--kube-api-qps float32          The maximum queries-per-second of requests sent to the Kubernetes API. (default 20)
```

stefanprodan:
This issue is about the CLI I think.

PR #1349, opened 2021-04-26
Avoid throttling when some Flux CRDs are not registered

stefanprodan:
When running `flux get all` users that haven't installed the image automation will see:
```
I0426 13:38:52.372570    1374 request.go:655] Throttling request took 1.000968421s, request: GET:https://35.228.232.234/apis/helm.toolkit.fluxcd.io/v2beta1?timeout=32s
✗ no matches for kind "ImageRepository" in version "image.toolkit.fluxcd.io/v1alpha2"
```

This PR increases the client-side rate limits to avoid the above. fix: #1346

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -156,6 +156,10 @@ func KubeConfig(kubeConfigPath string, kubeContext string) (*rest.Config, error)
 		return nil, fmt.Errorf("kubernetes configuration load failed: %w", err)
 	}
 
+	// avoid throttling request when some Flux CRDs are not registered
+	cfg.QPS = 50
+	cfg.Burst = 100
+
 	return cfg, nil
 }
 


internal/utils/utils.go
```diff
@@ -156,6 +156,10 @@ func KubeConfig(kubeConfigPath string, kubeContext string) (*rest.Config, error)
 		return nil, fmt.Errorf("kubernetes configuration load failed: %w", err)
 	}
 
+	// avoid throttling request when some Flux CRDs are not registered
+	cfg.QPS = 50
+	cfg.Burst = 100
```
hiddeco, line 161: Maybe better to add these as arguments to function so that making a `rootArgs` flags based configuration option will be easier in a next iteration?

stefanprodan: Why would you want to change them?

hiddeco: Because there does not seem to be a consensus on the right values to use, with config values ranges from `100/100` to `1e6/1e6` in various popular projects (prometheus-operator, nginx-controller, etc).

Providing the option to configure the value means that if for some reason K8s (or a specific distro) start to behave differently, users can apply a workaround themselves without having to wait on a fix from our side.

hiddeco (approved):
Still of opinion that flag control would be a welcome addition, but not in a way that makes this blocking.

PR #1350, opened 2021-04-26
Add bootstrap option to set a prefix for the commit messages

DriesVerachtert:
Flux bootstrap already has an option to set a suffix for the commit messages. This change also adds a prefix. It's handy in environments where commit messages have to start with some ticket number for example.

I'm not a go developer: I basically copied the code of the 'commitMessageAppendix' option.

Signed-off-by: Dries Verachtert <dries.verachtert@dries.eu>


hiddeco (commented):
Thank you for submitting a PR :cake: 

I think that if there is a need to be able to control both the suffix and the appendix of the Git commit message, it would probably be better to deprecate the `-appendix` flag in favor of one that allows you to configure a full Git commit message template. Given this would prevent a wild-growth of commit message related flags that all modify different parts of the message.

olivercp3:
hello, so how can i set a prefix for the commit messages now

dhoppe:
@hiddeco I would appreciate a solution to this problem. My GitLab CI pipeline fails every time there is a commit message like "Update Flux manifests".

matheuscscp:
With Flux Operator bootstrap does not require push access to Git:

https://fluxcd.control-plane.io/operator/

PR #1354, opened 2021-04-26
Use Case Docs: Image automation Auto-PR with GitHub Actions

kingdonb:
The linked github actions workflow (from the original note in the Image Update guide) does not actually do what you need, I believe I have found a better one and effectively demonstrated how to use it in this example.

The "GitHub Actions Auto PR" use case could be expanded upon, but I think maybe this short version already says everything it needs to say!

### DIFF
--- a/docs/use-cases/gh-actions-auto-pr.md
+++ b/docs/use-cases/gh-actions-auto-pr.md
@@ -2,7 +2,7 @@
 
 In the [Image Update Guide] we saw we can [Push updates to a different branch] by using `.spec.git.push.branch` to push image updates to a different branch than the one used for checkout.
 
-In this example, we configure an `ImageUpdateAutomation` resource to push to a `staging` branch, which we could set up separately as a preview environment to deploy automatic updates in a staging cluster or namespace.
+In this example, we configure an `ImageUpdateAutomation` resource to push to a `staging` branch, (which we could set up separately as a preview environment to deploy automatic updates in a staging cluster or namespace.)
 
 ```yaml
 kind: ImageUpdateAutomation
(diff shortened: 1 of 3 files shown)


docs/use-cases/gh-actions-auto-pr.md
```diff
@@ -0,0 +1,58 @@
...
+        pr_title: "Pulling ${{ github.ref }} into main"
+        pr_body: ":crown: *An automated PR*"
+        pr_reviewer: "kingdonb"
+        pr_draft: true
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+```
+
+You can use the [github-pull-request-action] workflow to automatically open a pull request against a destination branch. In this case, when `staging` is merged into the `main` branch, changes are deployed in production.
+
+This way you can automatically push changes to a `staging` branch and require manual approval of any automatic image updates before they are applied on your production clusters.
```
kingdonb, line 54: The ending this way is anti-climactic, I think it should go on a bit longer than this.

Explain that if more image updates come along before this pull request is approved, since they will be appended by Flux's automation to the pull request, they can be approved as a batch to move them all to Production together simultaneously. Demonstrate how this workflow has the behavior of rolling out changes as a coordinated release.

kingdonb:
@dholbach Stefan suggested we incorporate this use case example into the Flux blog post for this month; it probably needs some minor changes and I'm happy to help figure out how to best incorporate it to the blog post as well, if you need help with that.

Please take a look! I will also need someone to review this for merge. TY

dholbach:
Thanks for the heads-up - I'll make sure it's mentioned.

hiddeco:
To prevent us from having to move things over tomorrow, it is probably best if this PR is filled against the `fluxcd/website` repository.

kingdonb:
Closing in favor of fluxcd/website#160

PR #1359, opened 2021-04-27
Copy docs assets for Netlify build

dholbach:


hiddeco (approved):
Confirmed both https://deploy-preview-1359--fluxcd-toolkit.netlify.app/components/source/helmrepositories/ and https://deploy-preview-1359--fluxcd-toolkit.netlify.app/install.sh to be working.

Thanks @dholbach :beers:

PR #1360, opened 2021-04-27
add redirects file

dholbach:
Add redirects from toolkit.fluxcd.io to fluxcd.io/docs - largely following https://github.com/fluxcd/flux2/discussions/367

One thing which was unfortunately overseen to date is:

User Guides | toolkit.fluxcd.io/guides | fluxcd.io/docs/flux
-- | -- | --

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,0 +1,17 @@
+# individual rules
+/core-concepts          https://fluxcd.io/docs/concepts                     301!
+/contributing           https://fluxcd.io/contributing                      301!
+
+# refer to https://github.com/fluxcd/flux2/discussions/367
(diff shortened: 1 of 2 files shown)


_redirects
```diff
@@ -0,0 +1,14 @@
+# individual rules
+/core-concepts          https://fluxcd.io/docs/concepts                     301!
+/contributing           https://fluxcd.io/contributing                      301!
+/guides
```
hiddeco, line 4: Think this line should be removed?

dholbach: Yes, an oversight.

hiddeco (approved):
:100:

Issue #1135, opened 2021-03-22
Move toolkit.fluxcd.io to netlify


dholbach:
In preparation for moving the docs into `f/website`, @stefanprodan suggested we move the docs to Netlify, so we can fix up redirects between the sites within Netlify more easily.

dholbach:
The Netlify app needs access to `f/flux2` - I just sent a request.

hiddeco:
Your request has been approved.

dholbach:
- [ ] change the netlify dns settings
- [ ] stop the deploy to gh-pages
- [ ] update GH settings accordingly

PR #1361, opened 2021-04-27
Stop deploy to Github pages

dholbach:
Fixes: #1135


hiddeco:
#1360 needs to be merged first.

stefanprodan:
Why not delete the whole workflow? it does nothing as it is.

hiddeco (approved):

PR #1363, opened 2021-04-27
Explain where the docs went

dholbach:
... also simplify netlify config.


hiddeco:
Also: I can not manually run Netlify builds against fork branches.

dholbach:
> We forgot a redirect rule: `https://toolkit.fluxcd.io/install.sh` -> `https://fluxcd.io/docs/install.sh`.

Good catch! Added!

hiddeco (approved):

Issue #1388, opened 2021-05-05
Remove 'command' from Flux CLI sidebar


stefanprodan:
We need to remove the `command` suffix for all the items in the flux CLI sidebar.

PR #1389, opened 2021-05-06
Remove ' command' from Flux CLI docs title

dholbach:
Fixes: #1388


stefanprodan (approved):

dholbach:
This will only be reflected on the website after the next `flux2` release: https://github.com/fluxcd/website/blob/main/hack/import-flux2-assets.sh#L152

PR #1390, opened 2021-05-06
Switch StringArrayVar flags to use StringSliceVar

Callisto13:
**Description**

StringSliceVar allows for more flexibility when passing vars to list flags.
Both formats will be supported:
- '--foo=one --foo=two'
- '--foo=one,two'

**Context**
For `eksctl`'s integration with Flux 2, we want to find a way to support as much (if not all) the `bootstrap` commands with as little interference as possible. Flux is developing quickly and we'd like users to get on with bootstrapping their new `eksctl` clusters without needing to come to us with requests to expose this-and-that feature or flag.
Our intended solution is to introduce a `map[string]string` into `eksctl`'s config so that users can provide an arbitrary list of flags for us to pass through to `bootstrap`.

`StringArrayVar` flags would make this plan impossible since you would end up with duplicate keys in the yaml:

```yaml
flags:
  team: one
  team: two
```

With `StringSliceVar`, users have the same experience they currently do, with the option to also provide a list of comma-separated strings , and therefore a yaml of:

```yaml
flags:
  team: one,two
```

An alternative for `eksctl` is to use `[]string`, but this is less tidy, and we hope this change is low-key enough to not cause issue.

(We only need this for `bootstrap` commands, but I have made all other commands which have `StringArrayVar` flags the same for completeness' sake.)

Thanks!


stefanprodan (approved):
LGTM

Thanks @Callisto13 🏅

Issue #1092, opened 2021-03-14
Grafana Dashboards Using k8s <1.16 Metrics Labels


robert-ovens:
I've had problems using the GitOps Control Plane dashboard in our kube-prometheus-stack - the gauges or panels do not display data. Hunting around I found a [stackoverflow question](https://stackoverflow.com/questions/59625516/grafana-dashboard-not-displaying-pod-name-instead-pod-name) that relates to the same issue.

I'm running k8s v1.20 and the metrics used in the dashboard were removed with k8s v1.16.

https://github.com/fluxcd/flux2/blob/022576697f723667e627a0e07c1a0aec25d1b432/manifests/monitoring/grafana/dashboards/control-plane.json#L70

Replacing `kubernetes_pod_name` with `pod` and `kubernetes_namespace_name` with `namespace` fixed the dashboard.

The new metric names were [added to k8s 1.14](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md#action-required-2). @stefanprodan, are you with dropping GitOps Control Plane dashboard support for K8s < 1.14? If you are, I'll submit a pull request.

Robert.

stefanprodan:
Changing those labels would break the dashboard, see https://github.com/fluxcd/flux2/blob/main/manifests/monitoring/prometheus/prometheus.yml#L48 

We could normalise our Prom config with kube-prometheus-stack, but will this break the dashboards for other users?

pierluigilenoci:
@stefanprodan K8s v1.16 is EOL since 2020-09-02, it is time to move on. ;) 
Ref: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history

xtalkme:
Also ran into this issue yesterday, it would be great to have recent labels.

SomtochiAma:
I will pick up this issue

PR #1391, opened 2021-05-06
Update Prometheus labels and dashboard

SomtochiAma:
Fixes: #1092

### DIFF
--- a/manifests/monitoring/grafana/deployment.yaml
+++ b/manifests/monitoring/grafana/deployment.yaml
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
         - name: grafana
-          image: "grafana/grafana:7.2.1"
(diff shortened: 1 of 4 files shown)


manifests/monitoring/grafana/deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
         - name: grafana
-          image: "grafana/grafana:7.2.1"
+          image: "grafana/grafana:7.5.5"
```
stefanprodan, line 21: This is not OK, we can't use Grafana under AGPL, the last Grafana version released under Apache is v7.5.4.

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🌷

Issue #1387, opened 2021-05-05
default branch inconsistency


rtoma:
### Describe the bug

While migrating to v2 I noticed the following.

The default branch for `flux bootstrap` is "main".
But, the default branch for `flux create source git` is "master".
This inconsistency is trivial but can be confusing. Also, "master" is not really expected anymore.

### To Reproduce

Steps to reproduce the behaviour:

```
$ flux bootstrap -h | grep branch
      --branch string                          Git branch (default "main")
```

vs

```
$ flux create source git -h | grep branch
...
      --branch string                          git branch (default "master")
```

### Expected behavior

Default branch to be consistent and as expected: "main".

### Additional context

- Kubernetes version:
- Git provider:
- Container registry provider:

Below please provide the output of the following commands:

```
$ flux --version
flux version 0.13.2
```

PR #1393, opened 2021-05-06
Make the Git ref required

stefanprodan:
Remove the default branch value from `flux create source git` and validate that one of the ref options are specified.

Fix: #1387


relu (approved):
LGTM :heavy_check_mark:

PR #1469, opened 2021-06-01
Fix and Refactor integrations

stealthybox:
Some of these bases were pretty broken.
I've added a Makefile to vet all of the kustomize builds, so it's easier to verify that the config is stacking as intended and without error.

None of these should be breaking changes:

1. Add Makefile to test integrations
1. Add /kbin/kubectl to _base integrations
1. Remove per-cloud /kbin/kubectl patches
1. Fix GCP integration container image
1. Template AzureIdentityBinding using $(AZ_IDENTITY_NAME) for integrations
1. Remove unused kustomizeconfigs from integrations
1. Fix eventhub integration config patches

It's worth noting that this Makefile is just for the integrations folder.
It could be useful for the whole manifests directory.
The only base that doesn't build on its own is `monitoring/monitoring-config` because it needs `LoadRestrictionsNone`.


stefanprodan (approved):
LGTM

Thanks @stealthybox for all the fixes 🥇

PR #1470, opened 2021-06-02
add gitImplementation as a bootstrap flag

NissesSenap:
The gitImplementation flag is needed as a bootstrap flag to be able to
clone using libgit2 for example when bootstraping in a azure devops repo.

Signed-off-by: Edvin Norling <edvin.norling@xenit.se>

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -41,6 +41,7 @@ type bootstrapFlags struct {
 
 	branch            string
 	recurseSubmodules bool
+	gitImplementation flags.GitImplementation
 	manifestsPath     string
 
 	defaultComponents  []string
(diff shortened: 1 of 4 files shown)


stefanprodan:
This is not possible as libgit2 requires C binding making the CLI unusable on any OS!

hiddeco:
The problem is that while this makes it configurable (for the controllers), the `flux` binary itself also relies on `go-git`. So to solve the issue for e.g. Azure DevOps, we need another Git implementation (probably using `git`) for such repositories.

hiddeco:
To give some more pointers, there is an interface here (that would probably need some adjustments as it was just brought in place to create _some_ abstraction): https://github.com/fluxcd/flux2/blob/main/internal/bootstrap/git/git.go

Another implementation on top of `git` would then probably look a bit like what we do in Flux v1: https://github.com/fluxcd/flux/blob/master/pkg/git/working.go, while providing some utilities to create a boxed working environment with some environmental configurations applied to provide e.g. access to the SSH agent.

NissesSenap:
Thanks for all the feedback and sorry for my slow response.
For now I will close this PR and instead follow the docs for azure: https://fluxcd.io/docs/use-cases/azure/#flux-installation-for-azure-devops

Issue #1524, opened 2021-06-15
ACR credentials sync: Can't create secret due to the default role setup


molszews:
### Describe the bug

https://github.com/fluxcd/flux2/blob/main/manifests/integrations/registry-credentials-sync/_base/sync.yaml#L104

This should not be enabled by default as:
```
Note: You cannot restrict create or deletecollection requests by resourceName. For create, this limitation is because the object name is not known at authorization time.
```
(https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)

The effect is k8s can't create such secret due to insufficient privileges (it took me some time to figure out the real reason)

`error: failed to create secret secrets is forbidden: User "system:serviceaccount:flux-system:default" cannot create resource "secrets" in API group "" in the namespace "flux-system"`

NissesSenap:
I will take a stab at a fix for it

PR #1528, opened 2021-06-16
Remove resourceNames in integration secrets

NissesSenap:
* Fixes: #1524
* We remove resourceName due to the following:
  Note: You cannot restrict create or deletecollection requests by resourceName.
  For create, this limitation is because the object name is not known at authorization time.
* Fix typo in azure-registry cronjob
Signed-off-by: Edvin Norling <edvin.norling@xenit.se>

### DIFF
--- a/manifests/integrations/eventhub-credentials-sync/_base/sync.yaml
+++ b/manifests/integrations/eventhub-credentials-sync/_base/sync.yaml
@@ -110,8 +110,8 @@ rules:
       - update
       - patch
     # Lock this down to the specific Secret name  (Optional)
-    resourceNames:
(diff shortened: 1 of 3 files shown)


NissesSenap:
ping @molszews

NissesSenap:
This is only a issue in the deployment solution, if you use a cronjob it will be checked when the job is created.
I can't come up with a good solution to be able to use a specific secret in the deployment solution and still have it as a variabel.

We could of course use a hard corded variabel name, and then we shouldn't get this issue if I understand the error correctly.

stefanprodan:
@NissesSenap please rebase instead of merge.

NissesSenap:
Sure I will remember that for the future. Thanks @stefanprodan

stefanprodan (approved):
LGTM 

Thanks @NissesSenap ☕

PR #1544, opened 2021-06-18
Add create target namespace arg to helmrelease cmd

stefanprodan:
Ref: #213


hiddeco (approved):
Thank you @stefanprodan :1st_place_medal:

PR #1575, opened 2021-06-30
fix Control Plane dashboard legend

dminca:
The legend was not showing the Pod name, instead the whole resource in
the dashboard

As a result, use the correct Prometheus label

For example, this was displayed previously
image

With this PR, it'll look like this
image

Resolves:
Related:


stefanprodan (approved):
LGTM

Thanks @dminca 🏅

PR #1590, opened 2021-07-06
Add no-header flag for get commands to omit printing the header

### LINKED ISSUE #1585: Add --no-headers flag to the flux get commands
### Discussed in https://github.com/fluxcd/flux2/discussions/1562

<div type='discussions-op-text'>

<sup>Originally posted by **jack-evans** June 25, 2021</sup>
this is similar to what kubectl command line exposes. Currently in flux, you get the following:
```
$ flux get source git 
NAME            READY	MESSAGE                                                              	REVISION                                           	SUSPENDED
my-repo	        True 	Fetched revision: my-branch/caa83ff9b56c8c43273c72b4fec4a18552061c5f	my-branch/caa83ff9b56c8c43273c72b4fec4a18552061c5f	False
flux-system   	True 	Fetched revision: main/e65c5154a42ee495a05d621ba7afa015a0daf616     	main/e65c5154a42ee495a05d621ba7afa015a0daf616     	False
```
what I am proposing is that adding a `--no-headers` option flag to the commands under get, producing the following:
```
$ flux get source git --no-headers
my-repo	        True 	Fetched revision: my-branch/caa83ff9b56c8c43273c72b4fec4a18552061c5f	my-branch/caa83ff9b56c8c43273c72b4fec4a18552061c5f	False
flux-system   	True 	Fetched revision: main/e65c5154a42ee495a05d621ba7afa015a0daf616     	main/e65c5154a42ee495a05d621ba7afa015a0daf616     	False
```

This would then allow a user to use scripting to automate processes, an example being:
```bash
sources=$(flux get source git --no-headers | awk '{print $1}')

while IFS=$'\n' read -r my_source; do
  flux reconcile source git "${my_source}"
done <<< "${sources}"
```

If we are looking for something more generic we could use something like --output and allow the end user to go templating? </div>

alex-petrov-vt:
This PR adds a  --no-header flag for all of the get commands. When the flag is provided, the table header is not printed. Closes #1585

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -40,6 +40,7 @@ var getCmd = &cobra.Command{
 
 type GetFlags struct {
 	allNamespaces bool
+	noHeader      bool
 }
 
 var getArgs GetFlags
@@ -47,6 +48,7 @@ var getArgs GetFlags
 func init() {
 	getCmd.PersistentFlags().BoolVarP(&getArgs.allNamespaces, "all-namespaces", "A", false,
 		"list the requested object(s) across all namespaces")
+	getCmd.PersistentFlags().BoolVarP(&getArgs.noHeader, "no-header", "", false, "skip the header when printing the results")
 	rootCmd.AddCommand(getCmd)
 }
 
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @alex-petrov-vt 🏅

PR #1594, opened 2021-07-06
Add status-selector flag for get commands to filter results based on status conditions

### LINKED ISSUE #1587: Add --status-selector to flux get commands
We could add a field called `status-selector` to Flux get commands so that users can specify the status condition name and the desired state to filter the result.

The filter format could be `flux get <kind> --status-selector=<condition-name>=<true|false|unknown>` .

Print sources that are not ready:

```
flux get source all --status-selector=ready=false
``` 

Print Kustomizations with failing health checks:

```
flux get kustomizations --status-selector=healthy=false
``` 

### Discussed in https://github.com/fluxcd/flux2/discussions/1578

<div type='discussions-op-text'>

<sup>Originally posted by **eloo** July  1, 2021</sup>
Hi,

i would like to suggest a filter function for the flux cli. 
In my current situation i would like to have an option to filter for all flux resources in a specific state.. for example ready==false

</div>

chanwit:
Signed-off-by: Chanwit Kaewkasi <chanwit@gmail.com>

This PR implements status-selector for the `get` commands.

Fixes #1587

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -39,8 +39,9 @@ var getCmd = &cobra.Command{
 }
 
 type GetFlags struct {
-	allNamespaces bool
-	noHeader      bool
+	allNamespaces  bool
+	noHeader       bool
+	statusSelector string
 }
 
 var getArgs GetFlags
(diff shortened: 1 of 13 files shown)


cmd/flux/get.go
```diff
@@ -49,13 +50,16 @@ func init() {
 	getCmd.PersistentFlags().BoolVarP(&getArgs.allNamespaces, "all-namespaces", "A", false,
 		"list the requested object(s) across all namespaces")
 	getCmd.PersistentFlags().BoolVarP(&getArgs.noHeader, "no-header", "", false, "skip the header when printing the results")
+	getCmd.PersistentFlags().StringVar(&getArgs.statusSelector, "status-selector", "",
+		"specify the status condition name and the desired state to filter the get result")
```
stefanprodan, line 54:
```suggestion
		"specify the status condition name and the desired state to filter the get result, e.g. ready=false")
```

stefanprodan (approved):
LGTM

Thanks @chanwit 🏅

PR #1609, opened 2021-07-09
Request reconcile using patch instead of update

hiddeco:
This should prevent the generation of the object getting bumped, as
observed on a GKE K8s 1.18 cluster using the  logic before this commit.

We only want to generation to increase when there are actual changes to
the `spec` of a resource, as some controllers use the `generation`
value to make assumptions about what they should do during a
reconciliation.

Ref: https://github.com/fluxcd/helm-controller/issues/295


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1642, opened 2021-07-22
Update website via CONTRIBUTING.md

kingdonb:
We are directing folks to #flux-dev which is a channel for bots, and doesn't really host conversations.

For contributors, I think we want these folks to find us in #flux-contributors (rather than visiting #flux-dev)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,8 @@ you can sign your commit automatically with `git commit -s`.
 
 For realtime communications we use Slack: To join the conversation, simply
 join the [CNCF](https://slack.cncf.io/) Slack workspace and use the
-[#flux-dev](https://cloud-native.slack.com/messages/flux-dev/) channel.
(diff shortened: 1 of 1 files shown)


kingdonb:
(This gets pulled into the website via: https://github.com/fluxcd/website/tree/main/external-sources)

kingdonb:
There are a number of issues on this page: https://fluxcd.io/docs/contributing/flux/#communications

I am having trouble finding the actual source of this doc, meanwhile please hold off from merging this. (I'd like to resolve all the bad links/incorrect markdown link syntax issues that I found before I lose track of all this.)

hiddeco:
> At a guess, what I think is happening is that `hack/gen-content.py` has started failing silently, and `external-sources` aren't actually being updated when website merges anymore.

This should not be possible, the website is rebuild every time a merge happens, and if it would fail, there would not be a page anymore.

PR #1653, opened 2021-07-27
Provide suggestion for some fields in bug report

hiddeco:
This commit adds suggestion for some of the required fields because
there can be a few exceptions in which these values can not be
provided. For example when the `flux` binary can not be installed, or
the set of controllers has been installed with just an `install.yaml`
file.

We use this approach instead of making the fields optional because it
guides people to provide the information whenever they can, which
should result in higher quality bug reports.

As a tiny addition, the reference to "Flux v2" has been renamed to just
"Flux" as we are slowly transitioning "Flux v1" into "Flux Legacy".


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1626, opened 2021-07-16
Fix trace for optional GitRepository.Spec.Reference

allenporter:
Check for existence of GitRepository.Spec.Reference when displaying a trace to
avoid error:

✗ template: tmpl:28:21: executing "tmpl" at <.GitRepository.Spec.Reference.Tag>: nil pointer evaluating *v1beta1.GitRepositoryRef.Tag

Fixes issue #1621
Manually tested using the use cases highlighted in the issue.

Signed-off-by: Allen Porter <allen@thebends.org>

Issue #1603, opened 2021-07-08
Rewrite e2e conformance tests using Golang


chanwit:
We should rewrite e2e conformance tests using Golang, for example:

From this,
```
      - name: flux check --pre
        run: |
          /tmp/flux check --pre
```

should be rewritten into this,
```
func TestFluxCheckPre(t *testing.T) {
	var lines []string
	var err error

	lines, err = run("kubectl", "version", "--short")
	assert.NoError(t, err)

	clientVersion := strings.SplitN(lines[0], ": v", 2)[1]
	serverVersion := strings.SplitN(lines[1], ": v", 2)[1]

	lines, err = run(fluxBinary, "check", "--pre")
	assert.NoError(t, err)

	assert.Equal(t, "► checking prerequisites", lines[0])
	assert.Equal(t, fmt.Sprintf("✔ kubectl %s >=1.18.0-0", clientVersion), lines[1])
	assert.Equal(t, fmt.Sprintf("✔ Kubernetes %s >=1.16.0-0", serverVersion), lines[2])
	assert.Equal(t, "✔ prerequisites checks passed", lines[3])
}
```

I'll go with this chuck-by-chuck as we have a load of tests to port.
A chuck is a PR containing 3-4 tests, for example. So it's going to make us easier to review.

- [ ] `flux check --pre`
- [ ] `flux install --manifests`
- [ ] `flux create secret`
- [ ] `flux create source git`
- [ ] `flux create source git export apply`
- [ ] `flux get sources git`
- [ ] `flux get sources git --all-namespaces`
- [ ] `flux create kustomization`
- [ ] `flux trace`
- [ ] `flux reconcile kustomization --with-source`
- [ ] `flux get kustomizations`
- [ ] `flux get kustomizations --all-namespaces`
- [ ] `flux suspend kustomization`
- [ ] `flux resume kustomization`
- [ ] `flux export`
- [ ] `flux delete kustomization`
- [ ] `flux create source helm`
- [ ] `flux create helmrelease --source=HelmRepository/podinfo`
- [ ] `flux create helmrelease --source=GitRepository/podinfo`
- [ ] `flux reconcile helmrelease --with-source`
- [ ] `flux get helmreleases`
- [ ] `flux get helmreleases --all-namespaces`
- [ ] `flux export helmrelease`
- [ ] `flux delete helmrelease podinfo-helm`
- [ ] `flux delete helmrelease podinfo-git`
- [ ] `flux delete source helm`
[...]

PR #1671, opened 2021-08-01
Add tests for flux trace command

allenporter:
Add tests for flux trace command that fake out the kubernetes client,
load objects from a yaml file and create them in the client, and
assert on the output of the trace command to an expected golden file.

Follow up from the suggestions in PR https://github.com/fluxcd/flux2/pull/1626 which suggested that additional
testing would be helpful. This test approach is modeled after the helm command tests.

This required some changes to the kubernetes client setup to make it possible to use a fake.  If we agree this pattern makes sense, it can be applied to other commands (kube client creation, and test library setup could be moved to a common place). An alternative could be to use `envtest` which seemed much more heavyweight. Another alternative could be to declare all the objects in go code instead of in yaml.

To set your expectations: I'm new to go, flux, and k8s libraries so am not familiar with many existing conventions, but very happy to adjust as needed.

Ref: #1603

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"github.com/fluxcd/flux2/internal/utils"
 	"github.com/fluxcd/flux2/pkg/manifestgen/install"
 )
 
@@ -124,6 +125,18 @@ func NewRootFlags() rootFlags {
 	}
 	rf.defaults.Version = "v" + VERSION
 	return rf
+}
+
+type rootContext struct {
+	kubeManager utils.KubeManager
+}
+
+var rootCtx = NewRootContext()
+
+func NewRootContext() rootContext {
+	var rc rootContext
+	rc.kubeManager = utils.DefaultKubeManager()
+	return rc
 }
 
 func main() {
--- a/cmd/flux/testdata/trace/deployment.txt
+++ b/cmd/flux/testdata/trace/deployment.txt
@@ -1,0 +1,26 @@
+
+Object:         deployment/podinfo
+Namespace:      podinfo
+Status:         Managed by Flux
+---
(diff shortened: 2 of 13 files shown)


stefanprodan (commented):
This looks great! @allenporter can you please move the test files inside `testdata/trace` please, this way we'll have a dir per command. Thanks

stefanprodan:
@allenporter can you rebase with main, squash both commits and signoff your commit please.

allenporter:
I see you also recommended following this pattern for the e2e tests too which is cool.

As I am thinking about adding coverage of more commands, wanted to share my early thoughts. In particular, just want to make sure this is easy to maintain:
- Will it be easy to detect bugs in yaml object files? Seems worth it for now, but not sure how this plays out
- Will maintaining the yaml files require lots of small changes across many files as new fields are added?
- Should simple test cases just inline the errors instead of using a golden output? (e.g. "object kind is required (--kind)", "object apiVersion is required (--api-version", etc).  I lean towards yes.

stefanprodan:
> Will it be easy to detect bugs in yaml object files?

We could use kubeval along with the Flux API specs to validate the inputs, e.g. https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/scripts/validate.sh

> Will maintaining the yaml files require lots of small changes across many files as new fields are added?

It depends on the test case, Flux APIs are stable, that means all new fields are either optional or have defaults. When adding a new field to a CRD, if that field is also added to the CLI `create` command, then yes, the golden file should be updated.

> Should simple test cases just inline the errors instead of using a golden output? (e.g. "object kind is required (--kind)", "object apiVersion is required (--api-version", etc). I lean towards yes.

Yes I agree, no point in creating in tones of one-line golden files. I think golden files should be used for when the output is a table view or a list view.

stefanprodan (approved):
LGTM

Thanks @allenporter 🥇

stefanprodan:
@allenporter FYI https://github.com/fluxcd/flux2/pull/1682

Issue #1586, opened 2021-07-05
Add --watch option to flux get commands


stefanprodan:
Flux get commands could have a watch option similar to `kubectl get --watch`. When `--watch` is set, the CLI would fist list the objects as a table, then watch for changes in-cluster and append a table row every time an abject changes. Note that `--watch` would work only for single resource type. e.g.

```
$ kubectl get all -w
error: you may only specify a single resource type
```

PR #1672, opened 2021-08-02
Adds a watch flag to the get command

souleb:
Fix: #1586 

The new flag fetch and display the request ressource and then continue
watching the ressource until timeout or cancellation.

A single ressource/ressource type is supported.

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -25,12 +25,35 @@ import (
 	"github.com/spf13/cobra"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	watchtools "k8s.io/client-go/tools/watch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluxcd/pkg/apis/meta"
 
 	"github.com/fluxcd/flux2/internal/utils"
 )
+
+type deriveType func(runtime.Object) (summarisable, error)
+
+type typeMap map[string]deriveType
+
+func (m typeMap) registerCommand(t string, f deriveType) error {
+	if _, ok := m[t]; ok {
+		return fmt.Errorf("duplicate type function %s", t)
+	}
+	m[t] = f
+	return nil
+}
+
+func (m typeMap) execute(t string, obj runtime.Object) (summarisable, error) {
+	f, ok := m[t]
+	if !ok {
+		return nil, fmt.Errorf("unsupported type %s", t)
+	}
+	return f(obj)
+}
 
 var getCmd = &cobra.Command{
 	Use:   "get",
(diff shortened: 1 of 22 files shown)


souleb:
@stefanprodan this is still in progress, but would like your opinion. The package [cli-runtime](https://pkg.go.dev/k8s.io/cli-runtime) offers a set of helpers that would give an experience similar to kubectl.

kingdonb:
I gave this a try, it works for Kustomizations so far, similar to kubectl the `all` resources alias is not supported with watch:

```
flux2 (flux-watch * u=)$ ./bin/flux get all -w
✗ expected a single resource type, but found all
yebyen@moo:~/go/src/github.com/fluxcd/flux2 (flux-watch * u=)$ ./bin/flux get ks -w
NAME       	READY	MESSAGE                                                        	REVISION                                     	SUSPENDED
flux-system	True 	Applied revision: main/839679c85b6abdee94628d869349110a2a7e282f	main/839679c85b6abdee94628d869349110a2a7e282f	False
cert-manager	True	Applied revision: main/839679c85b6abdee94628d869349110a2a7e282f	main/839679c85b6abdee94628d869349110a2a7e282f	False
my-secrets	True	Applied revision: main/839679c85b6abdee94628d869349110a2a7e282f	main/839679c85b6abdee94628d869349110a2a7e282f	False
flux-system	Unknown	reconciliation in progress	main/839679c85b6abdee94628d869349110a2a7e282f	False
flux-system	True	Applied revision: main/839679c85b6abdee94628d869349110a2a7e282f	main/839679c85b6abdee94628d869349110a2a7e282f	False
```

I assume this is meant to work eventually, part of the WIP that you mentioned:

```
$ ./bin/flux get source git -w
✗ This object type is not supported
```

This is good progress!

cmd/flux/get.go
```diff
@@ -141,28 +155,116 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
...
+		objToPrint := e.Object
+		var sink summarisable
+		switch k := get.apiType.kind; k {
+		case "Kustomization":
+			o := objToPrint.(*kustomizev1.Kustomization)
+			sink = kustomizationListAdapter{&kustomizev1.KustomizationList{
+				Items: []kustomizev1.Kustomization{
+					*o,
+				}}}
+		case "Alerts":
```
stefanprodan, line 223: Please don't hardcode the kinds, the kind names are already declared in the CLI e.g. `alertType.kind` or in each API package e.g. `notificationv1.AlertKind`

souleb: yes I'll do the necessary cleaning .

stefanprodan:
@souleb please rebase with main, solve the conflicts and force push.

stefanprodan:
@souleb please run `make test` before pushing changes. You have to adapt the client used for testing to match your changes.

souleb:
> @souleb please run `make test` before pushing changes. You have to adapt the client used for testing to match your changes.

sorry about that. Fixed

stefanprodan (commented):
LGTM

Thanks @souleb 🏅

stefanprodan (approved):

Issue #1658, opened 2021-07-29
Remove deprecated `io/ioutil` from projects


hiddeco:
All of our projects make use of Go 1.16, and `io/ioutil` has been deprecated since this version: https://golang.org/doc/go1.16#ioutil

We should remove all traces in our projects and replace them with the suggestions from the deprecation notice.

https://github.com/search?q=ioutil+org%3Afluxcd&type=code

krishnaindani:
Are there still changes to be made with this issue, could I work on this?

JosephWoodward:
Judging by the merged pull requests above, this issue is no longer relevant and can be closed?

allenporter:
There are still many to cleanup: https://github.com/search?q=ioutil+org%3Afluxcd&type=code

dholbach:
- [ ] source-watcher + website: (https://github.com/fluxcd/source-watcher/blob/bebe0f5f4265f9ae7fce90770322b571bb28f324/controllers/gitrepository_watcher.go and https://github.com/fluxcd/website/blob/ea2ccb00003611ec9062a5466e060d5114a3f4d7/content/en/docs/gitops-toolkit/source-watcher.md)
- [ ] flux2: https://github.com/fluxcd/flux2/blob/789aa322f19307054d595371934eb555954312ab/cmd/flux/tree_kustomization.go
- [ ] go-git-providers: https://github.com/fluxcd/go-git-providers/blob/5539357bac43be3035cb347eac103c25418b2597/stash/integration_suite_test.go and https://github.com/fluxcd/go-git-providers/blob/cda93bf5a5fa65bd994a60d6d3ef9ad119cfb684/gitlab/integration_test.go and https://github.com/fluxcd/go-git-providers/blob/cda93bf5a5fa65bd994a60d6d3ef9ad119cfb684/github/integration_test.go

dholbach:
Everything else is `flux`, `flux-recv` and `helm-operator` from what I can see.

dholbach:
`source-watcher` and `website` should be fixed separately - anyone interested in picking this up?

PR #1682, opened 2021-08-04
Refactor test helpers

stefanprodan:
Changes:
- move test helpers to main
- add support for inline golden values
- add test for `flux --version`


hiddeco (commented):
- [x] Thank you Stefan 🙏

hiddeco (approved):

allenporter:
Thanks, I had a similar PR in mind with some follow up improvements.

stefanprodan:
Oh one more thing, the test helpers are using `ioutil` this needs to go away, see https://github.com/fluxcd/flux2/issues/1658

allenporter:
Thanks, I was browsing issues and saw that and realized, I'll send a fix.

allenporter:
Moving to TestMain sounds good, will do. I'm assuming all tests that have times/dates in the golden files will need that setup -- that is a `TestMain` per test file. Is that right?

allenporter:
I realize now, TestMain can do setup and teardown to handle restoring any environment variables

Issue #1600, opened 2021-07-07
unlike create secret git, flux create secret helm does not show a success message 


chanwit:
While writing some test cases, I found that `flux create secret helm` does not show a success message, unlike `create secret git`.

Maybe we could improve this a bit by printing out a message like,

```
► secret 'helm-test' created in 'flux-system' namespace
```
to be consistent and also useful for testing.

PR #1683, opened 2021-08-04
feat: display success message while create secrets via `flux` cli.

charles-woshicai:
fix: #1600


stefanprodan (approved):
LGTM

Thanks @charles-woshicai 🏅

PR #1688, opened 2021-08-04
Replace init() with TestMain()

allenporter:
Replace use of `init()` function with a `TestMain` which is the right way to do test setup and teardown.

### DIFF
--- a/cmd/flux/main_test.go
+++ b/cmd/flux/main_test.go
@@ -17,9 +17,10 @@ import (
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-func init() {
+func TestMain(m *testing.M) {
 	// Ensure tests print consistent timestamps regardless of timezone
 	os.Setenv("TZ", "UTC")
+	os.Exit(m.Run())
 }
 
 func readYamlObjects(objectFile string) ([]client.Object, error) {


cmd/flux/main_test.go
```diff
@@ -18,9 +18,10 @@ import (
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-func init() {
+func TestMain(m *testing.M) {
 	// Ensure tests print consistent timestamps regardless of timezone
-	os.Setenv("TZ", "UTC")
+	os.Setenv("TZ", "US")
```
stefanprodan, line 23: Let's do UTC please

stefanprodan (changes requested):

stefanprodan:
Please rebase with upstream main.

cmd/flux/main_test.go
allenporter: Ah, whoops, good catch, I set this to try to make sure it actually was still running to reproduce a test failure.

Fixed.

hiddeco:
We may also want to seed random while we are at it.

stefanprodan (approved):

allenporter:
@hiddeco A couple questions:
1) i'm new to the code base but curious where this will come up: Is there randomness w.r.t. id generation somewhere?
2) i'm also a but new to go testing practices, so i looked up and saw this [discussion](https://stackoverflow.com/questions/50653929/seed-random-number-for-unit-tests-in-golang) where folks seemed to  indicate that test order was not deterministic in go so therefore the seeds would be different anyway.  
My initial thought is to adding a fixed random seed as needed (e.g. required by tests since we're asserting on some particular kube state or printed output) but i don't have enough context to advocate for adding it now.

hiddeco:
You can indeed ignore my comment, it is fine as is for now. I was mislead by https://github.com/fluxcd/source-controller/blob/273f8b155eb267e981120c38cdcccc0440659a2e/controllers/suite_test.go#L165, but this does not apply here.

allenporter:
> You can indeed ignore my comment, it is fine as is for now. I was mislead by https://github.com/fluxcd/source-controller/blob/273f8b155eb267e981120c38cdcccc0440659a2e/controllers/suite_test.go#L165, but this does not apply here.

I could see that maybe helping to reproduce flakes if we know the seed. Anyway thanks for the thoughtful review

PR #1695, opened 2021-08-06
Add `logs` flags `--since` and `--since-time`

ttulka:


allenporter:
So cool to see the tests getting use!

stefanprodan:
@ttulka sorry this took so long, but we've now sorted out the testing framework thanks to @allenporter. Can you please rebase with main upstream and rewrite the tests? Thank you!

PR #1696, opened 2021-08-07
Add tests for create source git

allenporter:
Add unit tests for `create source git`.

The create source tests are more interesting than the existing unit tests as they
create objects then wait for the flux source reconciler to complete. The tests
simulate this with a background goroutine that waits for an object to be
created then uses a test specific function to update it.  The reconciler was
implemented assuming that it might be possible to make it more generic
in the future if needed by additional tests.

The tests set a timeout so that if there is a failure they timeout somewhat
quickly rather than hanging for a longer period of time.

Issue #1603

### DIFF
--- a/cmd/flux/create_source_git_test.go
+++ b/cmd/flux/create_source_git_test.go
@@ -1,0 +1,131 @@
+// +build unit
+
+package main
+
+import (
+	"context"
+	"github.com/fluxcd/pkg/apis/meta"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+	"time"
+)
+
+var pollInterval = 50 * time.Millisecond
+var testTimeout = 10 * time.Second
+
+// Update the GitRepository once created to exercise test specific behavior
+type reconcileFunc func(repo *sourcev1.GitRepository)
+
+// reconciler waits for an object to be created, then invokes a test supplied
+// function to mutate that object, simulating a controller.
+// Test should invoke run() to run the background reconciler task which
+// polls to wait for the object to exist before applying the update function.
+// Any errors from the reconciler are asserted on test completion.
+type reconciler struct {
+	client    client.Client
+	name      types.NamespacedName
+	reconcile reconcileFunc
+}
+
+// Start the background task that waits for the object to exist then applies
+// the update function.
+func (r *reconciler) run(t *testing.T) {
+	result := make(chan error)
+	go func() {
+		defer close(result)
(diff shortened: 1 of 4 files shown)


allenporter:
@chanwit as an FYI

chanwit:
Thank you, Alen.

hiddeco (approved):
LGTM, and seems to be indeed the fix for the other test-related PR :-)

allenporter:
@hiddeco i'm having trouble merging this with the e2e tests that were just added. Do you have a workflow for running the  e2e tests with extra logging enabled? My iteration cycle is pretty high having to wait for github actions to run -- or perhaps you can lend a pair of eyes and spot if i added a bug.

hiddeco:
@chanwit @stefanprodan anything standing out from the PR that got merged into `main` that could cause conflicts with this one? I can't find an obvious reason on a first scan.

stefanprodan:
We can’t use the install test, I’m dropping it, PR will follow.

stefanprodan:
Please see https://github.com/fluxcd/flux2/pull/1721

stefanprodan:
@allenporter @hiddeco @chanwit I'm for dropping the mock API and use envtest.

chanwit:
No objection @stefanprodan

stefanprodan:
In the latest controller-runtime release, the testenv startup time has improve a lot, I think we need to port the `flux trace` from mock the API to testenv (without kind) and compare the timing.

cmd/flux/install_test.go
```diff
@@ -0,0 +1,50 @@
+// +build e2e
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+func TestInstallNoArgs(t *testing.T) {
```
stefanprodan, line 10: There is no such thing in main branch. Please rebase and start fresh.

hiddeco:
> Is this a useful direction or should I abandon this?

I think the new direction (yours) is more useful and we should probably look into transforming the others.

allenporter:
> > Is this a useful direction or should I abandon this?
> 
> I think the new direction (yours) is more useful and we should probably look into transforming the others.

OK if we get the unit tests using testenv then maybe there will be less of a distinction between the two scenarios and we can leave a few e2e tests.

stefanprodan:
Looks like the client is not wired correctly now, no types are registered.

allenporter:
> Looks like the client is not wired correctly now, no types are registered.

Correct -- the changes to the test harness to support the e2e end test have broken the fake harness wiring via `rootCtx.kubeManager`. 

More detail:
-  `main_e2e_test` invokes the test setup and `rootCtx.kubeManager`
- The tests themselves run again and run `cmdTestCase.runTestCmd` which also sets up the test environment and overrides `rootCtx.kubeManager` again.
- The PR to update the e2e harness introduced a test that calls `create source git` which is the same thing that is being tested in this PR coincidentally
- This PR updates `create source git` to use the  `rootCtx.kubeManager` to create the client library for the test since its tested with the fake.
- Most of the code has not yet been updated to use `rootCtx.kubeManager` so it works as is with the e2e test by setting the flags.

I'll start unwinding this...

hiddeco (approved):
LGTM, thank you @allenporter :100:

Issue #1686, opened 2021-08-04
Add unit tests for flux create secret


stefanprodan:
Using #1671 as an example we could add unit tests for the `flux create secret git/helm/tls --export` commands and validate the YAML output against golden file.

PR #1699, opened 2021-08-08
Adding tests to flux create secret commands

alex-petrov-vt:
This is my initial attempt at adding tests to flux create secret commands as suggested in issue #1686 based on the tests utilities added as part of #1671. Let me know if this is what you had in mind or if you want me to make changes.

Incidentally, I wasn't able to create a test for a passwordless private ssh key for a flux create secret git command. When I run  
```
flux create secret git podinfo-auth \
--url=ssh://git@github.com/stefanprodan/podinfo \
--private-key-file=rsa.private \
--export
```
flux correctly generates the secret but if I do it from a test
```go
func TestCreateGitSSHSecret(t *testing.T) {
	cmd := cmdTestCase{
		args:       "create secret git podinfo-auth --url=ssh://git@github.com/stefanprodan/podinfo --private-key-file=./testdata/create/secret/git/rsa.private --export",
		wantError:  false,
		goldenFile: "testdata/create/secret/git/git-ssh-secret.yaml",
	}
	cmd.runTestCmd(t)
}
```
I get 
```
--- FAIL: TestCreateGitSSHSecret (0.00s)
    main_test.go:128: Expected error='false', Got: ssh: key is not password protected
```

So I'm still trying to figure out what the issue might be with that.


stefanprodan (approved):
LGTM

Thanks @alex-petrov-vt 🎖️

stefanprodan:
@hiddeco could you please take a look at Alex comment?

hiddeco:
Seems to work just fine to me, added a test to proof it that works locally.

hiddeco:
Seems to be that something in the test setup isn't right, running the individual test case succeeds, running them all fails.

allenporter:
Perhaps the harness improvements in 
https://github.com/fluxcd/flux2/pull/1696 help this. I believe it's reusing a fake ckube client across tests in the existing version and in the write test I added, I made it create a new one every time (as well as supporting more complex write test cases where the command waits for a reconcile to happen in a background)

allenporter:
Hi, a bunch of the test harness has now been updated, though it means you're going to have to do some rewriting. I would recommend taking a look at the trace tests. In particular the things to look for are:
- The move to `assert` functions rather than multiple parameters to configure golden files
- Tests used a shared testEnv so you'll want to the `allocateNamespace` function and templates for the output files

alex-petrov-vt:
Hi Allen, thanks for the info. I'll try to take a look at those this weekend

alex-petrov-vt:
Hi, unfortunately I'm currently busy with other stuff and can't really dedicate too much time to contributing. Hopefully, I can revisit at some point later down the road. Thanks for your help!

Issue #1711, opened 2021-08-10
Inconsistent version check during bootstrap


jooooel:
### Describe the bug

Providing a version without a `v` prefix during bootstrap, will default to the latest version instead.

The following will not install v0.13.4, and it will not show any error or information.

```
flux bootstrap github \
		--owner=<owner> \
		--repository=<repo> \
		--branch=main \
		--path=./clusters/${{ env.CLUSTER_NAME }} \
		--version=0.13.4 \
		--private=true \
		--personal
```

Flux CLI will prefix the version with `v` when checking if the version exists, so it will find the consider it valid. But when the manifests are actually being pulled, the (invalid) version without v is being used, then defaulting to latest.

I ran into this and described the problem here: https://cloud-native.slack.com/archives/CLAJ40HV3/p1622464854135100

### Steps to reproduce

The following will install the latest manifest versions, instead of the version provided. 

```
flux bootstrap github \
		--owner=<owner> \
		--repository=<repo> \
		--branch=main \
		--path=./clusters/${{ env.CLUSTER_NAME }} \
		--version=0.13.4 \
		--private=true \
		--personal
```

### Expected behavior

That the manifests for v0.13.4 are installed or an error or some information is being displayed.

### OS / Distro

N/A

### Flux version

N/A

### Flux check

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1712, opened 2021-08-10
Validate input version and prefix with 'v' if missing

jooooel:
This is a fix for https://github.com/fluxcd/flux2/issues/1711

* Move prefixing to `getVersion` so that it will be used both by the initial check and when downloading the manifests.
* Handle errors from `getVersion` so that bootstrap is aborted for invalid versions.

Signed-off-by: Joel Guedj <joel.guedj@gmail.com>

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -88,7 +88,7 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringSliceVar(&bootstrapArgs.defaultComponents, "components", rootArgs.defaults.Components,
 		"list of components, accepts comma-separated values")
 	bootstrapCmd.PersistentFlags().StringSliceVar(&bootstrapArgs.extraComponents, "components-extra", nil,
-		"list of components in addition to those supplied or defaulted, accepts comma-separated values")
+		"list of components in addition to those supplied or defaulted, accepts values such as 'image-reflector-controller,image-automation-controller'")
 
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapArgs.registry, "registry", "ghcr.io/fluxcd",
 		"container registry where the toolkit images are published")
--- a/cmd/flux/install.go
+++ b/cmd/flux/install.go
@@ -37,10 +37,13 @@ var installCmd = &cobra.Command{
 	Long: `The install command deploys Flux in the specified namespace.
 If a previous version is installed, then an in-place upgrade will be performed.`,
 	Example: `  # Install the latest version in the flux-system namespace
-  flux install --version=latest --namespace=flux-system
+  flux install --namespace=flux-system
 
-  # Install a specific version and a series of components
-  flux install --version=v0.0.7 --components="source-controller,kustomize-controller"
+  # Install a specific series of components
+  flux install --components="source-controller,kustomize-controller"
+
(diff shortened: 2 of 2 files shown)


cmd/flux/bootstrap_git.go
```diff
@@ -107,9 +107,12 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 	}
 
 	// Manifest base
-	if ver, err := getVersion(bootstrapArgs.version); err == nil {
+	if ver, err := getVersion(bootstrapArgs.version); err != nil {
```
stefanprodan, line 110: We can't return an error if `getVersion` fills as it will break Flux on air-gapped systems, flux install and bootstrap should make work even if external calls to e.g. GitHub fail.

pkg/manifestgen/install/install.go
```diff
@@ -116,10 +116,6 @@ func GetLatestVersion() (string, error) {
 
 // ExistingVersion calls the GitHub API to confirm the given version does exist.
 func ExistingVersion(version string) (bool, error) {
-	if !strings.HasPrefix(version, "v") {
```
stefanprodan, line 119: We can't remove this from here as it breaks https://github.com/fluxcd/terraform-provider-flux

cmd/flux/version.go
```diff
@@ -12,6 +13,10 @@ func getVersion(input string) (string, error) {
 		return rootArgs.defaults.Version, nil
 	}
 
+	if !strings.HasPrefix(input, "v") {
+		input = "v" + input
+	}
+
```
jooooel, line 19: So basically this is the only change that I should keep @stefanprodan? (If you are interested in this PR)

stefanprodan: I'm for bailing out out when the version is not found instead of defaulting to latest.

jooooel: How do you propose we do that? If bootstrap and install must work if external calls to GitHub fail, we'd need to do something like this:

Abort bootstrap/install if:
* Version check against GH succeeded, but version not found.

Do not abort bootstrap/install if:
* Version check against GH succeeded, version found.
* Version check against GH failed.

I also noticed that there are discrepancies between how install and bootstrap work at the moment, when passing in different kinds of versions:

**Install:**

Version check fails, install aborted (I guess this goes against "should work in an air gapped env"? 🤔):
0.xxx.0
v0.xxx.0

Version check succeeds, installs latest:
0.13.1

Version check succeeds, installs correct:
v0.13.1

**Bootstrap:**

Version check fails, bootstraps latest:
0.xxx.0
v0.xxx.0

Version check succeeds, bootstraps latest:
0.13.1

Version check succeeds, boostraps correct:
v0.13.1

So I'm not really sure what to do? 😄

stefanprodan: We could create a custom flag that validates the string is a valid semver and that it has the v prefix.

jooooel: A valid semver could still be a "not found" version, so that would still default to latest, which from a users perspective is a bit weird I think.

What I would expect as a user:
* Invalid version -> Install/Boostrap fails
* No version set -> Default to latest

But if we make that change, I guess it would be a breaking change for air gapped envs (force them to not set a version, so that no external call is made).

PR #1721, opened 2021-08-17
Refactor e2e tests

stefanprodan:
Changes:
- drop install/uninstall tests due to namespace finalizer taking 1-5 minutes
- add install/uninstall to e2e setup/teardown
- split tests by tag `unit` (testenv only) and `e2e` (testenv + kind)
- add e2e tests for Kustomization and HelmRelease operations
- add e2e tests for image scanning
- isolate e2e tests in dedicated namespaces
- run e2e tests in parallel


cmd/flux/kustomization_test.go
```diff
@@ -0,0 +1,57 @@
...
+			"resume kustomization tkfg",
+			"testdata/kustomization/resume_kustomization_from_git.golden",
+		},
+		{
+			"delete kustomization tkfg --silent",
+			"testdata/kustomization/delete_kustomization_from_git.golden",
+		},
+	}
+
+	namespace := "tkfg"
```
hiddeco, line 42: I would use a func which adds a random suffix here (and in other places) to ensure accidental copy/paste without changes do not result in flakiness.

stefanprodan: We can't do this since the namespace is in the gold file.

hiddeco: Light templating? But valid reason to not do this now.

hiddeco (approved):
Thank you @stefanprodan 💐

cmd/flux/kustomization_test.go
stefanprodan: No idea how to do this templating, as the whitespace in tables changes based on length. I would unblock CI first then figure out if golden files that are templates make sense.

PR #1726, opened 2021-08-18
Remove fakeclient and use testenv for flux cmd tests

allenporter:
Remove use of the fake client, and replace with a real client connected to the
testEnv.

This required fixes to the yaml files as the testEnv has stricter verifcation
of objects. This also meant it was not possible to test a GitRepository with
a missing artifact since that is not a valid state.

The tests are slower than before, taking around 7-10 seconds each because the
 testEnv is setup and destroyed for every test. These will be sped up in a
 follow up PR.


Makefile
```diff
@@ -39,8 +39,8 @@ cleanup-kind:
 	kind delete cluster --name=flux-e2e-test
 	rm $(TEST_KUBECONFIG)
 
-test: $(EMBEDDED_MANIFESTS_TARGET) tidy fmt vet
-	go test ./... -coverprofile cover.out --tags=unit
+test: $(EMBEDDED_MANIFESTS_TARGET) tidy fmt vet setup-envtest
+	KUBEBUILDER_ASSETS=$(shell $(SETUP_ENVTEST) use -p path) go test ./... -coverprofile cover.out --tags=unit
```
stefanprodan, line 43: Please see https://github.com/fluxcd/pkg/blob/main/Makefile#L6 as it is now it will fail on macOS.

Makefile
```diff
@@ -25,7 +25,7 @@ ifeq (, $(shell which setup-envtest))
 	go get sigs.k8s.io/controller-runtime/tools/setup-envtest@latest ;\
 	rm -rf $$SETUP_ENVTEST_TMP_DIR ;\
 	}
-SETUP_ENVTEST=$(GOBIN)/setup-envtest
+SETUP_ENVTEST=$(GOPATH)/bin/setup-envtest
```
stefanprodan, line 28: Please use https://github.com/fluxcd/pkg/blob/main/Makefile#L8-L13

Makefile
allenporter: I'm going to need more hand holding here.

I copied this from that file, but am not sure how SETUP_ENVTEST is set there and how to reconcile this with the setup envtest lines above

stefanprodan: IF you undo the changes to Makefile and rebase with main it should be good.

allenporter: Thanks! I'll fix the e2e error also and rebase everything.

Can you explain how SETUP_ENVTEST is set outside of the GitHub action when running locally? Is `setup_envtest:` running implicitly somewhere to set that variable at the start?

stefanprodan: On your local machine you would just run `make test` and it should install the envtest binaries then run the unit tests. For e2e tests, you run first `make setup-kind` then `make e2e`, this will leave the kind cluster running so you debug things or run `make test-with-kind` that will create a cluster each time, run e2e then delete the cluster.

allenporter: Right, that was the case with this PR too. To clarify, I'm not asking a workflow question, I'm asking a makefile technical question on how the variable is set to increase my understanding (I'm trying to not blindly just copy the code from the other package.). Thanks for your patience.

stefanprodan: So `make test` runs `setup-envtest` then runs the `go test`.

allenporter: I guess my confusion is that `KUBEBUILDER_ASSETS` is using a variable at the top of the file before `setup-env` even runs. I'm guessing it must not evaluated until its used, and that is what I was missing...

stefanprodan (approved):
LGTM 

Thanks @allenporter

PR #1727, opened 2021-08-18
Wire kubebuilder assets to envtest bin

stefanprodan:
Help out with #1726 port from fluxcd/pkg


Makefile
```diff
@@ -1,6 +1,15 @@
 VERSION?=$(shell grep 'VERSION' cmd/flux/main.go | awk '{ print $$4 }' | tr -d '"')
 EMBEDDED_MANIFESTS_TARGET=cmd/flux/manifests
 TEST_KUBECONFIG?=/tmp/flux-e2e-test-kubeconfig
+ENVTEST_BIN_VERSION?=latest
+KUBEBUILDER_ASSETS?="$(shell $(SETUP_ENVTEST) use -i $(ENVTEST_BIN_VERSION) -p path)"
```
allenporter, line 5: Where is $(SETUP_ENVTEST) getting populated?

phillebaba (approved):

Makefile
allenporter: (non blocking)

stefanprodan: Here:
```
# Find or download setup-envtest
setup-envtest:
ifeq (, $(shell which setup-envtest))
	@{ \
	set -e ;\
	SETUP_ENVTEST_TMP_DIR=$$(mktemp -d) ;\
	cd $$SETUP_ENVTEST_TMP_DIR ;\
	go mod init tmp ;\
	go get sigs.k8s.io/controller-runtime/tools/setup-envtest@latest ;\
	rm -rf $$SETUP_ENVTEST_TMP_DIR ;\
	}
SETUP_ENVTEST=$(GOBIN)/setup-envtest
else
SETUP_ENVTEST=$(shell which setup-envtest)
endif
```

stefanprodan: In CI we have our own GitHub Action that builds envtest in container and copies the binaries to the GH runner, here https://github.com/fluxcd/flux2/blob/main/.github/workflows/e2e.yaml#L32

allenporter: I assumed on this like that setup-envtest hasn't run yet to set that variable. I guess don't understand how Makefiles work.

PR #1732, opened 2021-08-19
Avoid substitution issue in kustomize credentials sync

timja:
Signed-off-by: Tim Jacomb <tim.jacomb@hmcts.net>


manifests/integrations/eventhub-credentials-sync/_base/sync.yaml
```diff
@@ -71,16 +71,16 @@ spec:
...
+                  kill $job_ids
                 fi
                 wait
                 echo "Graceful exit -- $(date)"
               }
 
               trap graceful_exit INT TERM
 
-              echo "Loop started (period: ${SYNC_PERIOD} s) -- $(date)"
+              echo "Loop started (period: ${SYNC_PERIOD:-3600} s) -- $(date)"
```
timja, line 83: this is still getting substituted preventing the intended effect of the configmap overriding the value, you can set it via post build if you want otherwise any suggestions for fixing it properly?

(I personally don't need to override this)

timja:
@stefanprodan any chance of a merge? 

Thanks!

stefanprodan (approved):

timja:
@stefanprodan would you be able to take a look please

PR #1739, opened 2021-08-23
Make test harness more flexible with functions

allenporter:
Replace the 4 arguments to cmdTestCase with a function that
can let tests run arbitrary logic if it is more complex than
what is provided by the test harness. Move the existing logic
into functions that the test can use for common checks on
golden files and golden values.

These changes were pulled out of PR #1696 to make a smaller review.

### DIFF
--- a/cmd/flux/check_test.go
+++ b/cmd/flux/check_test.go
@@ -28,13 +28,11 @@ func TestCheckPre(t *testing.T) {
 
 	cmd := cmdTestCase{
 		args:            "check --pre",
-		wantError:       false,
 		testClusterMode: ExistingClusterMode,
-		templateValues: map[string]string{
+		assert: assertGoldenTemplateFile("testdata/check/check_pre.golden", map[string]string{
 			"clientVersion": clientVersion,
 			"serverVersion": serverVersion,
-		},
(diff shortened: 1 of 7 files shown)


cmd/flux/check_test.go
```diff
@@ -28,13 +28,11 @@ func TestCheckPre(t *testing.T) {
 
 	cmd := cmdTestCase{
 		args:            "check --pre",
-		wantError:       false,
 		testClusterMode: ExistingClusterMode,
-		templateValues: map[string]string{
+		check: expectGoldenTemplateFile("testdata/check/check_pre.golden", map[string]string{
```
stefanprodan, line 32: Would be better id we use `assert` instead of `check`?

allenporter: I have renamed all uses of `check` to `assert`. Also, your comment prompted me to do the same for `expect` which I also renamed to `assert`.  My recollection is that `expect` typically proceeds to keep running the test after a failure, where `assert` stops the test in place, and I think these are more like assertions than expectations.

stefanprodan (approved):
LGTM

Thanks @allenporter 🏅

stefanprodan:
Allen thanks a lot for splitting that PR into smaller changes!

PR #1740, opened 2021-08-23
Rename trace test golden files

allenporter:
Rename trace test golden files to match the convention used by other tests


stefanprodan:
Please rebase with upstream main.

stefanprodan (approved):
LGTM

Thanks @allenporter

PR #1743, opened 2021-08-23
Use shared envTest for unit tests

allenporter:
Speed up unit tests by using a shared envTest. This requires each
test to use its own namespace to avoid clobbering objects for
other tests. Tests previously took around 8 seconds each, and now
the initial test takes 2 seconds with follow up tests taking less
than a second each.

Share gold file template function with yaml files.

Remove the testClusterMode, and instead rely on MainTest to do
the appropriate test setup and rootArgs flag setup. Move the
rootArg flag setup out of NewTestEnvKubeManager to avoid
side effects.

A follow up change can be to push the individual setups
from NewTestEnvKubeManager() into their respective TestMain since
the harness share little code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,16 @@ You can run the unit tests by simply doing
 make test
 ```
 
+The e2e test suite uses [kind](https://kind.sigs.k8s.io/) for running kubernetes cluster inside docker containers. You can run the e2e tests by simply doing
+
+```bash
+make setup-kind
+make e2e
+
+# When done
+make cleanup-kind
+```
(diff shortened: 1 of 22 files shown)


stefanprodan:
Can you please use the namespace generator in all tests, there are 4 more tests with hardcoded namespaces now.

allenporter:
> Can you please use the namespace generator in all tests, there are 4 more tests with hardcoded namespaces now.

Done. I found 3 using the e2e setup namespace function.

stefanprodan (approved):
LGTM

Thanks @allenporter ☕

PR #1791, opened 2021-09-07
Fix Bucket SecretRef mapping

fynluk:
Found a bug while creating a bucket source with the ```--secret-ref``` Flag.

Signed-off-by: Fynn Späker <spaeker@23technologies.cloud>


mxmxchere (approved):
yes sure, has to be sourceBucketArgs

stefanprodan (approved):
LGTM

Thanks @fynluk

PR #1816, opened 2021-09-13
Add Azure E2E tests

phillebaba:
This change adds a larger set of tests to Azure specific functions. The plan is that the tests should be run nightly against the main branch and use a custom subscription to run the tests in.


tests/azure/util_test.go
```diff
@@ -0,0 +1,461 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	git2go "github.com/libgit2/git2go/v31"
```
stefanprodan, line 12: We can’t have this in flux CLI, please create a separate package with its own go.mod inside the test dir.

phillebaba: Done

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/fluxcd/flux2
 go 1.16
 
 require (
+	github.com/Azure/go-autorest/autorest v0.11.18 // indirect
```
stefanprodan, line 6: Can you undo these? I see no code changes in the CLI so I think we don't need any of these packages.

tests/azure/go.mod
```diff
@@ -0,0 +1,106 @@
+module github.com/fluxcd/flux2/tests/azure
+
+go 1.17
```
stefanprodan, line 3: Please use go 1.16 bumping the version to 1.17 should happen first in fluxcd/pkg then controllers and lastly flux2.

.github/workflows/e2e-azure.yaml
```diff
@@ -0,0 +1,64 @@
+name: e2e-azure
+
+on:
+  workflow_dispatch:
```
stefanprodan, line 4: Can you please add here branch `azure/e2e` to see if the tests are passing?

phillebaba: Tests have been run fully without any issues.

stefanprodan (approved):
LGTM

Thanks @phillebaba awesome work 🥇

PR #1828, opened 2021-09-15
Use a file to record successful manifests build

squaremo:
Using the directory cmd/flux/manifests as a prerequisite causes a problem: if the script that creates the files within fails, the next invocation of make will see the directory and assume it succeeded. Since the executable expects certain files to be present, but they are not explicit prerequisites of the recipe for building the binary, this results in a successful build but a broken `flux` executable.

Instead, depend on a file that's explicitly updated when the script has succeeded, and which itself depends on the inputs.

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -33,7 +33,7 @@ jobs:
         uses: fluxcd/pkg//actions/kustomize@main
       - name: Build
         run: |
-          make cmd/flux/manifests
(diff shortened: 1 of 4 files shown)


hiddeco (approved):
Seems like the right fix for the time being, the fixup commit needs to be squashed however. Thanks @squaremo :sunflower:

squaremo:
> the fixup commit needs to be squashed however

Of course, in fact I'm going to squash it into one commit -- the Makefile change doesn't make sense without the CI change.

Issue #971, opened 2021-02-23
Cannot seem to get the completions working in zsh


imle:
### Describe the bug

Running the `. <(flux completion zsh)` command in the terminal or adding it to the ~/.zshrc file produce no completion results. changing the `.` to `source` also produce no results.

### To Reproduce

1. Install flux via brew or the install.sh
2. Run `. <(flux completion zsh)`
3. Experience no completion

### Expected behavior

Completions occur when tab is used.

### Remediation
Running `flux completion zsh > /usr/share/zsh/site-functions/_flux` allows the completions to work but they will not remain updated.

Below please provide the output of the following commands (omitting most of the kubectl commands for brevity since they are not providing the completions):

```cli
$ flux --version
flux version 0.8.2

$ flux check
► checking prerequisites
✔ kubectl 1.20.2 >=1.18.0-0
✔ Kubernetes 1.19.4 >=1.16.0-0
► checking controllers
✔ helm-controller: healthy
► ghcr.io/fluxcd/helm-controller:v0.7.0
✔ kustomize-controller: healthy
► ghcr.io/fluxcd/kustomize-controller:v0.8.1
✔ notification-controller: healthy
► ghcr.io/fluxcd/notification-controller:v0.8.0
✔ source-controller: healthy
► ghcr.io/fluxcd/source-controller:v0.8.1
✔ all checks passed

$ kubectl -n flux-system get all
NAME                                           READY   STATUS    RESTARTS   AGE
pod/helm-controller-6779d46d69-qtsx5           1/1     Running   0          31m
pod/kustomize-controller-55dfcdfd58-z724s      1/1     Running   0          31m
pod/notification-controller-649754966b-kjls8   1/1     Running   0          31m
pod/source-controller-597cc769b-94xpb          1/1     Running   0          31m

NAME                              TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)   AGE
service/notification-controller   ClusterIP   10.43.3.207   <none>        80/TCP    31m
service/source-controller         ClusterIP   10.43.79.56   <none>        80/TCP    31m
service/webhook-receiver          ClusterIP   10.43.82.36   <none>        80/TCP    31m

NAME                                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/helm-controller           1/1     1            1           31m
deployment.apps/kustomize-controller      1/1     1            1           31m
deployment.apps/notification-controller   1/1     1            1           31m
[...]

PR #1833, opened 2021-09-17
Fix `flux completion zsh` to work with `source` only

superbrothers:
This PR fixes `flux completion zsh` to work with `source` only.

`flux completion zsh` does not work with `source <(flux completion zsh)`, and needs to be executed as follows:

```
source <(flux completion zsh) && compdef _flux flux
```

This is confusing to Kubernetes users because `kubectl completion zsh` works well with `source <(kubectl completion zsh)`.

Fixes https://github.com/fluxcd/flux2/issues/971

### DIFF
--- a/cmd/flux/completion_zsh.go
+++ b/cmd/flux/completion_zsh.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM 

Thanks @superbrothers 🏅

PR #1854, opened 2021-09-23
Allow users to use gpg signing for bootstrap commits

### LINKED ISSUE #1774: flux bootstrap does not sign its commits
### Describe the bug

Hello,

`flux bootstrap` does not GPG sign its commits, leading to "no PGP signature found for commit: [...]" on cluster with GPG commit verification enabled.

### Steps to reproduce

1. Create a cluster managed by Flux with [GPG commit verification](https://fluxcd.io/docs/components/source/gitrepositories/#gpg-signature-verification) enabled

2. Run
```yaml
flux bootstrap git \
  --url=ssh://git@github.com/xxx/xxx.git \
  --private-key-file flux.ssh.key \
  --silent \
  --branch=master \
  --path=cluster/ \
  --interval=1m
```

3. If flux bootstrap makes a commit (e.g., on upgrade, on config change), the commit is not GPG signed, therefore it will not be applied to the cluster.

### Expected behavior

flux should sign its commits with an user-defined GPG key (e.g., CLI arg for key fingerprint)

_No response_

### OS / Distro

N/A

### Flux version

0.17.0

### Flux check

N/A

### Git provider

_No response_

_No response_

### Additional context

_No response_

### Code of Conduct

SomtochiAma:
Close #1774


hiddeco (commented):
Thinking about future expansions of this feature, is it possible in Go to detect the user's running GPG agent?

internal/bootstrap/git/git.go
```diff
@@ -34,8 +34,10 @@ type Author struct {
 
 type Commit struct {
 	Author
-	Hash    string
-	Message string
+	Hash          string
+	Message       string
+	GpgKeyPath    string
+	GpgPassphrase string
```
hiddeco, line 40: Think this should not be added to this structure, as it is kind of pre-flight data which only has a purpose while a commit is made, and contains data you generally do not want to keep around for too long because of their sensitive nature.

internal/bootstrap/options.go
```diff
@@ -112,3 +112,24 @@ func (o loggerOption) applyGit(b *PlainGitBootstrapper) {
 func (o loggerOption) applyGitProvider(b *GitProviderBootstrapper) {
 	b.logger = o.logger
 }
+
+func WithGitCommitSigning(path, passphrase string) Option {
```
hiddeco, line 116: I think the `WithGitCommitSigning` (or `WithCommitSigning`) should be reserved for the most UX-friendly approach (trying to determine the user's agent), and this advanced option should be named something like: `WithCommitSigningKeyRing()` (or something else that properly explains what it entails).

internal/bootstrap/git/gogit/gogit.go
```diff
@@ -232,3 +244,26 @@ func (g *GoGit) Path() string {
...
+	}
+
+	if len(entityList) == 0 {
+		return nil, fmt.Errorf("no entity formed")
+	}
+
+	err = entityList[0].PrivateKey.Decrypt([]byte(passphrase))
+	if err != nil {
+		return nil, err
+	}
```
hiddeco, line 266: You are making an assumption here that a key ring only contains a single key. While for some people this might be true, others may contain multiple, and you probably want to provide people the option to select what key from the ring should be used (using e.g. a fingerprint selector of the public key).

PR #1855, opened 2021-09-23
Add shell completion installation to homebrew formula

darklore:
This PR adds installation of shell completion script to homebrew formula.


stefanprodan (approved):
LGTM

Thanks @darklore ☕

PR #1893, opened 2021-10-04
Add flux version command

### LINKED ISSUE #1892: Implement flux version command
Implement `flux version` commands for both client and server-side components with the following specs.

Example output of `flux version -h`:

```
Print the client and server-side components version information for the current context.

Flags:
      --client             if true, shows client version only (no server required)
  -o, --output string      one of 'yaml' or 'json'  (default "yaml")

Global Flags:
      --kubeconfig string   absolute path to the kubeconfig file
```

Example output of `flux version --client -o yaml`:

```yaml
flux: v0.17.0
```

Example output of `flux version -o yaml`:

```yaml
flux: v0.17.0
source-controller: v0.15.4
kustomize-controller: v0.14.0
helm-controller: v0.11.2
notification-controller: v0.16.0
image-reflector-controller: v0.11.1
image-automation-controller: v0.14.1
```

The components listing could use the same code as `flux check`. The component name and version should be extracted from the container image. Given that the Flux components except for source-controller are optional, we should error out only when the server is not accessible or when no components are found in the specified namespace.

SomtochiAma:
Close #1892 


cmd/flux/version.go
```diff
@@ -1,42 +1,113 @@
...
 	}
 
-	var err error
-	if input == install.MakeDefaultOptions().Version {
-		input, err = install.GetLatestVersion()
+	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
+	defer cancel()
+
+	info := map[string]string{}
+	info["flux"] = rootArgs.defaults.Version
```
stefanprodan, line 58:
```suggestion
	info["flux"] = fmt.Sprintf("v%s", rootArgs.defaults.Version)
```

cmd/flux/version.go
```diff
@@ -1,42 +1,113 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
```
stefanprodan, line 10:
```suggestion
	"sigs.k8s.io/yaml"
```

cmd/flux/version_utils.go
```diff
@@ -0,0 +1,42 @@
+package main
```
stefanprodan, line 1: Missing copyright header here.

cmd/flux/version.go
```diff
@@ -1,42 +1,113 @@
 package main
```
stefanprodan, line 1: Missing copyright header here.

cmd/flux/version.go
```diff
@@ -1,42 +1,129 @@
...
+	} else {
+		marshalled, err = yaml.Marshal(&info)
+	}
+
+	if err != nil {
+		return err
 	}
-	return input, nil
+
+	fmt.Fprintln(os.Stdout, string(marshalled))
```
stefanprodan, line 116:
```suggestion
	rootCmd.Print(string(marshalled))
```

stefanprodan (commented):
@SomtochiAma please add two tests for `version --client` and `version --client -o json`, see version_utils_test.go for an example.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma please rebase and squash.

PR #1898, opened 2021-10-06
Fix "get" commands to use stdout instead of stderr

superbrothers:
This PR fixes "get" subcommands to use stdout instead of stderr. "get" subcommands use stderr for output. This is not convenient to use it with grep, etc. 

The following PR changed to use stderr instead of stdout. But perhaps this is not the intended change.

- https://github.com/fluxcd/flux2/pull/1721/files#diff-37418cfe87a44e2ee792ac89d68362c5b44aa784adfccad9b2e7c280c97e1480R180


stefanprodan (approved):
LGTM

Thanks @superbrothers good catch!

PR #2018, opened 2021-10-27
Update Go to v1.17

relu:


stefanprodan:
This was included in #2271

PR #2034, opened 2021-11-01
Set default ns with `FLUX_SYSTEM_NAMESPACE` env var

stefanprodan:
This PR allows users to change the default namespace from `flux-system` to some other value using the `FLUX_SYSTEM_NAMESPACE` environment variable. 

Ref: https://github.com/fluxcd/flux2/discussions/1284

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -107,7 +107,8 @@ type rootFlags struct {
 var rootArgs = NewRootFlags()
 
 func init() {
-	rootCmd.PersistentFlags().StringVarP(&rootArgs.namespace, "namespace", "n", rootArgs.defaults.Namespace, "the namespace scope for this operation")
+	rootCmd.PersistentFlags().StringVarP(&rootArgs.namespace, "namespace", "n", rootArgs.defaults.Namespace,
+		"the namespace scope for this operation, can be set with FLUX_SYSTEM_NAMESPACE env var")
 	rootCmd.RegisterFlagCompletionFunc("namespace", resourceNamesCompletionFunc(corev1.SchemeGroupVersion.WithKind("Namespace")))
 
 	rootCmd.PersistentFlags().DurationVar(&rootArgs.timeout, "timeout", 5*time.Minute, "timeout for this operation")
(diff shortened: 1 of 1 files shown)


hiddeco:
How can it be that your example contains `gitops-system` default in the doc note?

```
  -n, --namespace string    the namespace scope for this operation, can be set with FLUX_SYSTEM_NAMESPACE env var (default "gitops-system")
```

cmd/flux/main.go
```diff
@@ -152,6 +153,16 @@ func configureKubeconfig() {
...
 
+func getDefaultNamespace() string {
+	fromEnv := os.Getenv("FLUX_SYSTEM_NAMESPACE")
+	switch {
+	case fromEnv != "":
+		return fromEnv
+	default:
+		return rootArgs.defaults.Namespace
+	}
+}
```
hiddeco, line 164: The logged doc default should not take the set env into account I think.

stefanprodan: I think it should, people should be aware they've changed the default, `--help` would display that.

stefanprodan:
> How can it be that your example contains gitops-system as default in the doc note?

This is what the current PR does, it allows people to override the **default value** of `--namespace` with an env var.

cmd/flux/main.go
hiddeco: But that no longer makes it display the default, but rather the current configuration. The default is still `flux-system` if the environment variable is `unset`, which is the information I would like to get from documentation.

Now I no longer know what it falls back to if I would `unset`, wile I could still do an `env | grep FLUX_SYSTEM_NAMESPACE` after having read "can be set with [..]".

stefanprodan: > Now I no longer know what it falls back to if I would unset it

```go
func getDefaultNamespace() string {
	fromEnv := os.Getenv("FLUX_SYSTEM_NAMESPACE")
	switch {
	case fromEnv != "":
		return fromEnv
	default:
		return rootArgs.defaults.Namespace
	}
}
```

it goes to `default` in the switch.

hiddeco: But that still requires you to unset the environment variable, which is what I describe above. I am going to need a second opinion on this.

stefanprodan: This works the same as the `KUBECONFIG` env var for kubectl, if found in env it uses that as the default, if you set the arg to some other value it uses that.

hiddeco: Yes, that is how it should work technically which is fine. The point I am not sure about is making the documentation change based on an environment variable that is set, I would expect it to be static.

hiddeco (approved):

PR #2037, opened 2021-11-02
feat: prompt for access tokens in 'bootstrap' command

### LINKED ISSUE #2011: Credentials exposed in environment variables and command-line arguments
Brought up by Ada Logics

> Flux-CLI uses credentials to bootstrap many commands. These are often placed in the
command line or environment variables. This means the credentials are exposed to a wider
audience than intended, namely any internal attacker with a host position if they have
sufficient position to see command line arguments (such as via ps).
For example, if someone has access to the host at which the command line was entered but
not the access to the token of a given Github, then the details can be leaked.

> Make it possible to write sensitive data in way that it won't be exposed, such as: ***** or not
showing any characters when being typed.

makkes:
This change adds functionality to both, `bootstrap github` and
`bootstrap gitlab` to prompt the user for the personal access tokens
if those can't be derived from the shell environment. Echoing is
turned off for better privacy.

Instead of having to interactively type the token or manually paste it
from the clipboard, users can also pipe it to Flux which comes in
handy e.g. when executing Flux remotely over an SSH connection:

```
$ echo 'asdf' | flux bootstrap github
```

Otherwise, Flux will prompt the user:

```
$ flux bootstrap github
Please type your GitHub personal access token:
```

closes #2011

Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -111,7 +111,11 @@ func init() {
 func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 	ghToken := os.Getenv(ghTokenEnvVar)
 	if ghToken == "" {
-		return fmt.Errorf("%s environment variable not found", ghTokenEnvVar)
+		var err error
+		ghToken, err = readPasswordFromStdin("Please type your GitHub personal access token: ")
+		if err != nil {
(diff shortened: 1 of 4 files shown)


makkes:
closing in favor of #2038

PR #2038, opened 2021-11-02
feat: prompt for access tokens in 'bootstrap' command

### LINKED ISSUE #2011: Credentials exposed in environment variables and command-line arguments

makkes:
This change adds functionality to both, `bootstrap github` and
`bootstrap gitlab` to prompt the user for the personal access tokens
if those can't be derived from the shell environment. Echoing is
turned off for better privacy.

Instead of having to interactively type the token or manually paste it
from the clipboard, users can also pipe it to Flux which comes in
handy e.g. when executing Flux remotely over an SSH connection:

```
$ echo 'asdf' | flux bootstrap github
```

Otherwise, Flux will prompt the user:

```
$ flux bootstrap github
Please type your GitHub personal access token:
```

closes #2011

Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -112,7 +112,7 @@ func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 	ghToken := os.Getenv(ghTokenEnvVar)
 	if ghToken == "" {
 		var err error
-		ghToken, err = readPasswordFromStdin("Please type your GitHub personal access token: ")
+		ghToken, err = readPasswordFromStdin("Please enter your GitHub personal access token (PAT): ")
 		if err != nil {
 			return fmt.Errorf("could not read token: %w", err)
 		}
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -111,7 +111,11 @@ func init() {
 func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 	ghToken := os.Getenv(ghTokenEnvVar)
 	if ghToken == "" {
-		return fmt.Errorf("%s environment variable not found", ghTokenEnvVar)
+		var err error
+		ghToken, err = readPasswordFromStdin("Please type your GitHub personal access token: ")
+		if err != nil {
+			return fmt.Errorf("could not read token: %w", err)
+		}
 	}
 
 	if err := bootstrapValidate(); err != nil {
--- a/cmd/flux/bootstrap_gitlab.go
+++ b/cmd/flux/bootstrap_gitlab.go
@@ -109,7 +109,7 @@ func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
 	glToken := os.Getenv(glTokenEnvVar)
 	if glToken == "" {
 		var err error
-		glToken, err = readPasswordFromStdin("Please type your GitLab personal access token: ")
+		glToken, err = readPasswordFromStdin("Please enter your GitLab personal access token (PAT): ")
 		if err != nil {
 			return fmt.Errorf("could not read token: %w", err)
 		}
--- a/cmd/flux/bootstrap_gitlab.go
+++ b/cmd/flux/bootstrap_gitlab.go
@@ -108,7 +108,11 @@ func init() {
 func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
 	glToken := os.Getenv(glTokenEnvVar)
 	if glToken == "" {
-		return fmt.Errorf("%s environment variable not found", glTokenEnvVar)
+		var err error
+		glToken, err = readPasswordFromStdin("Please type your GitLab personal access token: ")
+		if err != nil {
+			return fmt.Errorf("could not read token: %w", err)
+		}
 	}
 
 	if projectNameIsValid, err := regexp.MatchString(gitlabProjectRegex, gitlabArgs.repository); err != nil || !projectNameIsValid {
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -17,12 +17,15 @@ limitations under the License.
 package main
 
 import (
+	"bufio"
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 	corev1 "k8s.io/api/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
@@ -166,4 +169,23 @@ func homeDir() string {
 		return h
 	}
 	return os.Getenv("USERPROFILE") // windows
+}
+
+func readPasswordFromStdin(prompt string) (string, error) {
+	var out string
+	var err error
+	fmt.Fprint(os.Stdout, prompt)
+	stdinFD := int(os.Stdin.Fd())
+	if term.IsTerminal(stdinFD) {
+		var inBytes []byte
+		inBytes, err = term.ReadPassword(int(os.Stdin.Fd()))
+		out = string(inBytes)
+	} else {
+		out, err = bufio.NewReader(os.Stdin).ReadString('\n')
+	}
+	if err != nil {
+		return "", fmt.Errorf("could not read from stdin: %w", err)
+	}
+	fmt.Println()
+	return out, nil
 }
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -171,6 +172,9 @@ func homeDir() string {
 	return os.Getenv("USERPROFILE") // windows
 }
 
+// readPasswordFromStdin reads a password from stdin and returns the input
+// with trailing newline and/or carriage return removed. It also makes sure that terminal
+// echoing is turned off if stdin is a terminal.
 func readPasswordFromStdin(prompt string) (string, error) {
 	var out string
 	var err error
@@ -187,5 +191,5 @@ func readPasswordFromStdin(prompt string) (string, error) {
 		return "", fmt.Errorf("could not read from stdin: %w", err)
 	}
 	fmt.Println()
-	return out, nil
+	return strings.TrimRight(out, "\r\n"), nil
 }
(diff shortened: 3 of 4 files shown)


stefanprodan:
@makkes have you tested this on Linux, macOS and Windows?

cmd/flux/bootstrap_github.go
```diff
@@ -111,7 +111,11 @@ func init() {
 func bootstrapGitHubCmdRun(cmd *cobra.Command, args []string) error {
 	ghToken := os.Getenv(ghTokenEnvVar)
 	if ghToken == "" {
-		return fmt.Errorf("%s environment variable not found", ghTokenEnvVar)
+		var err error
+		ghToken, err = readPasswordFromStdin("Please type your GitHub personal access token: ")
```
seh, line 115: "[T]ype" sounds strange here. Perhaps "enter" instead? It's unlikely that anyone will be typing the token value.

Also, consider mentioning the common acronym "PAT."
```suggestion
		ghToken, err = readPasswordFromStdin("Please enter your GitHub personal access token (PAT): ")
```

cmd/flux/bootstrap_gitlab.go
```diff
@@ -108,7 +108,11 @@ func init() {
 func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
 	glToken := os.Getenv(glTokenEnvVar)
 	if glToken == "" {
-		return fmt.Errorf("%s environment variable not found", glTokenEnvVar)
+		var err error
+		glToken, err = readPasswordFromStdin("Please type your GitLab personal access token: ")
```
seh, line 112: Likewise.

makkes:
> @makkes have you tested this on Linux, macOS and Windows?

I tested it on Linux. @kingdonb I vaguely recall you having a Mac at hand. Would you mind taking over the macOS testing? These are the two test cases:

1. Run `echo 'asdf' | flux bootstrap github`
2. the command should just proceed using "asdf" as token. I verified that by adding a line printing the token value.

1. Run `flux bootstrap github`
2. the command should prompt for the token
3. when entering the token, no input should be echoed back to the terminal
4. the command should just proceed using the entered token. I verified that by adding a line printing the token value.

kingdonb:
I tested on MacOS, both use cases that you specified, in reverse order, first the interactive flow.

That one worked. I was able to paste my token into the prompt, press enter, and all worked as expected. The output did not include any echo of the PAT and looked like this:

```
Please enter your GitHub personal access token (PAT):
► connecting to github.com
► cloning branch "main" from Git repository "https://github.com/kingdon-ci/jenkins-infra.git"
✔ cloned repository
► generating component manifests
✔ generated component manifests
✔ committed sync manifests to "main" ("c5e8af481344cc43e83a925e7d52c58cd8c347a8")
► pushing component manifests to "https://github.com/kingdon-ci/jenkins-infra.git"
✔ reconciled components
► determining if source secret "flux-system/flux-system" exists
✔ source secret up to date
► generating sync manifests
✔ generated sync manifests
✔ sync manifests are up to date
► applying sync manifests
✔ reconciled sync configuration
◎ waiting for Kustomization "flux-system/flux-system" to be reconciled
✔ Kustomization reconciled successfully
► confirming components are healthy
✔ helm-controller: deployment ready
✔ image-automation-controller: deployment ready
✔ image-reflector-controller: deployment ready
✔ kustomize-controller: deployment ready
✔ notification-controller: deployment ready
✔ source-controller: deployment ready
✔ all components are healthy
```
✅ all as expected above.

The other example did not work for me.

Since I did not want the PAT to show up in my shell window, or shell history, I took my token in the clipboard and used `read -s MY_TOKEN` rather than the instructions you provided, then I pasted the token and pressed enter, then ran (with the binary I compiled from 04f3339):

```
echo $MY_TOKEN|~/bin/flux bootstrap github --owner kingdon-ci --repository jenkins-infra --branch=main \
  --components-extra=image-reflector-controller,image-automation-controller --path=./clusters/moo-cluster \
  --interval=40m0s --private=false
```

(the same parameters I always use here for bootstrapping)

The output included the token in it... eep, that's what I was hoping to avoid.

✗ failed to get Git repository "https://github.com/kingdon-ci/jenkins-infra": Get "https://api.github.com/repos/kingdon-ci/jenkins-infra": net/http: invalid header field value "Bearer gho_NNNNNsecrettokenMMMMM\n" for key Authorization

I then tried the same according to the specific instructions you gave, and got the same error.

I then tried again, with `echo -n` since I can recognize the newline at the end of the token is probably the one emitted by echo, that might be the source of the issue, and the result was less graphic but still a different problem:

```
 echo 'gho_NNNNNsecrettokenMMMMM' | ~/bin/flux bootstrap github --owner kingdon-ci --repository jenkins-infra --branch=main --components-extra=image-reflector-controller,image-automation-controller --path=./clusters/moo-cluster --interval=40m0s --private=false
Please enter your GitHub personal access token (PAT): ✗ could not read token: could not read from stdin: EOF
```

I looked up documentation on the ReadPassword method and I'm not sure why it does not consume the newline like it is supposed to. I guess there might be two newlines and only one is consumed, but I don't understand why, or if there is a different way I could have invoked it that would have worked. Maybe the newline is the wrong format because of OS-specific differences... maybe you need to strip out trailing newlines twice, just in case? (I hate that idea...)

But if you have any ideas for how to solve this, I'm happy to test again!

kingdonb:
Trying a few different things, with print statements...

some unicode inspectors indicate I have a CR-LF which is probably a platform-specific thing, (or maybe you get the same result on Linux?):

```
echo 'asdf' | ~/bin/flux bootstrap github --owner kingdon-ci --repository jenkins-infra --branch=main --components-extra=image-reflector-controller,image-automation-controller --path=./clusters/moo-cluster --interval=40m0s --private=false
Please enter your GitHub personal access token (PAT): from ReadString: 'asdf
'
► connecting to github.com
✗ failed to get Git repository "https://github.com/kingdon-ci/jenkins-infra": Get "https://api.github.com/repos/kingdon-ci/jenkins-infra": net/http: invalid header field value "Bearer asdf\n" for key Authorization
```

I updated `readPasswordFromStdin` for debugging:
```diff
  func readPasswordFromStdin(prompt string) (string, error) {
          var out string
          var err error
          fmt.Fprint(os.Stdout, prompt)
          stdinFD := int(os.Stdin.Fd())
          if term.IsTerminal(stdinFD) {
                  var inBytes []byte
                  inBytes, err = term.ReadPassword(int(os.Stdin.Fd()))
+                 fmt.Printf("from ReadPassword: '%s'", inBytes)
                  out = string(inBytes)
          } else {
                  out, err = bufio.NewReader(os.Stdin).ReadString('\n')
+                 fmt.Printf("from ReadString: '%s'", out)
          }
          if err != nil {
                  return "", fmt.Errorf("could not read from stdin: %w", err)
          }
          fmt.Println()
```

relu:
I've managed to test on Windows and it seems I can reproduce the same issue @kingdonb reported on Mac. Any value piped through in PowerShell is trailed by a newline char, which seems to be impossible to avert. You can replace echo with `Write-Output` or even remove it altogether and pipe the string directly, the result seems to be the same :heavy_multiplication_x:.

```
echo <token> | ./bin/flux.exe bootstrap github --owner=relu --repository=test-repo --personal=true   
Please enter your GitHub personal access token (PAT):
► connecting to github.com
✗ failed to get Git repository "https://github.com/relu/test-repo": Get "https://api.github.com/repos/relu/test-repo": net/http: invalid header field value "Bearer <token>\r\n" for key Authorization
```

Note the `\r\n` (CR + LF).

I think a solution would be to sanitize the input and remove any newline chars.

Pasting the token value into the input prompt seems to work as expected, there is no output :heavy_check_mark:.

cmd/flux/main.go
```diff
@@ -167,3 +170,22 @@ func homeDir() string {
...
+
+func readPasswordFromStdin(prompt string) (string, error) {
+	var out string
+	var err error
+	fmt.Fprint(os.Stdout, prompt)
+	stdinFD := int(os.Stdin.Fd())
+	if term.IsTerminal(stdinFD) {
+		var inBytes []byte
+		inBytes, err = term.ReadPassword(int(os.Stdin.Fd()))
+		out = string(inBytes)
```
relu, line 182: This should do it.
```suggestion
		out = strings.TrimRight(string(inBytes), "\r\n")
```

relu (approved):
Nice work, @makkes! :+1: 

Added a suggestion around handling the newline issue.

cmd/flux/main.go
```diff
@@ -167,3 +170,22 @@ func homeDir() string {
...
+	} else {
+		out, err = bufio.NewReader(os.Stdin).ReadString('\n')
```
relu, line 184: From the docs:
> ReadString reads until the first occurrence of delim in the input, returning a string containing the data up to and including the delimiter.

cmd/flux/main.go
```diff
@@ -167,3 +170,22 @@ func homeDir() string {
...
+	}
+	if err != nil {
+		return "", fmt.Errorf("could not read from stdin: %w", err)
+	}
+	fmt.Println()
+	return out, nil
```
relu, line 190:
```suggestion
	out = strings.TrimRight(out, "\r\n")
	return out, nil
```

relu (changes requested):

makkes:
Thanks @relu  and @kingdonb for testing. I'll spend some time on this PR later today.

makkes:
@kingdonb @relu pushed a new commit (had to force-push after rebasing on top of main) that should fix all the problems you mentioned. Would you mind testing again on macOS and Windows, please? 🙏🏻

relu (approved):

kingdonb:
Tested on MacOS, there is no echo for either method (`echo MY_ACCESS_TOKEN|flux bootstrap github ...` or interactive entry method). LGTM 👍

stefanprodan (approved):
LGTM

Thanks @makkes 🏅

kingdonb (commented):
LGTM

Issue #2040, opened 2021-11-02
GitHub bootstrap: You're using an RSA key with SHA-1, which is no longer allowed.


stefanprodan:
The default SSH deploy key setting used by `flux bootstrap` no longer works with GitHub.

Relevant changes: https://github.blog/2021-09-01-improving-git-protocol-security-github/

HeroCC:
I started getting this error, what is the recommended way to resolve the error? Is re-running the entire bootstrap safe, or is there a way to force it to just regenerate the SSH key?

tellisnz-shift:
> I started getting this error, what is the recommended way to resolve the error? Is re-running the entire bootstrap safe, or is there a way to force it to just regenerate the SSH key?

@HeroCC @AnkurJain10 @dexoid - as per https://cloud-native.slack.com/archives/CLAJ40HV3/p1641866183079500?thread_ts=1641866131.079300&cid=CLAJ40HV3 - upgrade flux, delete the flux-system secret, and re bootstrap worked for me and seemed non detructive

or-shachar:
What should we do if we don't use the flux bootstrap command line but the Terrraform provider?

Seems like something has changed today:
![image](https://user-images.githubusercontent.com/2962963/148966307-50add1ec-3f2f-467d-ab0f-0fa076ab9504.png)

The thing is - the `known_hosts` field in the secret was updated months ago to the right value:
```
github.com ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBEmKSENjQEezOmxkZMy7opKgwFB9nkt5YRrYMjNuG5N87uRgg6CLrbo5wAdT/y6v0mKV0U2w0WZ2YB/++Tpockg=
```

Maybe we should be using the `Ed25519` key?

snahelou:
With Terraform provider @or-shachar  you just have to create a new key 
`
resource "tls_private_key" "main" {
  algorithm   = "ECDSA"
  ecdsa_curve = "P256"
}
`

instead of RSA. Apply and modify git repo if not managed by TF

or-shachar:
Was just about to post that the problem was not in the known_hosts but with our deploy key.
The solution it to replace: 
```terraform
resource "tls_private_key" "main" {
  algorithm   = "RSA"
  ecdsa_curve = "4096"
}
```
with:
```terraform
resource "tls_private_key" "main" {
[...]

PR #2041, opened 2021-11-02
bootstrap: Set ECDSA as the default SSH key algorithm

stefanprodan:
Motivation: RSA SHA-1 SSH keys are no longer accepted by GitHub https://github.blog/2021-09-01-improving-git-protocol-security-github/.

Given this we are switching the default from RSA to ECDSA for `git`, `github` and `gitlab` variants of `flux bootstrap`.

Fix: #2040

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -140,7 +140,7 @@ func NewBootstrapFlags() bootstrapFlags {
 	return bootstrapFlags{
 		logLevel:           flags.LogLevel(rootArgs.defaults.LogLevel),
 		requiredComponents: []string{"source-controller", "kustomize-controller"},
-		keyAlgorithm:       flags.PublicKeyAlgorithm(sourcesecret.RSAPrivateKeyAlgorithm),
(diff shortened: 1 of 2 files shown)


hiddeco (approved):

hiddeco:
NB: switching to the RSA-SHA2 alternative is not possible, as Go lacks support for this format.

tarasbsbr:
@stefanprodan Just curious, why ECDSA? github seems to [recommend](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) ed25519, others do [as well](https://goteleport.com/blog/comparing-ssh-keys/)?

antoinedeschenes:
ECDSA has trust issues all over, Ed25519 could have been a better choice here. 

RSA keys aren't going away, just usage of SHA-1 signatures during authentication, which affects older clients. 

However as @hiddeco mentioned, looks like golang doesn't handle it yet, even though it's been part of OpenSSH for a while. https://github.com/golang/go/issues/37278

hiddeco:
Due to various parts of Flux depending on `libgit2`, the recommendation is different: https://github.blog/2021-09-01-improving-git-protocol-security-github/#libgit2-and-other-git-clients

PR #2046, opened 2021-11-03
Use full domain name for notification-controller

vespian:
We need to use full domain name for http urls as some users may deploy flux in environments that use http proxy to access internet.

In such environments they normally add `.cluster.local` and `.local` suffixes to `no_proxy` variable in order to prevent cluster-local traffic from going through http proxy. Without fully specified domain they need to mention `notifications-controller`  explicitly in `no_proxy` variable after debugging http proxy logs.

### DIFF
--- a/pkg/manifestgen/install/manifests.go
+++ b/pkg/manifestgen/install/manifests.go
@@ -65,7 +65,15 @@ func fetch(ctx context.Context, url, version, dir string) error {
 
 func generate(base string, options Options) error {
 	if containsItemString(options.Components, options.NotificationController) {
-		options.EventsAddr = fmt.Sprintf("http://%s/", options.NotificationController)
+		// We need to use full domain name here, as some users may deploy flux
+		// in environments that use http proxy.
+		//
+		// In such environments they normally add `.cluster.local` and `.local`
+		// suffixes to `no_proxy` variable in order to prevent cluster-local
+		// traffic from going through http proxy. Without fully specified
+		// domain they need to mention `notifications-controller` explicity in
+		// `no_proxy` variable after debugging http proxy logs.
+		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.%s/", options.NotificationController, options.Namespace, options.ClusterDomain)
 	}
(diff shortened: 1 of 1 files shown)


pkg/manifestgen/install/manifests.go
```diff
@@ -65,7 +65,15 @@ func fetch(ctx context.Context, url, version, dir string) error {
...
-		options.EventsAddr = fmt.Sprintf("http://%s/", options.NotificationController)
+		// We need to use full domain name here, as some users may deploy flux
+		// in environments that use http proxy.
+		//
+		// In such environments they normally add `.cluster.local` and `.local`
+		// suffixes to `no_proxy` variable in order to prevent cluster-local
+		// traffic from going through http proxy. Without fully specified
+		// domain they need to mention `notifications-controller` explicity in
+		// `no_proxy` variable after debugging http proxy logs.
+		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.cluster.local/", options.NotificationController, options.Namespace)
```
hiddeco, line 76: Instead of assuming this hardcoded value is the right thing for every setup, I think it would be better to allow modification of the address suffix (`.svc.cluster.local`).

vespian: What use-cases do you forsee?

stefanprodan: Hardcoding the cluster domain is not acceptable, the domain should be taken from https://github.com/fluxcd/flux2/blob/main/pkg/manifestgen/install/options.go#L37

stefanprodan: > What use-cases do you forsee?

May people change the cluster domain, we can't assume that `.local` works on any cluster.

vespian: > May people change the cluster domain, we can't assume that .local works on any cluster.

Makes sense. Please have another look.

stefanprodan (approved):
LGTM

Thanks @vespian

makkes (approved):

Issue #2049, opened 2021-11-04
Add a note to YAML files which should not be changed manually


rombert:
### Describe the bug

Certain files created by `flux bootstrap` will be overwritten if manually changed. This is not immediately obvious and can lead to confusing situations.

I suggest that these files have a comment that explains this, something along the line of

```yaml
# Do not make changes to this file, it will be overwritten if you run flux bootstrap again 
```

### Steps to reproduce

1. Install Flux (`flux bootstrap`)
2. Change the `gotk-sync.yaml` file
3. Run Flux bootstrap again

### Expected behavior

I should be made aware of the fact that the file will be overwritten when opening it.

See also https://github.com/fluxcd/flux2/discussions/1967 .

### OS / Distro

openSUSE Tumbleweed

### Flux version

0.19.0

### Flux check

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
I propose a shorter comment (similar to what controller-gen does):

```
# This manifest was generated by flux bootstrap. DO NOT EDIT.
```

rombert:
Awesome, thanks for the quick turnaround!

PR #2050, opened 2021-11-04
Add `DO NOT EDIT` warn to bootstrap sync manifests

stefanprodan:
Fix: #2049


hiddeco (commented):
Should we maybe provide a hint and link people to the customize documentation?

pkg/manifestgen/manifest.go
```diff
@@ -24,6 +24,8 @@ import (
 	securejoin "github.com/cyphar/filepath-securejoin"
 )
 
+const GenWarning = "# This manifest was generated by flux bootstrap. DO NOT EDIT."
+
```
hiddeco, line 28:
```suggestion
const GenWarning = `# This manifest was generated by flux bootstrap. DO NOT EDIT.
# For modifications to the Flux configuration, see the "customize Flux" documentation.
# Ref: https://fluxcd.io/docs/installation/#customize-flux-manifests
`

```

stefanprodan: What about Terraform? this comment ends up everywhere manifestgen is used

stefanprodan: I think we need to have a dedicated page for "Customize Flux manifests" where we address both the CLI and TF bootstrap options. After that we could update this comment and point to the new page.

hiddeco: We should record this in a follow-up issue.

hiddeco (approved):

PR #2062, opened 2021-11-08
Edit grafana dashboard to be able to filter by namespace

drazul:
Add new variable to filter by exported namespace.

Edit definition of namespace variable to use grafana custom promql function `label_values`.
Rename variable namespace to operator_namespace.
Rename variable exported_namespace to namespace

Signed-off-by: Daniel AguadoAraujo <daniel.aguadoaraujo@gfk.com>

### DIFF
--- a/manifests/monitoring/grafana/dashboards/cluster.json
+++ b/manifests/monitoring/grafana/dashboards/cluster.json
@@ -15,7 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1602679512025,
+  "iteration": 1636369574387,
   "links": [],
   "panels": [
     {
@@ -23,7 +23,6 @@
       "description": "",
       "fieldConfig": {
         "defaults": {
-          "custom": {},
           "decimals": 0,
           "mappings": [],
           "thresholds": {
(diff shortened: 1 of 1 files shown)


dminca (approved):
lgtm 👍

stefanprodan:
> Rename variable namespace to operator_namespace.
> Rename variable exported_namespace to namespace

I understand the changes but not the reason for them, what are you trying to achieve here?

drazul:
> > Rename variable namespace to operator_namespace.
> > Rename variable exported_namespace to namespace
> 
> I understand the changes but not the reason for them, what are you trying to achieve here?

In this context the operator namespace only matter if we have more than one operator running in the same cluster, and the name 'exported_namespace' has meaning only because namespace variable name is already in use.

In my opinion renaming both variables add a better understanding about the dashboard to not technical people.

But I can keep both variables as they are written in the prometheus metric if you want.

stefanprodan:
> In my opinion renaming both variables add a better understanding about the dashboard to not technical people.

Thanks, now I get it.

stefanprodan (approved):
LGTM

Thanks @drazul can you please rebase with upstream main and force push.

drazul:
> LGTM
> 
> Thanks @drazul can you please rebase with upstream main and force push.

Done

Issue #2056, opened 2021-11-04
Inconsistency in manifests generated by install vs bootstrap commands


aliusmiles:
### Describe the bug

Manifests generated by `flux install --export` have extra lines compared to those from `flux bootstrap`:

```sh
--- a/clusters/dev/flux-system/gotk-components.yaml
+++ b/clusters/dev/flux-system/gotk-components.yaml
@@ -1,3 +1,6 @@
+---
+# Flux version: v0.21.1
+# Components: source-controller,kustomize-controller,helm-controller,notification-controller,image-reflector-controller,image-automation-controller
 apiVersion: v1
 kind: Namespace
 metadata:
@@ -6228,3 +6231,4 @@ spec:
       app: notification-controller
   policyTypes:
   - Ingress
+---
```

It's nothing critical but rather annoying when cluster update is always done by `flux install --export` (I like this way since I can preview changes plus it can be automated) and then for some reason, like key recreation in v0.21.0, `flux bootstrap` must be executed.

### Steps to reproduce

1. bootstrap using `flux bootstrap ...` command
2. go to repo/clusters/name/flux-system and run `flux install --export ... > gotk-components.yaml`
3. run `git diff`

### Expected behavior

No differences in manifests.

### OS / Distro

N/A

### Flux version

v0.21.1

### Flux check

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jack-evans:
Agreed, I think if there was a consistent message at the top for both the install and the bootstrap then that would be a lot better....

maybe something like this in both cases:
```yaml
# This manifest was generated by flux. DO NOT EDIT.
# Flux version: v0.21.1
# Components: source-controller,kustomize-controller,helm-controller,notification-controller,image-reflector-controller,image-automation-controller
```

jack-evans:
happy to put in a PR for this if we think the above is an improvement?

stefanprodan:
@jack-evans your proposal looks good to me, I'm not sure how would this would work with Flux Terraform provider as the version is a property of the binary.

jack-evans:
so my thoughts were it would live in the flux2 command itself, on install (with `--export`) and bootstrap (when pushing to git/vcs) we could have something that prepends:
```yaml
# This manifest was generated by flux. DO NOT EDIT.
# Flux version: v0.21.1
# Components: source-controller,kustomize-controller,helm-controller,notification-controller,image-reflector-controller,image-automation-controller
```
[...]

PR #2066, opened 2021-11-08
Make flux bootstrap and install comments consistent

jack-evans:
Signed-off-by: Jack Evans <jack.evans1@ibm.com>

Fix: #2056

Changes:
- removed bootstrap from GenWarning constant as to make it more generic
- remove custom addition of the prints in `flux install --export`
- Created new function GetGenWarning that generates a consistent generation header warning for both bootstrap and install commands

First PR so please let me know if there is anything a miss 🙂


pkg/manifestgen/install/install_test.go
```diff
@@ -41,5 +41,10 @@ func TestGenerate(t *testing.T) {
 		t.Errorf("toleration key '%s' not found", opts.TolerationKeys[0])
 	}
 
+	warning := GetGenWarning(opts)
+	if !strings.Contains(output.Content, GetGenWarning(opts)) {
```
hiddeco, line 45: Given we know the position it should be in (at the start), I think it would be better to do:

```suggestion
	if !strings.HasPrefix(output.Content, GetGenWarning(opts)) {
```

jack-evans: Ahh yep good call! Brain always immediately goes for `strings.Contains’ 😄

jack-evans:
Will squash commits

pkg/manifestgen/install/install_test.go
```diff
@@ -41,5 +41,10 @@ func TestGenerate(t *testing.T) {
 		t.Errorf("toleration key '%s' not found", opts.TolerationKeys[0])
 	}
 
+	warning := GetGenWarning(opts)
+	if !strings.HasPrefix(output.Content, GetGenWarning(opts)) {
+		t.Errorf("Generation warning '%s' not found", warning)
+	}
```
hiddeco, line 47:
```suggestion
	warning := GetGenWarning(opts)
	if !strings.HasPrefix(output.Content, warning) {
		t.Errorf("Generation warning '%s' not found", warning)
	}
```

jack-evans: ahh yep another one that skipped straight through 🤦🏻

hiddeco (approved):
Thank you @jack-evans, and :hand: on first PR! :sunflower:

Issue #2063, opened 2021-11-08
Skip remote Kustomizations in `tree kustomization` command


makkes:
When running the `tree kustomization` command I want all Kustomizations that have the `.spec.kubeConfig` field set to be skipped so that all other Kustomizations are processed properly.

### Background

If any Kustomization on the path of the `tree kustomization` command has the `.spec.kubeConfig` field set, the output looks similar to this:

```
 $ flux tree kustomization control-plane
✗ failed to find object: kustomizations.kustomize.toolkit.fluxcd.io "workspace" not found
```

### Discussion

https://cloud-native.slack.com/archives/CLAJ40HV3/p1636380087272400

PR #2071, opened 2021-11-09
Skip remote clusters in flux tree cmd

stefanprodan:
Given that Flux Kustomizations and HelmReleases can target remote clusters, it is not possible to discover the reconciled resources by just querying the management cluster API. In the future we may want to consider fetching the `kubeConfig` and connect to the remote clusters, but for now we'll just skip the discovery for those Flux CRs.

Fix: #2063

### DIFF
--- a/cmd/flux/tree_kustomization.go
+++ b/cmd/flux/tree_kustomization.go
@@ -169,7 +169,9 @@ func treeKustomization(ctx context.Context, tree tree.ObjMetadataTree, item *kus
 		}
 
 		if objMetadata.GroupKind.Group == kustomizev1.GroupVersion.Group &&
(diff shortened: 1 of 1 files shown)


hiddeco (commented):
Just adding a comment here so that it is recorded somewhere publicly: I think that from a UX perspective, it would be nice to know as a user that the further tree is omitted with a reason (remote cluster being targeted) instead of things just "magically hiding".

Based on a chat with @stefanprodan however, it seems to be a side effect of the strong typed `objMetadataTree`. Changing this would be better to do in a separate PR, or while we add things like the Status of the object, which would require us to revise the structs anyway.

makkes (approved):
👏🏻

PR #2075, opened 2021-11-11
Remove trailing `---` for `flux install` to match `flux bootstrap` generated YAML

jack-evans:
Signed-off-by: Jack Evans <jack.evans1@ibm.com>

fix: #2056 

- remove trailing `---` that I missed in the first PR 😢 
- flux bootstrap and flux install commands now generate the exact same thing


stefanprodan (approved):
LGTM

Thanks @jack-evans

PR #2091, opened 2021-11-16
Update Alpine to v3.14

vespian:
Followup for https://github.com/fluxcd/helm-controller/pull/360.

Unify Alpine minor-version used by container by bumping to 3.14.


stefanprodan (approved):
LGTM

Thanks @vespian for doing the Alpine upgrade across all Flux projects 🏅

PR #2092, opened 2021-11-16
[RFC] Access control for cross-namespace source refs

stefanprodan:
This RFC proposes a way to share sources between namespaces when running Flux on multi-tenant clusters.

Supersedes #582

### DIFF
--- a/rfcs/0002-source-acl/README.md
+++ b/rfcs/0002-source-acl/README.md
@@ -1,0 +1,157 @@
+# RFC-0002 Access control for source references
+
+## Summary
+
+Cross-namespace references to Flux sources should be subject to
+Access Control Lists (ACLs) as defined by the owner of a particular source.
+
+## Motivation
+
+As of v0.23.0, Flux allows for `Kustomizations` and `HelmReleases` to reference sources in different namespaces.
+This poses a serious security risk for multi-tenant environments as Flux does not prevent tenants from accessing
+known sources outside of their namespaces.
+
+Flux does not allow for an `ImageUpdateAutomation` to reference a `GitRepository` in a different namespace.
+This means users have to copy the Git auth secret and the `GitRepository` object in all namespaces
+where `ImageUpdateAutomations` are used. This poses a serious security risk for multi-tenant environments,
+as tenants could use the Git secret to push changes to repositories by impersonating Flux.
+
+Flux allows for `ImagePolicies` to reference `ImageRepositories` in a different namespace only 
+if the ACL present on the `ImageRepository` grants access to the namespace where the `ImagePolicy` is.
+This has been implemented in
+[fluxcd/image-reflector-controller#162](https://github.com/fluxcd/image-reflector-controller/pull/162).
+
+Flux should be consistent when dealing with cross-namespace references by extending the
+Image Policy/Repository approach to all the other APIs.
+
+### Goals
+
+- Allow source owners to choose which sources are shared and with which namespaces.
+- Allow cluster admins to enforce source ACLs.
+
+### Non-Goals
+
+- Enforce source ACLs by default. 
+
+## Proposal
+
+Extend the current Image Policy/Repository ACL implementation to all the others Flux resources
+as described in [flux2#1704](https://github.com/fluxcd/flux2/issues/1704).
+
+When a Flux resource (`Kustomization`, `HelmRelease` or `ImageUpdateAutomation`)
+refers to a source (`GitRepository`, `HelmRepository` or `Bucket`) in a different namespace,
+access is granted based on the source ACL.
+
+The ACL check is performed only if `--enable-source-acl` flag is set to `true` for the following controllers:
+
+- kustomize-controller
+- helm-controller
+- image-automation-controller
+
+### User Stories
+
+#### Story 1
+
+> As a cluster admin, I want to share Helm Repositories approved by the platform team with all tenants.
+
+If the owner of a Flux `HelmRepository` wants to grant access to the repository for all namespaces in a cluster,
+an empty `matchLabels` can be used:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: HelmRepository
+metadata:
+  name: bitnami
+  namespace: flux-system
+spec:
+  url: https://charts.bitnami.com/bitnami
+  accessFrom:
+    namespaceSelectors:
+      - matchLabels: {}
+```
+
+If the `accessFrom` field is not present and `--enable-source-acl` is set to `true`,
+means that a source can't be accessed from any other namespace but the one where it currently resides.
+
+#### Story 2
+
+> As a tenant, I want to share my app repository with another tenant
+> so that they can deploy the application in their own namespace.
+
+If `dev-team1` wants to grant read access to their repository to `dev-team2`,
+a `matchLabels` that selects the namespace owned by `dev-team2` can be used:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: GitRepository
+metadata:
+  name: app1
+  namespace: dev-team1
+spec:
+  url: ssh://git@github.com/<org>/app1-deploy
+  secretRef:
+    name: app1-ro-ssh-key
+  accessFrom:
+    namespaceSelectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: dev-team2
+```
+
+#### Story 3
+
+> As a cluster admin, I want to let tenants configure image automation in their namespaces by
+> referring to a Git repository managed by the platform team.
+
+If the owner of a Flux `GitRepository` wants to grant write access to `ImageUpdateAutomations` in a different namespace,
+a `matchLabels` that selects the image automation namespace can be used:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: GitRepository
+metadata:
+  name: cluster-config
+  namespace: flux-system
+spec:
+  url: ssh://git@github.com/<org>/cluster-config
+  secretRef:
+    name: read-write-ssh-key
+  accessFrom:
+    namespaceSelectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: dev-team1
+```
+
+The `dev-team1` can refer to the `cluster-config` repository in their image automation config:
+
+```yaml
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImageUpdateAutomation
+metadata:
+  name: app1
+  namespace: dev-team1
+spec:
+  sourceRef:
+    kind: GitRepository
+    name: cluster-config
+    namespace: flux-system
+```
+
+### Alternatives
+
+An alternative solution to source ACLs is showcased in the current multi-tenancy example, where an
+admission controller such as Kyverno or OPA Gatekeeper is used to block cross-namespace access to sources.
+
+Another alternative is to rely on impersonation and create a `ClusterRoleBinding` per named source and tenant account
+as described in [fluxcd/flux2#582](https://github.com/fluxcd/flux2/pull/582). 
+
+Yes another alternative is to introduce a new API kind `SourceReflection` as described in
+[fluxcd/flux2#582-821027543](https://github.com/fluxcd/flux2/pull/582#issuecomment-821027543).
+
+And finally, this is more of an improvement of the current proposal, is for source-controller to compile the ACLs
+to Kubernetes RBAC and dynamically create `ClusterRoleBindings` for all tenants accounts
+every time a source or namespace changes as described in
+[fluxcd/flux2#582-821388906](https://github.com/fluxcd/flux2/pull/582#issuecomment-821388906).
+
+## Implementation History
+
+- ACL support for allowing cross-namespace access to `ImageRepositories` was first released in flux2 **v0.23.0**.
--- a/rfcs/0002-source-acl/README.md
+++ b/rfcs/0002-source-acl/README.md
@@ -1,5 +1,11 @@
 # RFC-0002 Access control for source references
 
+**Status:** provisional
+
+**Creation date:** 2021-11-16
+
+**Last update:** 2022-02-03
+
 ## Summary
 
 Cross-namespace references to Flux sources should be subject to
@@ -148,7 +154,7 @@ Another alternative is to rely on impersonation and create a `ClusterRoleBinding
 as described in [fluxcd/flux2#582](https://github.com/fluxcd/flux2/pull/582). 
 
 The current proposal is more flexible than RBAC and implies less work for Flux users. ALCs act more like
-Kubernetes Network Policies where access is define based on labels, with RBAC every time a namespace is added,
+Kubernetes Network Policies where access is defined based on labels, with RBAC every time a namespace is added,
 the platform admins have to create new RBAC rules to target that namespace.
 
 #### Source reflection CRD
@@ -172,3 +178,4 @@ each namespace that uses the same Git or Helm repository due to the requirement 
 ## Implementation History
 
 - ACL support for allowing cross-namespace access to `ImageRepositories` was first released in flux2 **v0.23.0**.
+- Disabling cross-namespace access to sources was first released in flux2 **v0.26.0**.
--- a/rfcs/0002-source-acl/README.md
+++ b/rfcs/0002-source-acl/README.md
@@ -11,26 +11,14 @@ caller's namespace.
 
 ## Motivation
 
+As of [version 0.26](https://github.com/fluxcd/flux2/releases/tag/v0.26.0) (Feb 2022),
+Flux allows for `Kustomizations`, `HelmReleases` and `ImageUpdateAutomations` to reference sources in different namespaces.
+On multi-tenant clusters, platform admins can disable this behaviour with the `--no-cross-namespace-refs` flag
+as described in the [multi-tenancy lockdown documentation](https://fluxcd.io/docs/installation/#multi-tenancy-lockdown).
+
 This proposal tries to solve the "cross-namespace references side-step namespace isolation" issue (explained in
-[RFC-0001](https://github.com/fluxcd/flux2/tree/main/rfcs/0001-authorization#cross-namespace-references-side-step-namespace-isolation)).
-
-As of [version 0.25](https://github.com/fluxcd/flux2/releases/tag/v0.25.0) (Ian 2022),
-Flux allows for `Kustomizations` and `HelmReleases` to reference sources in different namespaces.
-This poses a serious security risk for multi-tenant environments as Flux does not prevent tenants from accessing
-known sources outside of their namespaces.
-
-Flux does not allow for an `ImageUpdateAutomation` to reference a `GitRepository` in a different namespace.
-This means users have to copy the Git auth secret and the `GitRepository` object in all namespaces
-where `ImageUpdateAutomations` are used. This poses a serious security risk for multi-tenant environments,
-as tenants could use the Git secret to push changes to repositories by impersonating Flux.
-
-Flux allows for `ImagePolicies` to reference `ImageRepositories` in a different namespace only 
-if the ACL present on the `ImageRepository` grants access to the namespace where the `ImagePolicy` is.
-This has been implemented in
-[fluxcd/image-reflector-controller#162](https://github.com/fluxcd/image-reflector-controller/pull/162).
-
-Flux should be consistent when dealing with cross-namespace references by extending the
-Image Policy/Repository approach to all the other APIs.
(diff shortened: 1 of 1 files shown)


rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,143 @@
...
+  sourceRef:
+    kind: GitRepository
+    name: cluster-config
+    namespace: flux-system
+```
+
+### Alternatives
+
+An alternative solution to source ACLs is showcased in the current multi-tenancy example, where an
+admission controller such as Kyverno or OPA Gatekeeper is used to block cross-namespace access to sources.
```
hiddeco, line 139: Have we considered alternative designs within our own API?

vterdunov:
@stefanprodan I'd like to say the problem that I see in the ACL proposal. As a tenant somehow I need to know each `GitRepository` name+namespace allowed to use by me. It's a bit implicit for me.
On the other hand, the `SourceReflection` looks more explicitly, however, I don't know how to properly prevent creating SourceReflections resources by one tenant in another tenant namespace.

Another implicit behavior for me is an implicit reference from `ImageUpdateAutomation` to `ImagePolicy`.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,154 @@
...
+### Non-Goals
+
+- Enforce source ACLs by default. 
+
+## Proposal
+
+Extend the current Image Policy/Repository ACL implementation to all the others Flux resources
+as described in [flux2#1704](https://github.com/fluxcd/flux2/issues/1704).
+
+When a Flux resource (`Kustomization`, `HelmRelease` or `ImageUpdateAutomation`)
```
jonathan-innis, line 41: How is this mechanism extended for things like Receivers, which have Resources that can refer to any resource across the cluster?

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+The ACL check is performed only if `--enable-source-acl` flag is set to `true` for the following controllers:
+
+- kustomize-controller
+- helm-controller
+- image-automation-controller
+
+### User Stories
+
+#### Story 1
```
jonathan-innis, line 53: My assumption here is that if `--enable-source-acl` is enabled and no `accessFrom` is listed, that the default behavior is that you will not be able to access it from any other namespace but the one that it currently resides. Can this be explicitly called out?

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,154 @@
...
+An alternative solution to source ACLs is showcased in the current multi-tenancy example, where an
+admission controller such as Kyverno or OPA Gatekeeper is used to block cross-namespace access to sources.
+
+Another alternative is to rely on impersonation and create a `ClusterRoleBinding` per named source and tenant account
+as described in [fluxcd/flux2#582](https://github.com/fluxcd/flux2/pull/582). 
+
+Yes another alternative is to introduce a new API kind `SourceReflection` as described in
+[fluxcd/flux2#582-821027543](https://github.com/fluxcd/flux2/pull/582#issuecomment-821027543).
+
+And finally, this is more of an improvement of the current proposal, is for source-controller to compile the ACLs
```
jonathan-innis, line 147: The current implementation of ACLs is not described in detail in this current doc, so it is a bit unclear to me what the current gap is between improving the current implementation and the implementation described in this paragraph.

I understand that the discussion might be listed elsewhere, but it would be good in one of these RFCs to explicitly lay out what the controller does and what service account and permissions it runs under when attempting to check the ACL

jonathan-innis:
I think the primary issue I see with ACLs is understanding what is available to you in terms of sources. As much as I like the idea of having some sort of CLI support for grabbing the available sources, I cannot think of a good way to show the available sources in a secure way to the tenant user

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,154 @@
...
+      - matchLabels:
+          kubernetes.io/metadata.name: dev-team2
+```
+
+#### Story 3
+
+> As a cluster admin, I want to let tenants configure image automation in their namespaces by
+> referring to a Git repository managed by the platform team.
+
+If the owner of a Flux `GitRepository` wants to grant write access to `ImageUpdateAutomations` in a different namespace,
```
souleb, line 102: I think it would be good to lay out the list of possible access... If as a user I grant access to another namespace, depending on the running controllers on that NS I might re-consider.

rfcs/0002-source-acl/README.md
stefanprodan: This RFC is only about sharing sources to reconcilers (either in-cluster or in-git syncs). For the notification API we'll need a different mechanism, my proposal will be to deny cross-namespace refs from Receivers and Alerts in RFC-0003.

rfcs/0002-source-acl/README.md
stefanprodan: > Can this be explicitly called out?

Added, thanks!

stefanprodan:
> I think the primary issue I see with ACLs is understanding what is available to you in terms of sources. As much as I like the idea of having some sort of CLI support for grabbing the available sources, I cannot think of a good way to show the available sources in a secure way to the tenant user

ACLs are very much like network polices. As far as know, there is no way for a tenant to query Kubernetes API and get a list of endpoints that allow ingress traffic from its namespace.

rfcs/0002-source-acl/README.md
stefanprodan: @jonathan-innis I've added the notification API restrictions to RFC-0003. Thanks for raising this.

rfcs/0002-source-acl/README.md
stefanprodan: The level of access to the underlying source depends on how you've configured the SSH key or token, if it has write access to just read. `ImageUpdateAutomations` need write access to patch image tags, while a `HelmRelease` or `Kustomization` needs only read access.

jonathan-innis (approved):
+1 from Microsoft 🚀

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,157 @@
...
+  name: cluster-config
+  namespace: flux-system
+spec:
+  url: ssh://git@github.com/<org>/cluster-config
+  secretRef:
+    name: read-write-ssh-key
+  accessFrom:
+    namespaceSelectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: dev-team1
```
pjbgf, line 121: This pattern is very flexible for resource discovery and I can easily understand the benefits of using it in this use case. The main challenge I can think of using this as part of flux's authorisation model in addition to the built-in RBAC is that a cluster admin who is unaware of flux's inner workings may be blindsided by this behaviour, leading to less secure flux deployments. 

Given that RBAC will stop being the single source of truth from an authorisation perspective, it may be useful to also provide a sub-command on `flux` to allow admins to inspect effective permissions across a cluster.

pjbgf:
> I think the primary issue I see with ACLs is understanding what is available to you in terms of sources. As much as I like the idea of having some sort of CLI support for grabbing the available sources, I cannot think of a good way to show the available sources in a secure way to the tenant user

A way to achieve this could be by a platform admin leveraging the built-in RBAC, having a pre-agreed namespace (say `flux-shared`) and provide `list` permissions to all sources that are available to a given tenant. Effectively allowing them to query them via `kubectl`. But it would be the platform admin's responsibility to ensure both RBAC and ACL permissions are in-sync.

pjbgf (commented):
A potential alternative approach to rely solely on the built-in RBAC instead of creating a new authorisation mechanism, is to leverage the `use` verb. Below is a simple example that would provide flux's tenant the ability to use all sources from namespace `flux-shared`. It also gives `specific-tenant-user` the ability to query (list/get) all such resources:

```yaml
kind: ClusterRole
metadata:
  name: flux-use-sources
rules:
- apiGroups: ['source.toolkit.fluxcd.io']
  resources:
  - bucket
  - gitrepository
  - helmchart
  - helmrepository
  verbs:
  - use
  - list
  - get
---
kind: RoleBinding
metadata:
  name: flux-shared-resources
  namespace: flux-shared
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: flux-use-sources
subjects:
- kind: ServiceAccount
  name: flux
  namespace: specific-tenant
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: specific-tenant-user
```

It is not as flexible as the proposed solution, but it requires less effort to implement and delays the need to change flux's API. The main impact being that users won't be able to use selectors, but instead will have to create `RoleBinding` and `ClusterRoleBinding` objects to enable the use of flux sources. 

The main benefit being that the built-in RBAC would still be the single point of truth for authorisation of flux objects.

jhaumont:
Hello
Do you have plan to do the same thing between `ImagePoliciy` and `ImageUpdateAutomation` ressources ?
An example:
```
apiVersion: image.toolkit.fluxcd.io/v1beta1
kind: ImagePolicy
metadata:
  name: podinfo
  namespace: flux-system
spec:
  imageRepositoryRef:
    name: podinfo
  policy:
    semver:
      range: 5.0.x
  accessFrom:
    namespaceSelectors:
      - matchLabels: 
           kubernetes.io/metadata.name: flux-system
```
or maybe I misunderstood how the update automation controller works ^^

stefanprodan:
> Do you have plan to do the same thing between ImagePoliciy and ImageUpdateAutomation ressources ?

Yes, this RFC is for the source-controller APIs, once we settled on a solution, I’ll create a new RFC for image automation.

jhaumont (approved):

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,165 @@
...
+  name: app1
+  namespace: dev-team1
+spec:
+  sourceRef:
+    kind: GitRepository
+    name: cluster-config
+    namespace: flux-system
+```
+
+### Alternatives
```
squaremo, line 147: The purpose of this section is not just to mention alternatives, but to explain why the proposed solution is better than each of them.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+An alternative solution to source ACLs is showcased in the current multi-tenancy example, where an
+admission controller such as Kyverno or OPA Gatekeeper is used to block cross-namespace access to sources.
+
+Another alternative is to rely on impersonation and create a `ClusterRoleBinding` per named source and tenant account
+as described in [fluxcd/flux2#582](https://github.com/fluxcd/flux2/pull/582). 
+
+Yes another alternative is to introduce a new API kind `SourceReflection` as described in
```
hiddeco, line 155:
```suggestion
Yet another alternative is to introduce a new API kind `SourceReflection` as described in
```

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,165 @@
...
+Access Control Lists (ACLs) as defined by the owner of a particular source.
+
+Similar to [Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/),
+Flux ACLs define policies for restricting the access to the source artifact server based on the
+caller's namespace.
+
+## Motivation
+
+This proposal tries to solve the "cross-namespace references side-step namespace isolation" issue (explained in
+[RFC-0001](https://github.com/fluxcd/flux2/tree/main/rfcs/0001-authorization#cross-namespace-references-side-step-namespace-isolation)).
```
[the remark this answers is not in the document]
stefanprodan: Added to alternatives.

rfcs/0002-source-acl/README.md
stefanprodan: Added please check.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,185 @@
...
+sources.
+
+#### Kubernetes RBAC
+
+Another alternative is to rely on impersonation and create a `ClusterRoleBinding` per named source and tenant account
+as described in [fluxcd/flux2#582](https://github.com/fluxcd/flux2/pull/582). 
+
+The current proposal is more flexible than RBAC and implies less work for Flux users. ALCs act more like
+Kubernetes Network Policies where access is define based on labels, with RBAC every time a namespace is added,
+the platform admins have to create new RBAC to target that namespace.
```
hiddeco, line 164:
```suggestion
the platform admins have to create new RBAC rules to target that namespace.
```

or "manifests"?

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+### Alternatives
+
+#### Admission controllers
+
+An alternative solution to source ACLs is showcased in the current multi-tenancy example, where an
+admission controller such as Kyverno or OPA Gatekeeper is used to block cross-namespace access to sources.
+
+The current proposal offers the same feature but without the need to manager yet another controller to guard
```
hiddeco, line 154:
```suggestion
The current proposal offers the same feature but without the need to manage yet another controller to guard
```

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,185 @@
...
+of creating objects in namespaces where it has no control over.
+
+#### Remove cross-namespace refs
+
+An alternative is to simply remove cross-namespace references from the Flux API.
+
+I personally find this unacceptable as it would break the current behavior and people would 
+have to make substantial changes to their repo structure and workflow. In some cases, this would
+mean source-controller would use way more memory and network bandwidth that grows with
+each namespace that uses the same Git or Helm repository.
```
hiddeco, line 181:
```suggestion
This would break with current behavior, and users would have to make substantial changes to their
repository structure and workflow. In cases where e.g. a resource is common (across many namespaces),
this would mean the source-controller would use way more memory and network bandwidth that grows with
each namespace that uses the same Git or Helm repository due to the requirement of having to duplicate
"common" resources.
```

stefanprodan: Thank you!

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+Similar to [Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/),
+Flux ACLs define policies for restricting the access to the source artifact server based on the
+caller's namespace.
+
+## Motivation
+
+This proposal tries to solve the "cross-namespace references side-step namespace isolation" issue (explained in
+[RFC-0001](https://github.com/fluxcd/flux2/tree/main/rfcs/0001-authorization#cross-namespace-references-side-step-namespace-isolation)).
+
+As of [version 0.25](https://github.com/fluxcd/flux2/releases/tag/v0.25.0) (Ian 2022),
```
makkes, line 17:
```suggestion
As of [version 0.25](https://github.com/fluxcd/flux2/releases/tag/v0.25.0) (Jan 2022),
```
?

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+Flux allows for `Kustomizations` and `HelmReleases` to reference sources in different namespaces.
```
makkes, line 18: Does this refer to the `.spec.sourceRef` field? IIRC this has been a `CrossNamespaceSourceReference` for a long time, now, not only since 0.25, hasn't it?

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+This poses a serious security risk for multi-tenant environments as Flux does not prevent tenants from accessing
+known sources outside of their namespaces.
```
makkes, line 20: Maybe we should explicitly describe here that this only allows for read access, not write access.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+Flux does not allow for an `ImageUpdateAutomation` to reference a `GitRepository` in a different namespace.
+This means users have to copy the Git auth secret and the `GitRepository` object in all namespaces
+where `ImageUpdateAutomations` are used. This poses a serious security risk for multi-tenant environments,
+as tenants could use the Git secret to push changes to repositories by impersonating Flux.
```
makkes, line 25: So this is actually less secure than the CrossNamespaceSourceReference used in HelmReleases and Kustomizations. Something to call out here explicitly, I'd say.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+Image Policy/Repository approach to all the other APIs.
+
+### Goals
+
+- Allow source owners to choose which sources are shared and with which namespaces.
+- Allow cluster admins to enforce source ACLs.
+
+### Non-Goals
+
+- Enforce source ACLs by default.
```
makkes, line 42: This will maintain backwards compatibility as far as I understand. Let's be explicit about that here, too: This RFC doesn't introduce backwards-incompatible changes.

rfcs/0002-source-acl/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+When a Flux resource (`Kustomization`, `HelmRelease` or `ImageUpdateAutomation`)
+refers to a source (`GitRepository`, `HelmRepository` or `Bucket`) in a different namespace,
+access is granted based on the source ACL.
+
+The ACL check is performed only if `--enable-source-acl` flag is set to `true` for the following controllers:
+
+- kustomize-controller
+- helm-controller
+- image-automation-controller
```
makkes, line 57: I'm very likely just confused but should source-controller enforce this instead of the origin controllers?

makkes (commented):
Great RFC! I added mostly requests for clarification. Otherwise lgtm

relu (approved):
Looks solid to me, I don't have any comments.

rfcs/0002-source-acl/README.md
stefanprodan: This is enforced on the client side, source-controller has no way of verifying if a HTTP call to download a tar.gz should be allowed or not. To enforce this on the server-side, we'll need something like mTLS and JWT (containing the claims of the tenant maybe?), we'll end up building a little service mesh just for Flux :))

darkowlzz (approved):
LGTM!

stealthybox:
> > I think the primary issue I see with ACLs is understanding what is available to you in terms of sources. As much as I like the idea of having some sort of CLI support for grabbing the available sources, I cannot think of a good way to show the available sources in a secure way to the tenant user
> 
> ACLs are very much like network polices. As far as I know, there is no way for a tenant to query Kubernetes API and get a list of endpoints that allow ingress traffic from its namespace.

@jonathan-innis @stefanprodan 

Should we move forward with ACL's, implementing the ACL's via intermediary RoleBindings and relying on User Impersonation to enforce the policy will have the benefit of making the list of allowed Sources available via the kubectl/flux CLI as long as the CLI user also has access to impersonate the reconcile User.

This is another great reason to not build our own authorization mechanism.
Since the policy is recorded and enforced by the kube-apiserver, all clients that request the resource can be constrained to the proper rules.

example:
```
kubectl auth can-i get gitrepository \
  --all-namespaces --as system:serviceaccount:flux-system:reconciler
```
The apiserver is happy to tell you via RBAC whether or not a particular User can access resources, either generally or by name in whichever namespace.

stefanprodan:
@stealthybox we currently offer a way to disable cross-ns refs and enforce a default SA on multi-tenant environments https://fluxcd.io/docs/installation/#multi-tenancy-lockdown

We’ve reached out to TAG Security for guidance on ACLs. IMO ACLs made like network polices are way easier to setup than RBAC.

phillebaba:
@stefanprodan I think that an important feature for this should be an additional CLI command to to make it clear which namespaces can access what resources. Generally the most difficult part of network policies is debugging.

phillebaba (approved):

stefanprodan:
We should investigate if [ReferenceGrant](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/3766-referencegrant/README.md) is better alternative to the ACL.

Sturgelose:
@stefanprodan, sorry to write in this closed PR, but, I'm understanding that until ReferenceGrant doesn't go GA by the SIG Auth API Group, FluxCD community won't do any movement in order to support cross-namespace support?

I'm asking because there is very little movement in the KEP, and seems this could go for long. I do not know if FluxCD community could provide at least some sort of feature to allow people to get the cross-namespace functionality (and its huge value) right now, with a big disclaimer that once it is supported by K8s core, it will be migrated to it.

From what I see, the API the SIG will implement seems pretty clear and in use in the gateway-api SIG, from where they want to export it from, so even Flux could already work knowing which will how the final implementation might look like.

Of course I understand that making work to have it deprecated soon(ish) has little value, but seeing how things are moving might make sense to invest the time, specially when there are some PRs around already with lots of dicussion or close to have the feature/value implemented.

PR #2086, opened 2021-11-15
[RFC-0001] Define Flux tenancy models

stefanprodan:
The main goal of this RFC is to define the Kubernetes tenancy models supported by Flux.

PR #2093, opened 2021-11-16
[RFC] Flux Multi-Tenancy Mode

stefanprodan:
For multi-tenant environments, we want to offer an easy way of configuring Flux to enforce tenant isolation.

This RFC describes how that can be achieved with a single flag added to `flux bootstrap`.

Supersedes #582

### DIFF
--- a/rfcs/0003-multi-tenancy-mode/README.md
+++ b/rfcs/0003-multi-tenancy-mode/README.md
@@ -1,0 +1,197 @@
+# RFC-0003 Flux Multi-Tenancy Mode
+
+## Summary
+
+For multi-tenant environments, we want to offer an easy way of configuring Flux to enforce tenant isolation
+(as defined by the Soft Multi-Tenancy model from RFC-0001).
+
+When running in the multi-tenant mode, Flux will lock down access to sources (as defined by RFC-0002),
+and will use the tenant service account instead of defaulting to `cluster-admin`.
+
+From an end-user perspective, the multi-tenancy mode means that:
+
+- Platform admins have to create a Kubernetes service account and RBAC in each namespace where
+  Flux performs source-to-cluster reconciliation on behalf of tenants.
+  By default, Flux will have no permissions to reconcile the tenants sources onto clusters.
+- Source owners have to specify with which tenants they wish to share their sources.
+  By default, nothing is shared between tenants.
+
+## Motivation
+
+As of Flux v0.23.0, configuring Flux for soft multi-tenancy requires additional tooling such as Kyverno or OPA Gatekeeper
+to overcome caveats such as:
+- Flux does not require for a service account name to be specified on Flux custom resources that perform
+  source-to-cluster reconciliation. When a service account is not specified, Flux defaults to cluster-admin.
+- Flux does not prevent tenants from accessing known sources outside of their namespaces.
+- Flux does not prevent tenants from subscribing to other tenant's events.
+
+Flux users have been asking for a way to enforce multi-tenancy
+without having to use 3rd party validation webhooks e.g.
+[fluxcd/kustomize-controller#422](https://github.com/fluxcd/kustomize-controller/issues/422).
+
+### Goals
+
+- Enforce service account impersonation for source-to-cluster reconciliation.
+- Enforce ACLs for cross-namespace access to sources.
+
+### Non-Goals
+
+- Enforce tenant's workload isolation with network policies and pod security standards as described
+  [here](https://kubernetes.io/blog/2021/04/15/three-tenancy-models-for-kubernetes/#security-considerations).
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+> As a platform admin, I want to install Flux with lowest privilege/permission level possible.
+
+#### Story 2
+
+> As a platform admin, I want to give tenants full control over their assigned namespaces.
+> So that tenants could use their own repositories and manager the app delivery with Flux.
+
+#### Story 3
+
+> As a platform admin, I want to prevent tenants from changing the cluster-wide configuration.
+> If a tenant adds to their repository a cluster-scoped resource such as a namespace or cluster role,
+> Flux should reject the change and notify the tenant that this operation is not allowed.
+
+### Multi-tenant Bootstrap
+
+When bootstrapping Flux, platform admins should have the option to lock down Flux for multi-tenant environments e.g.:
+
+```shell
+flux bootstrap --security-profile=multi-tenant
+```
+
+The security profile flag accepts two values: `single-tenant` and `multi-tenant`.
+Platform admins may switch between the two modes at any time, either by rerunning bootstrap
+or by patching the Flux manifests in Git.
+
+The `multi-tenant` profile is just a shortcut to setting the following container args in the Flux deployment manifests:
+
+```yaml
+      containers:
+      - name: manager
+        args:
+          - --default-service-account=flux
+          - --enable-source-acl=true
+```
+
+And for disabling cross-namespace references when using the notification API:
+
+```yaml
+kind: Deployment
+metadata:
+  name: notification-controller
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        args:
+          - --same-namespace-refs=true
+```
+
+When running in the `multi-tenant` mode, Flux behaves differently:
+
+- The source-to-cluster reconciliation no longer runs under the service account of
+  the Flux controllers. The controller service account, is only used to impersonate
+  the service account specified in the Flux custom resources (`Kustomizations`, `HelmReleases`).
+- When no service account name is specified in a Flux custom resource,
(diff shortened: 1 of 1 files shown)


scottrigby:
Note to document this in https://github.com/fluxcd/website/issues/598

rfcs/0003-multi-tenancy-mode/README.md
```diff
@@ -0,0 +1,179 @@
...
+          - --default-service-account=flux
+          - --enable-source-acl=true
+```
+
+When running in the `multi-tenant` mode, Flux behaves differently:
+
+- The source-to-cluster reconciliation no longer runs under the service account of
+  the Flux controllers. The controller service account, is only used to impersonate
+  the service account specified in the Flux custom resources (`Kustomizations`, `HelmReleases`).
+- When no service account name is specified in a Flux custom resource,
```
jonathan-innis, line 87: How are we ensuring that the service account exists? Or are we just assuming that if someone hasn't gone to the lengths to ensure that the service account exists that we will just fail reconciliation immediately?

rfcs/0003-multi-tenancy-mode/README.md
```diff
@@ -0,0 +1,179 @@
...
+  a default will be used e.g. `system:serviceaccount:<tenant-namespace>:flux`.
+- When a Flux custom resource (`Kustomizations`, `HelmReleases`, `ImagePolicies`, `ImageUpdateAutomations`)
```
jonathan-innis, line 89: Can we clarify the exact permissions that we are giving to the Flux controllers in this mode? I'm assuming GET/LIST access to .fluxcd.io CRDS, secrets, etc. across all namespaces?

rfcs/0003-multi-tenancy-mode/README.md
```diff
@@ -0,0 +1,179 @@
...
+Platform admins may switch between the two modes at any time, either by rerunning bootstrap
+or by patching the Flux manifests in Git.
+
+The `multi-tenant` profile is just a shortcut to setting the following container args in the Flux deployment manifests:
+
+```yaml
+      containers:
+      - name: manager
+        args:
+          - --default-service-account=flux
```
jonathan-innis, line 78: +1 for this being configurable

rfcs/0003-multi-tenancy-mode/README.md
stefanprodan: We currently set the ready status condition to `false` with the reason “service account not found”.

rfcs/0003-multi-tenancy-mode/README.md
stefanprodan: The controller role is defined here: https://github.com/fluxcd/flux2/blob/main/manifests/rbac/controller.yaml

Should I paste this in RFC?

jonathan-innis: Yes, this would be good to add into the RFC for everyone's visibility

stefanprodan: I'm assuming that people read Flux docs not just this RFC. The RBAC and impersonation is covered here https://fluxcd.io/docs/security/#controller-permissions.

jonathan-innis: Considering that this RFC will drop the cluster-reconciler cluster role binding that the kustomize-controller and helm-controller would normally have had, it would probably be good to call out

stefanprodan: We can’t drop the cluster role binding to cluster-admin, admins can switch between single-tenant and multi-tenant at any time. The controllers use the cluster-admin to impersonate the tenant account.

rfcs/0003-multi-tenancy-mode/README.md
```diff
@@ -0,0 +1,197 @@
...
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: tenant1
+  namespace: tenant1
+spec:
+  interval: 10m0s
+  path: ./
+  prune: true
+  serviceAccountName: flux
```
jonathan-innis, line 178: With this proposed method, the kustomize-controller or helm-controller can reconcile _any_ service account within the namespace of the CR. With this in mind, a user needs to be hyper-aware that any service account in that namespace could be used for reconciliation and must not role-bind to give any service account any kind of elevated privilege that it did not intend for the flux reconcilers (such as giving access to create HelmReleases across the cluster, which essentially opens up Pandora's box). Will this be noted somewhere specifically/is there a way to ensure that flux can only reconcile with service accounts with .metadata.labels intended specifically for flux?

stefanprodan: I think we shouldn't enforce any limitations for the service accounts that are provisioned by platform admins because they may chose to allow a tenant to own multiple namespaces, so they can bind a SA to all those namespaces. A tenant can't create cluster role bindings so it's not possible for them to evade their boundary.

hiddeco: Wouldn't this same concern apply to e.g. a `Deployment` which accepts any arbitrary service account as a configuration value? Which I assume, isn't a problem. Raising the question why it would be different for a Flux Custom Resource?

squaremo: > Wouldn't this same concern apply to e.g. a `Deployment` which accepts any arbitrary service account as a configuration value?

These are the same technically -- a Deployment can run a container that uses kubectl to cause mischief. But the opportunity for mistakes is higher when Flux is introduced. With _only_ Deployments and other "normal" objects you would not usually consider creating a service account with wide-ranging permissions. When you're using Flux, you have much more (RBAC) to think about.

stefanprodan: > With only Deployments and other "normal" objects you would not usually consider creating a service account with wide-ranging permissions. 

Depend on what you consider to be "normal" here, every single Kubernetes addon needs wide-ranging permissions, are these abnormal?

squaremo: Yes, those are abnormal, in this context. The equivalence Hidde was inviting us to draw was between a Deployment and (e.g.,) a HelmRelease, as used by a tenant. They both have a field for supplying a service account, so why bother making especial mention of it for HelmReleases?

My point is: if I'm an admin setting permissions up for a tenant to be able to run a Deployment, all it really needs to be able to do is run pods and talk over the network to other processes. But if I'm setting permissions up for a tenant to be able to create a HelmRelease, the risk is much higher, because it will need to have permissions to create other resources. As Jonathan says, you need to be hyper-aware that such a service account will be usable by anything in that namespace.

hiddeco: I think my point is more that if you make extensive use of RBAC, you should be aware of what this entails (for your tenants). 

The difference between not sharing a service account between multiple resources, or using any arbitrary account that is available to you, does not change for a `Deployment` or a Flux resource.

hiddeco: Not saying here that this means we should not highlight or be upfront about it, but rather that I am not really enthusiastic about adding e.g. label selectors for service account.

squaremo: > I am not really enthusiastic about adding e.g. label selectors for service accounts.

Right -- requiring labels is not a protection, since any process that can run with the service account can do mischief, not just the Flux controllers.

jonathan-innis (approved):
+1 from Microsoft 🚀

rfcs/0003-multi-tenancy-mode/README.md
pjbgf: In this mode, `cluster-reconciler` could bind to a ClusterRole that only provides the impersonation powers it requires. (i.e. allowing a Platform administrator to further limit the ClusterRoleBinding to specific accounts it can run as)

rfcs/0003-multi-tenancy-mode/README.md
pjbgf: The main challenge on the model in which the tenant's service account is created in the tenant's namespace is that anyone with `pod create` permissions in the tenant's namespace would be able to privesc to whatever permissions the flux tenant service account is running as - current example is cluster-admin within that namespace.

In this model platform admins may have to lean on admission controllers to block that path to privilege escalation.

pjbgf: I added some recommendations on how the tenant can be structured to protect their service accounts in the RFC001 (xref: https://github.com/fluxcd/flux2/pull/2086#discussion_r767691190) without requiring admission controllers.

schrej (approved):
One policy to get rid of :smile: 
+1 from DTAG

rfcs/0003-multi-tenancy-mode/README.md
kubasobon: Likewise. This provides a nice way to adopt ServiceAccounts that are already set up.

PR #2113, opened 2021-11-22
Remove deprecated `io/ioutil` from `flux tree`

stefanprodan:
Fix: #1658 this was the last place where `ioutil` was used in Flux v2 and the GitOps Toolkit controllers.

### DIFF
--- a/cmd/flux/tree_kustomization.go
+++ b/cmd/flux/tree_kustomization.go
@@ -23,7 +23,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


dholbach:
There's 3 more small pieces of use of `io/ioutil`, see https://github.com/fluxcd/flux2/issues/1658#issuecomment-975665583 - not sure if you'd like to have separate issues for those?

stefanprodan:
@dholbach the website and the watcher example are not part of flux2 distribution. Going to fix now go-git-providers.

hiddeco (approved):

PR #2141, opened 2021-11-25
Simplify arguments of flux trace command

schrej:
With the `--kind` and `--api-version` flags `flux trace` is pretty tedious to use, especially with custom resources. This PR adds support for using `<resource>/<name>` and `<resource> <name>` instead, like with `kubectl` commands.

Instead of
```
flux trace --api-version helm.toolkit.fluxcd.io/v2beta1 --kind HelmRelease -n flux-system prometheus
```
you can use
```
flux trace -n flux-system helmrelease prometheus
```

It is using the same logic from `k8s.io/cli-runtime` that's used within `kubectl`.

`internal/utils/restclientgetter.go` is something that should really be added to `k8s.io/cli-runtime` as it's needed to get from a `rest.Config` to a `resource.Builder`, which implements resource fetching with only resource and name.

There are also no tests yet. Just adding them to the list in `trace_test.go` is not sufficient, probably because resource discovery is not supported by the test client.

The previous command signature is also still supported.

CC @stefanprodan

### DIFF
--- a/cmd/flux/trace.go
+++ b/cmd/flux/trace.go
@@ -27,8 +27,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/resource"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluxcd/flux2/internal/utils"
@@ -39,20 +41,26 @@ import (
 )
 
 var traceCmd = &cobra.Command{
-	Use:   "trace [name]",
-	Short: "Trace an in-cluster object throughout the GitOps delivery pipeline",
-	Long: `The trace command shows how an object is managed by Flux,
-from which source and revision it comes, and what's the latest reconciliation status.'`,
-	Example: `  # Trace a Kubernetes Deployment
-  flux trace my-app --kind=deployment --api-version=apps/v1 --namespace=apps
+	Use:   "trace <resource> <name> [<name> ...]",
+	Short: "Trace in-cluster objects throughout the GitOps delivery pipeline",
+	Long: `The trace command shows how one or more objects are managed by Flux,
+from which source and revision they come, and what the latest reconciliation status is.
 
-  # Trace a Kubernetes Pod
-  flux trace redis-master-0 --kind=pod --api-version=v1 -n redis
+You can also trace multiple objects with different resource kinds using <resource>/<name> multiple times.`,
+	Example: `  # Trace a Kubernetes Deployment
+  flux trace -n apps deployment my-app
+
+  # Trace a Kubernetes Pod and a config map
+  flux trace -n redis pod/redis-master-0 cm/redis
 
   # Trace a Kubernetes global object
-  flux trace redis --kind=namespace --api-version=v1
+  flux trace namespace redis
 
   # Trace a Kubernetes custom resource
+  flux trace -n redis helmrelease redis
+  
+  # API Version and Kind can also be specified explicitly
+  # Note that either both, kind and api-version, or neither have to be specified.
   flux trace redis --kind=helmrelease --api-version=helm.toolkit.fluxcd.io/v2beta1 -n redis`,
 	RunE: traceCmdRun,
 }
@@ -73,19 +81,6 @@ func init() {
 }
 
 func traceCmdRun(cmd *cobra.Command, args []string) error {
-	if len(args) < 1 {
-		return fmt.Errorf("object name is required")
-	}
-	name := args[0]
-
-	if traceArgs.kind == "" {
-		return fmt.Errorf("object kind is required (--kind)")
-	}
-
-	if traceArgs.apiVersion == "" {
-		return fmt.Errorf("object apiVersion is required (--api-version)")
-	}
-
 	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


internal/utils/restclientgetter.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (r *restClientGetterAdapter) ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error) {
+
+	// The more groups you have, the more discovery requests you need to make.
+	// given 25 groups (our groups + a few custom resources) with one-ish version each, discovery needs to make 50 requests
+	// double it just so we don't end up here again for a while.  This config is only used for discovery.
+	// r.config.Burst = f.discoveryBurst
+
+	cacheDir := filepath.Join(homedir.HomeDir(), ".kube", "cache")
```
stefanprodan, line 43: Flux CLI does not require kubectl to be installed. Will this fail if the `.kube` dir does not exists?

schrej: I've tested locally, it'll just create the directory. You can also see that [in the code](https://github.com/kubernetes/client-go/blob/v0.22.2/discovery/cached/disk/cached_discovery.go#L175).

schrej:
Have you seen/considered [`genericclioptions.ConfigFlags`](https://pkg.go.dev/k8s.io/cli-runtime@v0.22.4/pkg/genericclioptions#ConfigFlags) as a replacement for the current kubeconfig flags yet?
It basically takes over parsing those arguments and producing a rest.Config. It would also eliminate the need for `internal/utils/restclientgetter.go` in this implementation since `ConfigFlags` is already a `RESTClientGetter`.

stefanprodan (commented):
Can you please add an e2e test for the new format here https://github.com/fluxcd/flux2/blob/main/.github/workflows/e2e.yaml#L106

stefanprodan:
> Have you seen/considered genericclioptions.ConfigFlags as a replacement for the current kubeconfig flags yet?

I we could migrate to it, without breaking the current CLI usage, I'm in favour of this change.

stefanprodan:
@schrej I'm having some concerns about using the disk for caching, as it will crash on read-only file systems and on distroless containers. Please take a look here how we currently use the discovery API https://github.com/fluxcd/flux2/blob/main/cmd/flux/completion.go#L61

schrej:
> @schrej I'm having some concerns about using the disk for caching, as it will crash on read-only file systems and on distroless containers. Please take a look here how we currently use the discovery API https://github.com/fluxcd/flux2/blob/main/cmd/flux/completion.go#L61

That should probably also use the cache on disk. In memory is pretty useless for a CLI I'd say, as it gets thrown away after each run. From looking at the code, the Cache seems to fail gracefully (only printing logs) when it can't access the filesystem, so it should be fine. We could also switch to in-memory if the `~/.kube` directory doesn't exist and there is no cache flag set. With switching to cli-runtime that should also be easy to decide at a global level.

Unfortunately the genericclioptions conflicts with the `-s` on some subcommands, which is used for the api server by genericclioptions :/

stefanprodan:
@schrej Ok then can you please refactor `completion.go` to use the RestClientGetter.

gberche-orange:
@schrej did you consider supporting plural versions of resource such as `flux trace -n flux-system helmreleases prometheus`?

Clients such as k9s don't yet have singular resource type, and the conversion from plural to singular is cumbersome, see 
https://github.com/derailed/k9s/issues/1338#issuecomment-986787548
https://github.com/derailed/k9s/pull/1356#issuecomment-992241337

schrej:
> @schrej did you consider supporting plural versions of resource such as `flux trace -n flux-system helmreleases prometheus`?

This is using the same logic as `kubectl`. If it works with `kubectl get` it should also work here.

cmd/flux/trace.go
```diff
@@ -44,16 +46,16 @@ var traceCmd = &cobra.Command{
...
-  flux trace redis-master-0 --kind=pod --api-version=v1 -n redis
+  flux trace -n redis pod redis-master-0
 
   # Trace a Kubernetes global object
-  flux trace redis --kind=namespace --api-version=v1
+  flux trace namespace redis
 
   # Trace a Kubernetes custom resource
-  flux trace redis --kind=helmrelease --api-version=helm.toolkit.fluxcd.io/v2beta1 -n redis`,
+  flux trace -n redis helmrelease redis`,
```
hiddeco, line 58: I would still leave an example with the old flags to both document and highlight backwards compatibility with the previous flag set.

cmd/flux/trace.go
```diff
@@ -137,6 +118,74 @@ func traceCmdRun(cmd *cobra.Command, args []string) error {
...
+
+	infos, err := r.Infos()
+	if err != nil {
+		return nil, err
+	}
+	if len(infos) == 0 {
+		return nil, fmt.Errorf("failed to find object: %w", err)
+	}
+	if len(infos) > 1 {
+		return nil, fmt.Errorf("found more than one object")
```
hiddeco, line 180: Would it be an idea to provide some data from `infos` here to help / guide the user tweak their selector(s)?

cmd/flux/trace.go
schrej: I've expanded it a bit, wdyt?

cmd/flux/trace.go
schrej: I've changed the command to allow tracing multiple objects at once. You can now do both `resource name name ...` and `resource/name resource/name ...`.
They'll get printed with an empty line between different objects, but also with a `---` divider if the output gets processed.

cmd/flux/trace.go
```diff
@@ -137,14 +139,85 @@ func traceCmdRun(cmd *cobra.Command, args []string) error {
...
+		Unstructured().
+		NamespaceParam(*kubeconfigArgs.Namespace).DefaultNamespace().
+		ResourceTypeOrNameArgs(false, args...).
+		ContinueOnError().
+		Latest().
+		Do()
+
+	if err := r.Err(); err != nil {
+		if resource.IsUsageError(err) {
+			return nil, fmt.Errorf("either `<resource>/<name>` or `<resource> <name>` is required as arguments")
```
hiddeco, line 189:
```suggestion
			return nil, fmt.Errorf("either `<resource>/<name>` or `<resource> <name>` is required as an argument")
```

cmd/flux/trace_test.go
```diff
@@ -10,7 +10,7 @@ import (
 func TestTraceNoArgs(t *testing.T) {
 	cmd := cmdTestCase{
 		args:   "trace",
-		assert: assertError("object name is required"),
+		assert: assertError("either `<resource>/<name>` or `<resource> <name>` is required as arguments"),
```
hiddeco, line 13:
```suggestion
		assert: assertError("either `<resource>/<name>` or `<resource> <name>` is required as an argument"),
```

hiddeco (approved):
It requires a rebase on `main`, a DCO sign-off on the latter commit, and a tiny change to the tests; but other than that LGTM. Thank you @schrej :strawberry:  

@stefanprodan can you have another look at the multi-selector / divider?

stefanprodan (approved):
LGTM

Thanks @schrej 🏅

PR #2144, opened 2021-11-25
Use k8s.io/cli-runtime for kubernetes flags

schrej:
Use [k8s.io/cli-runtime](https://pkg.go.dev/k8s.io/cli-runtime) to parse kubernetes related flags. This way the `flux` command accepts (almost) the same flags as `kubectl` to specify which api server to connect to.

One exception is that the api server can only be specified with `--server` and not `-s` since the latter is in conflict with the `--silent` flag that's used by some subcommands.

Flags are managed using `genericclioptions.ConfigFlags`, from which kubernetes client config can be generated. It also simplifies using other components of cli-runtime like the resource.Builder() (see #2141 ).

This change looks very big, but a lot of changes are related to replacing some `rootArgs.*` with `kubeconfigArgs.*` and changes to the `utils` package to accept `genericclioptions.ConfigFlags` or rather `genericclioptions.RESTConfigGetter` (which the former implements) as an input.

The most relevant changes are in `cmd/flux/main.go` and `internal/utils/utils.go`.

Note that I haven't tested this extensively, thus marked as Draft.

CC @stefanprodan

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -21,13 +21,13 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"github.com/fluxcd/flux2/pkg/manifestgen/install"
@@ -99,9 +99,6 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.`,
 var logger = stderrLogger{stderr: os.Stderr}
 
 type rootFlags struct {
-	kubeconfig   string
-	kubecontext  string
-	namespace    string
 	timeout      time.Duration
 	verbose      bool
 	pollInterval time.Duration
@@ -109,19 +106,26 @@ type rootFlags struct {
 }
 
 var rootArgs = NewRootFlags()
+var kubeconfigArgs = genericclioptions.NewConfigFlags(false)
 
 func init() {
-	rootCmd.PersistentFlags().StringVarP(&rootArgs.namespace, "namespace", "n", rootArgs.defaults.Namespace,
-		"the namespace scope for this operation, can be set with FLUX_SYSTEM_NAMESPACE env var")
-	rootCmd.RegisterFlagCompletionFunc("namespace", resourceNamesCompletionFunc(corev1.SchemeGroupVersion.WithKind("Namespace")))
-
 	rootCmd.PersistentFlags().DurationVar(&rootArgs.timeout, "timeout", 5*time.Minute, "timeout for this operation")
 	rootCmd.PersistentFlags().BoolVar(&rootArgs.verbose, "verbose", false, "print generated objects")
-	rootCmd.PersistentFlags().StringVarP(&rootArgs.kubeconfig, "kubeconfig", "", "",
-		"absolute path to the kubeconfig file")
 
-	rootCmd.PersistentFlags().StringVarP(&rootArgs.kubecontext, "context", "", "", "kubernetes context to use")
+	configureDefaultNamespace()
+	kubeconfigArgs.APIServer = nil // prevent AddFlags from configuring --server flag
+	kubeconfigArgs.Timeout = nil   // prevent AddFlags from configuring --request-timeout flag, we have --timeout instead
+	kubeconfigArgs.AddFlags(rootCmd.PersistentFlags())
+
+	// Since some subcommands use the `-s` flag as a short version for `--silent`, we manually configure the server flag
+	// without the `-s` short version. While we're no longer on par with kubectl's flags, we maintain backwards compatibility
+	// on the CLI interface.
+	apiServer := ""
+	kubeconfigArgs.APIServer = &apiServer
+	rootCmd.PersistentFlags().StringVar(kubeconfigArgs.APIServer, "server", *kubeconfigArgs.APIServer, "The address and port of the Kubernetes API server")
+
 	rootCmd.RegisterFlagCompletionFunc("context", contextsCompletionFunc)
+	rootCmd.RegisterFlagCompletionFunc("namespace", resourceNamesCompletionFunc(corev1.SchemeGroupVersion.WithKind("Namespace")))
 
 	rootCmd.DisableAutoGenTag = true
 	rootCmd.SetOut(os.Stdout)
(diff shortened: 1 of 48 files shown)


cmd/flux/main.go
```diff
@@ -99,29 +99,32 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.`,
...
 	verbose      bool
 	pollInterval time.Duration
 	defaults     install.Options
 }
 
 var rootArgs = NewRootFlags()
+var kubeconfigArgs = genericclioptions.NewConfigFlags(false)
 
 func init() {
-	rootCmd.PersistentFlags().StringVarP(&rootArgs.namespace, "namespace", "n", rootArgs.defaults.Namespace,
```
stefanprodan, line 114: We need to preserver the default value from `rootArgs.defaults.Namespace` otherwise we'll break everyone's clusters. Flux default namespace is `flux-system` not `default`.

stefanprodan (changes requested):

cmd/flux/main.go
schrej: Yeah, I've just fixed that with the force push :smile:

cmd/flux/main.go
```diff
@@ -138,30 +141,17 @@ func NewRootFlags() rootFlags {
...
-		}
-	}
-}
-
 func configureDefaultNamespace() {
+	*kubeconfigArgs.Namespace = rootArgs.defaults.Namespace
 	fromEnv := os.Getenv("FLUX_SYSTEM_NAMESPACE")
-	if fromEnv != "" && rootArgs.namespace == rootArgs.defaults.Namespace {
-		rootArgs.namespace = fromEnv
+	if fromEnv != "" {
```
stefanprodan, line 153: This is not ok, please restore the previous behaviour, we only look for `FLUX_SYSTEM_NAMESPACE ` if the namespace arg is different to the default `flux-system`.

stefanprodan (changes requested):

cmd/flux/main.go
stefanprodan: The behaviour is that a non default value takes precedence over the env var.

schrej: The precedence is still the same. It uses the cli-arg, then the variable, then the default.
We set `kubeconfigArgs.Namespace` to default, then - if set - to the env variable. The `AddFlags()` method will then use whatever is in `kubeconfigArgs.Namespace` as the default for the flag.

stefanprodan: This should use the `bar` namespace:
```
export FLUX_SYSTEM_NAMESPACE=foo
flux get all -n bar
```

This should use the `foo` namespace:
```
export FLUX_SYSTEM_NAMESPACE=foo
flux get all -n flux-system
```

This should use the `foo` namespace:
```
export FLUX_SYSTEM_NAMESPACE=foo
flux get all
```

schrej: That's not the current behavior though:
```
~/projects$ flux version --client                                                                                                                                  schiff-controlplane-reftmdc.reftmdc.bn
flux: v0.24.0
~/projects$ export FLUX_SYSTEM_NAMESPACE=gitops; flux reconcile source git -n flux-system deployment                                               schiff-controlplane-reftmdc.reftmdc.bn
✗ gitrepositories.source.toolkit.fluxcd.io "deployment" not found
~/projects$ flux reconcile source git -n gitops deployment                                                                                         schiff-controlplane-reftmdc.reftmdc.bn
► annotating GitRepository deployment in gitops namespace
```

stefanprodan:
So now we have two overlapping flags:
```
      --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
      --timeout duration               timeout for this operation (default 5m0s)
```

I think we could mark `timeout` as deprecated and not used, set the `request-timeout` default to `5m`, after a couple of minor release we would remove `timeout`. This implies wiring `request-timeout` instead of `timeout` for all commands...

schrej:
> I think we could mark `timeout` as deprecated and not used, set the `request-timeout` default to `5m`, after a couple of minor release we would remove `timeout`. This implies wiring `request-timeout` instead of `timeout` for all commands...

But they do different things. `--request-timeout` configures a timeout for kubernetes requests. If there are multiple requests the total duration could exceed `--request-timeout`.

We could just remove the `--request-timeout` flag. That would reduce parity with kubectl again, but the current behavior of `--timeout` might make more sense for flux.

stefanprodan:
> We could just remove the --request-timeout flag. That would reduce parity with kubectl again, but the current behavior of --timeout might make more sense for flux.

Agreed!

cmd/flux/main.go
```diff
@@ -138,30 +142,17 @@ func NewRootFlags() rootFlags {
...
 		logger.Failuref("%v", err)
 		os.Exit(1)
 	}
 }
 
-func configureKubeconfig() {
-	switch {
-	case len(rootArgs.kubeconfig) > 0:
-	case len(os.Getenv("KUBECONFIG")) > 0:
-		rootArgs.kubeconfig = os.Getenv("KUBECONFIG")
```
stefanprodan, line 153: Does cli-runtime reads `KUBECONFIG` from env?

schrej: it uses `clientcmd.NewDefaultClientConfigLoadingRules()` to load the config, which includes the `KUBECONFIG` env var.

schrej: But I don't think it gets added to `kubeconfigArgs.KubeConfig`, it's just respected when calling `kubeconfigArgs.ToRESTConfig()`.

stefanprodan (approved):
LGTM

Thanks @schrej 🏅 

Going to ask for a 2nd review in case I missed something.

schrej:
Just to state this again: I didn't test this very thoroughly, also because I don't have a proper setup nor the time to run all of the commands to see if they still work. I assume you did that already, but I strongly recommend to perform some manual testing on this before merging.
I'll remove the "Draft" status now.
Once it's merged I'll also refactor #2141.

stefanprodan:
@schrej can you please rebase with latest main, I'm going to test this manually once again and merge it.

schrej:
I've also rebased #2141 since that becomes much cleaner once this is merged.

hiddeco (approved):
Function wise this looks solid to me, and a much welcome improvement for UX. Thank you @schrej  :bow: :100: 

One (follow-up) remark while running some simple tests with the binary is the styling of some of the flag argument descriptions, but this is not directly related to the changes in this PR:

```console
Flags:
      --as string                      Username to impersonate for the operation
      --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --cache-dir string               Default cache directory (default "/home/hidde/.kube/cache")
      --certificate-authority string   Path to a cert file for the certificate authority
      --client-certificate string      Path to a client certificate file for TLS
      --client-key string              Path to a client key file for TLS
      --cluster string                 The name of the kubeconfig cluster to use
      --context string                 The name of the kubeconfig context to use
  -h, --help                           help for flux
      --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kubeconfig string              Path to the kubeconfig file to use for CLI requests.
  -n, --namespace string               If present, the namespace scope for this CLI request (default "flux-system")
      --server string                  The address and port of the Kubernetes API server
      --timeout duration               timeout for this operation (default 5m0s)
      --tls-server-name string         Server name to use for server certificate validation. If it is not provided, the hostname used to contact the server is used
      --token string                   Bearer token for authentication to the API server
      --user string                    The name of the kubeconfig user to use
      --verbose                        print generated objects
  -v, --version                        version for flux
```

(see e.g. `--verbose`, `-v, --version`)

PR #2179, opened 2021-12-06
Explain authorisation model and mechanisms

squaremo:
The multi-tenancy implementations described rely on impersonation and
remote apply; to make this RFC stand by itself, those need to be
explained, along with the authorisation model (how Flux "decides" what
it's allowed to do).

This commit adds a summary of the authorisation model, impersonation,
and remote apply, and rejigs the headings a little to make space.

### DIFF
--- a/rfcs/0001-multi-tenancy/README.md
+++ b/rfcs/0001-multi-tenancy/README.md
@@ -2,37 +2,82 @@
 
 ## Summary
 
-This RFC defines two models for multi-tenancy using Flux and describes their reference implementations.
+This RFC explains the mechanisms available in Flux for implementing multi-tenancy, defines two
+models for multi-tenancy, and gives reference implementations for those models.
 
 ## Motivation
 
-The documentation [here](https://fluxcd.io/docs/) describes the security model of Flux.
-To this point, the Flux project has provided
-[examples of multi-tenancy](https://github.com/fluxcd/flux2-multi-tenancy/tree/v0.1.0),
-but not explained exactly how they relate to Flux's security model.
-This RFC explains two multi-tenancy implementations, their security properties,
-and how they map to the security model.
+To this point, the Flux project has provided [examples of
+multi-tenancy](https://github.com/fluxcd/flux2-multi-tenancy/tree/v0.1.0), but not explained exactly
+how they relate to Flux's authorisation model. This RFC explains two multi-tenancy implementations,
+their security properties, and how they are implemented within the authorisation model.
 
 ### Goals
 
+- Explain the mechanisms available in Flux for supporting multi-tenancy
 - Define two models for multi-tenancy, "soft multi-tenancy" and "hard multi-tenancy".
 - Explain when each model is appropriate.
-- List the tenancy models supported by Flux.
 - Describe a reference implementation of each model with Flux.
 
 ### Non-Goals
 
-- Give an exhaustive account of multi-tenancy implementations.
-- Provide an [end-to-end workflow](](https://github.com/fluxcd/flux2-multi-tenancy/tree/v0.1.0))
-  of how to setup multi-tenancy with Flux.
+- Give an exhaustive account of multi-tenancy implementations in general.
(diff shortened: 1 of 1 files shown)


rfcs/0001-multi-tenancy/README.md
```diff
@@ -2,37 +2,82 @@
...
+on their behalf. Since tenants control Flux via its API objects, this becomes a matter of attaching
+RBAC rules to Flux API objects. There are two mechanisms that do this, "impersonation" and "remote
+apply".
+
+#### Impersonation
+
+The Kustomize controller and Helm controller both apply arbitrary sets of Kubernetes configuration
+to a cluster. These controllers are subject to authorisation on two counts:
+
+ - when accessing the Kubernetes resources that specify a particular "apply" operation -- for
```
stefanprodan, line 54: This is not what Flux currently does, the sources are not subject to RBAC but to ACL (when RFC-0002 gets implemented).

rfcs/0001-multi-tenancy/README.md
```diff
@@ -2,37 +2,82 @@
...
+   example, the `GitRepository` object that gives access to the piece of configuration to be
+   applied; and,
+ - when creating, updating and deleting Kubernetes resources in the process of applying a piece of
```
stefanprodan, line 57: Also health checks, querying for secrets and configmap (valuesFrom, decryption, kubeConfig)

rfcs/0001-multi-tenancy/README.md
```diff
@@ -2,37 +2,82 @@
...
+`.spec.serviceAccountName` in a `Kustomization` object][serviceAccountName]) for both accessing
+resources and applying configuration. This lets a user constrain the operations mentioned above with
+RBAC.
+
+#### Remote apply
+
+The Kustomize controller and Helm controller are able to apply a set of configuration to a cluster
+other than the cluster in which they run. If the specification [refers to a secret containing a
+"kubeconfig" file][kubeconfig], the controller will construct a client using that kubeconfig, then
+the client used to apply the specified set of configuration. THe effect of this is that the
```
stefanprodan, line 71:
```suggestion
the client used to apply the specified set of configuration. The effect of this is that the
```

rfcs/0001-multi-tenancy/README.md
squaremo: Owps, getting ahead of myself.

rfcs/0001-multi-tenancy/README.md
squaremo: That's what I meant inn the previous line item. All the things it needs to prepare for applying. Are the things that do and don't get fetched with the impersonated account documented (in one place)?

stefanprodan: After KC creates the Kubernetes client using the service account impersonation, it runs the following actions in order:
- fetch decryption secret
- decrypt the source
- build the Kustomize overlay
- fetch the variables values from secrets and configmaps
- run post-build substitutions
- generate the finally manifests
- validate the manifests using dry-run server-side apply
- apply the manifests using server-side apply
- wait for reconciled objects to become ready using kstatus

squaremo: I think we need to put this in the documentation, then refer to it from a summary here, rather than spelling it out here. I'll put in a PR for that.

stefanprodan (approved):
LGTM

Thanks @squaremo 🥇

rfcs/0001-multi-tenancy/README.md
squaremo: > After KC creates the Kubernetes client using the service account impersonation, it runs the following actions in order:
> 
>  - fetch decryption secret

This is true in the sense that the decryption secret is fetched _after_ the controller creates the impersonating account, but misleading, because the impersonating account is not used for fetching decryption secrets:
https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_controller.go#L606

squaremo:
(I'm not a maintainer here, someone else will have to merge this)

Issue #2192, opened 2021-12-08
Monitoring guide is broken


kingdonb:
### Describe the bug

We're just doing an internal demo and we noticed that the Monitoring guide appears to be broken. The Flux Control Plane dashboard appears on the Grafana deployment as expected, but none of the pods match.

I have a customized version of the manifests from a fork, which could have introduced problems, but we had someone follow along with the instructions from https://fluxcd.io/docs/guides/monitoring/ and they had similar results. Metrics are not being collected from the Flux pods for some reason.

@nalum found the issue that I was experiencing was not unique to my customized version. I have one custom pod and that pod was scraped successfully for some reason. His installation had zero customized pods and zero pods were scraped.

### Steps to reproduce

Follow the monitoring guide. Notice that metrics are not being collected for the flux-system pods as expected.

### Expected behavior

Expected to see that Flux control plane was being monitored and metrics are collected. As a double-check, I deleted all the pods in the cluster and allowed them to restart. It didn't change anything.

### Screenshots and recordings

<img width="1413" alt="Screen Shot 2021-12-08 at 12 41 31 PM" src="https://user-images.githubusercontent.com/3286998/145256990-4e7e1415-2990-46d1-b98a-1c46983ae1b1.png">

### OS / Distro

MacOS 12.0.1 (Monterey)

### Flux version

v0.24.0

### Flux check

$ flux check
► checking prerequisites
✔ Kubernetes 1.21.7 >=1.19.0-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.14.0
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.18.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.14.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.18.1
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.19.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.19.0
✔ all checks passed

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
Duplicate of #2150

stefanprodan:
To fix this we need to add each controller with its own key in the pod monitor.

Nalum:
[...]

Issue #2150, opened 2021-11-28
monitoring-config/podmonitor.yaml only pulls data from image-reflector-controller


ajrpayne:
### Describe the bug

Of the controllers listed, only the last controller in the list has any data in Prometheus. If I change the list, it will always pull data from the final controller listed only.

 selector:
    matchExpressions:
      - key: app
        operator: In
        values:
          - helm-controller
          - source-controller
          - kustomize-controller
          - notification-controller
          - image-automation-controller
          - image-reflector-controller

### Steps to reproduce

Follow steps listed here deploying in kind. https://fluxcd.io/docs/guides/monitoring/

### Expected behavior

I would expect to see data from all controllers not just one.

### OS / Distro

arch linux using kind

### Flux version

flux version --client flux: v0.23.0

### Flux check

flux check
► checking prerequisites
✗ flux 0.23.0 <0.24.0 (new version is available, please upgrade)
✔ Kubernetes 1.21.1 >=1.19.0-0
► checking controllers
✔ all checks passed

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

piermotte:
I have the same problem, there is no active targets if i use flux `podmonitor.yaml`

kingdonb:
We found two ways to solve this:

Either pin kube-prometheus-operator at 19.3.0 as in:
* https://github.com/fluxcd/flux2/pull/2194

or upgrade to prometheus-operator v0.52.1 or greater. This should happen naturally as soon as a release of kube-prometheus-operator includes the v0.52.1 or greater appVersion.

https://github.com/kingdonb/flux2/commit/bcaa4c692881e0be24cd18ea44630b4d10b2b132

Here's an example of how to resolve it without downgrading in the mean-time. (Set values invoking the new prom-operator)

```
prometheusOperator:
  image:
    tag: v0.52.1
  prometheusConfigReloader:
    image:
      tag: v0.52.1
```

piermotte:
After setting prometheusOperator image, it works !  thx @kingdonb !

PR #2193, opened 2021-12-08
Split PodMonitor into separate objects

Nalum:
This relates to issues #2192 and #2150

We are breaking the initial manifest up into multiple as it doesn't like when `selector.matchExpressions[*].values` has more than 1 value. Let me know if there is more to this.

Please let me know if you'd prefer this broken out into multiple files. I'll be able to test this change out tomorrow.

### DIFF
--- a/manifests/monitoring/monitoring-config/podmonitor.yaml
+++ b/manifests/monitoring/monitoring-config/podmonitor.yaml
@@ -1,7 +1,8 @@
+---
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
-  name: flux-system
+  name: helm-controller
   namespace: flux-system
   labels:
     app.kubernetes.io/part-of: flux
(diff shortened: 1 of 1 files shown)


kingdonb:
> it doesn't like when selector.matchExpressions[*].values has more than 1 value

This seems like a bug, and I'm pretty sure this worked before. I went looking through release notes:
https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-22x-to-23x

and didn't find anything that particularly jumped out at me as explaining this. However we do have this release from the guide configured without any version number:

https://github.com/fluxcd/flux2/blob/main/manifests/monitoring/kube-prometheus-stack/release.yaml

This explains how it broke!

I suspect if we reverted `kube-prometheus-stack` to an earlier version, we would find it still works there.

It's up to you if we want to figure out where that initial problem came from. To me, if this selector can't match multiple values in the array, that's a bug and that should be addressed. I don't want to say "not our bug" (but for the time being, we have to do something as our guide is broken!)

```
selector:
  matchExpressions:
    - key: app
      operator: In
      values:
        - kustomize-controller
```

I tested your change and it resolves the issue for now 👍

kingdonb:
I definitely see 6 controllers with the original config on:

```
$ ...
kube-prometheus-stack	21.0.5
```

and I did not see the 6 controllers (only the last of them, image-reflector-controller, which happened to be the last array element) on this version:

```
23.1.2
```

It should take me a couple of minutes to bisect and find out exactly what version broke it!

Nalum:
Good work 👍

I feel it's probably better to pin the version and then monitor for it to be fixed upstream. I can spend a bit of time on that tomorrow and look for/create an issue for it on the upstream project.

kingdonb:
It looks like the problem was introduced in the [20.0.0 chart release](https://github.com/prometheus-community/helm-charts/releases/tag/kube-prometheus-stack-20.0.0).

Everything is fine on 19.2.3, then 20.0.0 is the next release, and subsequent releases all have this issue.

Just to confirm, 23.1.2 (the release from just two hours ago) also has this issue.

I'll submit a docs update to pin the version at 19.2.3 for the time being, since I agree that is the best way for now.

kingdonb:
Our issue: https://github.com/prometheus-operator/prometheus-operator/issues/4435

grafjo:
@kingdonb  this issue is fixed with release [kube-prometheus-stack-23.2.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-prometheus-stack-23.2.0) and dashboards have data again

PR #2194, opened 2021-12-08
monitoring: Pin kube-prometheus-stack  to v19.3.0

kingdonb:
Something wrong (a bug) in kube-prometheus-stack 20.0.0 and on has broken our example.

See:

* https://github.com/fluxcd/flux2/pull/2193 and
* #2192 for more information.

### DIFF
--- a/manifests/monitoring/kube-prometheus-stack/release.yaml
+++ b/manifests/monitoring/kube-prometheus-stack/release.yaml
@@ -6,6 +6,7 @@ spec:
   interval: 5m
   chart:
     spec:
+      version: 19.3.0
       chart: kube-prometheus-stack
       sourceRef:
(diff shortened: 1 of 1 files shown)


kingdonb:
It looks like the issue is known and resolved upstream, just waiting for a chart revision bump to publish and deploy prometheus-operator v0.52.1

https://github.com/prometheus-operator/prometheus-operator/issues/4435#issuecomment-985192529

I think they're likely to publish the rev now, I think we probably don't have to merge this PR. (If the prom-operator chart release happens soon, then it would be better to not have to revert this change when their upgrade has finally resolved it.)

stefanprodan (approved):
LGTM

Thanks @kingdonb 🏅

PR #2195, opened 2021-12-09
Removing Kubernetes API Request Duration Graph

Nalum:
Fix: #2192

Updated this will look like the following:

Screenshot 2021-12-09 at 09-45-13 Flux Control Plane - Grafana


stefanprodan (approved):
LGTM

Thanks @Nalum 🏅

PR #2197, opened 2021-12-09
Update bootstrap github docs to list permissions

SamLR:
Document the required permissions for bootstrapping pre-existing repositories.

I'm not sure if this is the correct place to document this as it feels to detailed for CLI output 
but (as I understand it) the output of this is then passed to the website automatically and, not
knowing this took a while to debug the problems I was having.

### DIFF
--- a/cmd/flux/bootstrap_github.go
+++ b/cmd/flux/bootstrap_github.go
@@ -38,11 +38,12 @@ import (
 var bootstrapGitHubCmd = &cobra.Command{
 	Use:   "github",
 	Short: "Bootstrap toolkit components in a GitHub repository",
-	Long: `The bootstrap github command creates the GitHub repository if it doesn't exists and
+	Long: `The bootstrap github command creates the GitHub repository if it doesn't exist and
 commits the toolkit components manifests to the main branch.
 Then it configures the target cluster to synchronize with the repository.
 If the toolkit components are present on the cluster,
-the bootstrap command will perform an upgrade if needed.`,
+the bootstrap command will perform an upgrade if needed.
+Admin permissions are needed on pre-existing repositories.`,
 	Example: `  # Create a GitHub personal access token and export it as an env var
   export GITHUB_TOKEN=<my-token>
 


cmd/flux/bootstrap_github.go
```diff
@@ -38,11 +38,13 @@ import (
...
 	Short: "Bootstrap toolkit components in a GitHub repository",
-	Long: `The bootstrap github command creates the GitHub repository if it doesn't exists and
+	Long: `The bootstrap github command creates the GitHub repository if it doesn't exist and
 commits the toolkit components manifests to the main branch.
 Then it configures the target cluster to synchronize with the repository.
 If the toolkit components are present on the cluster,
-the bootstrap command will perform an upgrade if needed.`,
+the bootstrap command will perform an upgrade if needed.
+Admin permissions are needed on pre-existing repositories
+in order to create a deploy key.`,
```
hiddeco, line 47: I think the managing of team permissions requires elevated repository permissions as well (as do some other things that happen if you do a strict reconcile, e.g. setting the repository from public to private and vice-versa.

SamLR: I suspect you're right here. I included the deploy key bit as that was the specific one I hit. If it looks like there's going to be a long list happy to remove the specific actions and just cut back to 'Admin permissions are need on pre-existing repositories' : )

SamLR: @hiddeco I've updated it to remove the specific actions and just list the role

hiddeco: But that's also not true because of `--token-auth` in combination with not using `--team` will also happily reconcile without administrative permissions I think, as long as you are able to commit :-).

That makes the summary a bit difficult.

hiddeco: "Admin permissions may be required on the repository for certain configuration actions to succeed"?

SamLR: I was worried this might be the case. Feels that this may be better suited in a different location, either a specific page in the docs or spread between the specific flags if they modify the defaults?

SamLR: How about "for most configurations admin permissions on the repository are required but this can be lowered by removing requirements to use deploy keys and modify team access"? It's relatively consise and nudges towards the modifications that are being done that might raise/lower permissions (could also link the docs although the URL is huge https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role)

stefanprodan: I think this info is better placed here:  https://github.com/fluxcd/website/blob/main/content/en/docs/installation.md#github-and-github-enterprise

SamLR: > I think this info is better placed here: https://github.com/fluxcd/website/blob/main/content/en/docs/installation.md#github-and-github-enterprise

I'm not sure why I forgot about this page. Thanks!

PR #2216, opened 2021-12-14
Fix port overwrite on bootstrap

achetronic:
**The problem:**
When you use the bootstrap command on Flux CLI using the generic git subcommand, the port is always removed from the URL that is written on gotk-sync.yaml. If you use a Git server on non-standard port this is a problem. Removing this line will keep the port untouched, when it is standard and when it is not.

**The proof:**
Screenshot from 2021-12-14 12-32-56

**The fix:**
Simply avoid the overwrite of the port when parsing the fields of the url

**Contributors:**
Alby Hernández (@achetronic)
Sebastián Vargas (@develolux)

### DIFF
--- a/cmd/flux/bootstrap_git.go
+++ b/cmd/flux/bootstrap_git.go
@@ -164,7 +164,6 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		// Configure repository URL to match auth config for sync.
 		repositoryURL.User = nil
 		repositoryURL.Scheme = "https"
-		repositoryURL.Host = repositoryURL.Hostname()
 	} else {
 		secretOpts.PrivateKeyAlgorithm = sourcesecret.PrivateKeyAlgorithm(bootstrapArgs.keyAlgorithm)
 		secretOpts.Password = gitArgs.password
--- a/.github/runners/prereq.sh
+++ b/.github/runners/prereq.sh
@@ -21,6 +21,7 @@ set -eu
 KIND_VERSION=0.11.1
 KUBECTL_VERSION=1.21.2
 KUSTOMIZE_VERSION=4.1.3
+HELM_VERSION=3.7.2
 GITHUB_RUNNER_VERSION=2.285.1
 PACKAGES="apt-transport-https ca-certificates software-properties-common build-essential libssl-dev gnupg lsb-release jq"
 
@@ -51,6 +52,12 @@ curl -Lo ./kustomize.tar.gz https://github.com/kubernetes-sigs/kustomize/release
   && tar -zxvf kustomize.tar.gz \
   && rm kustomize.tar.gz
 install -o root -g root -m 0755 kustomize /usr/local/bin/kustomize
+
+# install helm
+curl -Lo ./helm.tar.gz https://get.helm.sh/helm-v${HELM_VERSION}-linux-arm64.tar.gz \
+  && tar -zxvf helm.tar.gz \
+  && rm helm.tar.gz
+install -o root -g root -m 0755 linux-arm64/helm /usr/local/bin/helm
 
 # download runner
 curl -o actions-runner-linux-arm64.tar.gz -L https://github.com/actions/runner/releases/download/v${GITHUB_RUNNER_VERSION}/actions-runner-linux-arm64-${GITHUB_RUNNER_VERSION}.tar.gz \


achetronic:
@stefanprodan Hello sir, 

I mention you here because I didn't open an issue for this but fixed the bug here and is my first PR on this project, so I don't control all the contributing guidelines yet :)

Could you please review this PR to fix the bug?
Thank you

hiddeco:
As a starter, thank you for your first contribution :bow: :100: 

While I do think that the observation of faulty behavior is correct, I think that this proposed changes introduces a new issue.

Given the scenario where someone configures a `--url ssh://git@example.com:9999/repository.git` to perform Git operations from their machine over SSH, but also sets `--token-auth` to enable HTTPS on the cluster side, it will now result in the SSH port being configured for the HTTPS URL.

Due to this, I think I missed a configuration possibility while writing this bit of code, and the `--ssh-hostname` flag option should actually be a `--in-cluster-hostname` (or something a bit more accurate and self-explanatory), to overwrite local vs cluster config, instead of overwriting _just_ SSH. WDYT?

achetronic:
> As a starter, thank you for your first contribution bow 100
> 
> While I do think that the observation of faulty behavior is correct, I think that this proposed change introduces a new issue.
> 
> Given the scenario where someone configures a `--url ssh://git@example.com:9999/repository.git` to perform Git operations from their machine over SSH, but also sets `--token-auth` to enable HTTPS on the cluster side, it will now result in the SSH port being configured for the HTTPS URL.
> 
> Due to this, I think I missed a configuration possibility while writing this bit of code, and the `--ssh-hostname` flag option should actually be a `--in-cluster-hostname` (or something a bit more accurate and self-explanatory), to overwrite local vs cluster config, instead of overwriting _just_ SSH. WDYT?

Hello again,

We read your message and started working on changes to fix the PR. Changed the behavior and it is supposed to work properly now. Please, review it but it seems to be ok on our organic tests. 

For now, we kept the name of the previous flag, but the intention is to replace it with `--in-cluster-host` to include the port on possible overwrites ( `--in-cluster-hostname` is not accurate with the expected behavior )

What do you think about the changes? :)

cmd/flux/bootstrap_git.go
```diff
@@ -152,7 +152,9 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		Namespace:    rootArgs.namespace,
 		TargetPath:   gitArgs.path.String(),
 		ManifestFile: sourcesecret.MakeDefaultOptions().ManifestFile,
+		SSHHostname:  repositoryURL.Host,
```
souleb, line 155: What if we don't use ssh but only https?

cmd/flux/bootstrap_git.go
```diff
@@ -164,7 +166,6 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		// Configure repository URL to match auth config for sync.
 		repositoryURL.User = nil
 		repositoryURL.Scheme = "https"
-		repositoryURL.Host = repositoryURL.Hostname()
```
souleb, line 167: I think this should stay because `bootstrapArgs.tokenAuth` take precedence. Why not make sure we don't loose the port number here?

cmd/flux/bootstrap_git.go
```diff
@@ -205,6 +201,15 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		RecurseSubmodules: bootstrapArgs.recurseSubmodules,
 	}
 
+	// Modify the synchronization URL on GitRepository using the --ssh-hostname flag
+	if (bootstrapArgs.sshHostname != "") && (repositoryURL.Host != bootstrapArgs.sshHostname) {
```
souleb, line 205: I think `bootstrapArgs.sshHostname` shall be evaluated only if `tokenAuth` is not specified.

cmd/flux/bootstrap_git.go
achetronic: In the tests I did yesterday it seems to work with both cases, when setting the repo using https and over ssh. The problem is that, according to [this comment](https://github.com/fluxcd/flux2/pull/2216#issuecomment-993690220), the flag should be called in another way because of the behavior (it is useful to write another host in the flyx-system GitRepository, am I wrong?) so I re-used the same flag to rename it correctly in another PR

cmd/flux/bootstrap_git.go
achetronic: If the expected behavior is to modify the host on GitRepository, should not be evaluated in both cases? I mean, it is possible that the naming of the flag is not right or I missunderstood what is expected for that flag.

This is done this way because you should keep the port or change it in both cases, for example, in out company we have all ports changed, even the SSH ones

cmd/flux/bootstrap_git.go
achetronic: This proposal was the first approach we thought about, but we wanted to fix the root cause and for that is this PR, because the same could happens with ssh connections when not in the port 22 :)

cmd/flux/bootstrap_git.go
souleb: The `bootstrapArgs.sshHostname` is used to specify an ssh host if it's different that the https one. We don't need  it when using only https.

With  `repositoryURL.Host = my-git-server.com:7990`,  `bootstrapArgs.sshHostname = my-git-server.com:7999` and the option `bootstrapArgs.tokenAuth = true`, I expect the sync to happen on `my-git-server.com:7990` and discard the `bootstrapArgs.sshHostname`.

achetronic: Knowing this, I could let the original code and just removing the line `repositoryURL.Host = repositoryURL.Hostname()` would do the trick, no? because it will keep the value for `repositoryURL.Host` and this value can come with port or without it from the `--url` flag

souleb: Yes I think we can test that.

achetronic: Fixed. It seems to keep the port when --token-auth enabled by not changing the `repositoryURL.Host`.
I tested bootstraping with ssh too and it seems to work.
Could you please confirm the behavior?

cmd/flux/bootstrap_git.go
achetronic: Fixed as we discussed :) 
https://github.com/fluxcd/flux2/pull/2216#discussion_r770030286

souleb (commented):
can you please squash your commits?

achetronic:
> can you please squash your commits?

Done :) 
thank you for the reminder!

achetronic:
Closed because the clean fix is in another PR pending of review

PR #2222, opened 2021-12-16
RFC0005 - attributes in image automation

GregoireW:
Is flux image automation a good solution to extract additional data from image tag?

Signed-off-by: GregoireW <24318548+GregoireW@users.noreply.github.com>


dungdm93:
It would be nice if image SHA256 available.
In some case I wanna use `app/foobar@sha256:294d69bb580a614ed3128969b95b5355c480e84704d826cdf73e790b5a6e63fc`

relu:
@GregoireW I'm sorry for the delay on this, I've started to review this and I'd like to resume the discussions. If you're still interested in contributing, would you be able to reach out to me in [CNCF Slack workspace](https://slack.cncf.io/)?

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+**Status:** provisional
+
+**Creation date:** 2021-12-16
+
+**Last update:** 2021-12-16
+
+## Summary
+
+Image automation controller can update some attributes of a kubernetes object.
```
relu, line 11: Would you mind changing this to something that more directly explains what the RFC covers? Something like "Flux should allow referencing more container image metadata attributes in the image automation Setters strategy"

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
+# RFC-0005 Managed attributes on Image Automation
```
relu, line 1:
```suggestion
# RFC-0005 Extend supported list of image automation marker reference attributes
```

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Some automation or observability tools can use label to identify better a 
+kubernetes object. It can be linked to a version, to a date, to a code 
+origin... For multiple reason, the image tag can reflect poorly this
+data. An example can be given by the image reflector controller which 
+can extract a part of the tag and use it to sort and select the correct one.
+
+### Goals
+
+This RFC aims to describe a way to extract such additional value from the
```
relu, line 25: Maybe change this to a list and add the digest information extraction under the umbrella of this RFC:
- Allow referencing `filterTags` pattern extracted parts from image policies
- Extract digest information and make it available in the markers https://github.com/fluxcd/image-reflector-controller/issues/214

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+This raise the question on should this be included in flux or not.
+
+## Design Details
+
+Simple update on the image automation controller should be enough. Today a 
+filter in the image policy is like: 
+
+```yaml
+    extract: $ts
+    pattern: ^pr-(?P<pr>.*)-(?P<ts>\d*)-(?P<sha1>.*)$
```
relu, line 57: What would happen if we'd just use regular capture groups instead of named groups? Maybe we should be able to support that as well by referencing the indices?

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+It is possible to modify the image automation to take comment like:
+
+```yaml
+  # {"$imagepolicy": "{namespace}:{imagepolicy}:{attributes}"
```
relu, line 63: It might make sense to delimitate the attributes extracted from `.spec.filterTags.pattern` from the others to avoid any collisions or confusion. 
I'm proposing we go with: `# {"$imagepolicy": "{namespace}:{imagepolicy}:tagparts[{attribute}]"`, where `attribute` can be either a named capture group or a index number.

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+with `attributes` a name of a capture group on the pattern. 
+
+From previous pattern example, accepted attributes will be:
```
relu, line 68: What should happen if the user would reference a non-existent attribute?

rfcs/0005-managed-attributes-image-automation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### Alternatives
+
+An alternative would be to build a mutation web hook which would be able to 
+filter all object and interact with them directly. It would be more generic 
+but heavier to build. 
+This raise the question on should this be included in flux or not.
+
+## Design Details
```
relu, line 50: Since this heavily depends on what information image policies expose I think we need to detail how this would be implemented in the IRC as well.

relu (changes requested):
Here are some suggestions after a first pass, I hope these help.

Thank you for putting this together!

rfcs/0005-managed-attributes-image-automation/README.md
GregoireW: I rewrote a little bit this part.

rfcs/0005-managed-attributes-image-automation/README.md
GregoireW: I'm not found of regular capture group when the use of the capture is not in the same code space as the capture itself.   The definition is on the ImagePolicy then the use is on a comment of another kubernetes object. What's 1, 2, 3  in this context... not sure.   (I Am Not A Number, I Am A Free Man! )

rfcs/0005-managed-attributes-image-automation/README.md
GregoireW: I think there is 2 aspects to consider:
The simplicity would means to use `{namespace}:{policy}:{attribute}`  simple to read and to understand. I added a part on collision where a warning should be logged, but the value should remains as today norm.

The tag parts would be interesting if in the future multiple way to retrieve data would be permitted. For instance label from OCI container ...    if you then be interesting to split  "extracted attributes" to "manifest attributes".  Or perhaps it would simply be a configuration on the ImagePolicy like "attributes X = json path in the manifest" which would kept  Image Automation Controller simple.

rfcs/0005-managed-attributes-image-automation/README.md
GregoireW: I guess a warning in the log and no update.  BTW a capture producing empty attributes will need to be tested

GregoireW:
@dungdm93 

You mention you are interrested by the image digest.  Can you elaborate on your use case ?

PR #2231, opened 2021-12-17
Fix port overwrite for `flux bootstrap git`

achetronic:
Hello
This PR is related to [this one](https://github.com/fluxcd/flux2/pull/2216) and both them should be closed once this one is merged :)

**The problem:**
When you use the bootstrap command on Flux CLI using the generic git subcommand, the port is always removed from the URL that is written on gotk-sync.yaml. If you use a Git server on non-standard port this is a problem. Removing this line will keep the port untouched, when it is standard and when it is not.

**The proof:**
Screenshot from 2021-12-14 12-32-56

As you can see in the commits, I squashed previous ones but I can not clean it better now again because some others were merged in the middle and some of them are not shown in the git rebase -i
As far as I can not control the right behavior to upload the clean PR, I preferred to open this one without dust

### DIFF
--- a/cmd/flux/bootstrap_git.go
+++ b/cmd/flux/bootstrap_git.go
@@ -161,10 +161,15 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 			secretOpts.CAFilePath = bootstrapArgs.caFile
 		}
 
+		// Remove port of the given host when not syncing over HTTP/S to not assume port for protocol
+		// This _might_ be overwritten later on by e.g. --ssh-hostname
+		if repositoryURL.Scheme != "https" && repositoryURL.Scheme != "http" {
+			repositoryURL.Host = repositoryURL.Hostname()
+		}
+
 		// Configure repository URL to match auth config for sync.
 		repositoryURL.User = nil
 		repositoryURL.Scheme = "https"
-		repositoryURL.Host = repositoryURL.Hostname()
 	} else {
 		secretOpts.PrivateKeyAlgorithm = sourcesecret.PrivateKeyAlgorithm(bootstrapArgs.keyAlgorithm)
 		secretOpts.Password = gitArgs.password


cmd/flux/bootstrap_git.go
```diff
@@ -164,7 +164,6 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		// Configure repository URL to match auth config for sync.
 		repositoryURL.User = nil
 		repositoryURL.Scheme = "https"
-		repositoryURL.Host = repositoryURL.Hostname()
```
souleb, line 167: Thanks for your patience on this. One last thing to address @hiddeco comment
>> Given the scenario where someone configures a --url ssh://git@example.com:9999/repository.git to perform Git operations from their machine over SSH, but also sets --token-auth to enable HTTPS on the cluster side, it will now result in the SSH port being configured for the HTTPS URL.

I don't think that it's a common use case, but to preserve behavior, can you keep `repositoryURL.Host = repositoryURL.Hostname()` for non http schemes?

souleb: Sorry for being unclear. 

```go
if bootstrapArgs.tokenAuth {
  ...
  repositoryURL.Host = repositoryURL.Hostname()
}
```

to 
```go
if bootstrapArgs.tokenAuth {
  ...
  repositoryURL.Host = repositoryURL.Hostname()
    if repositoryURL.Scheme != "https" && repositoryURL.Scheme != "http" {
      repositoryURL.Host = repositoryURL.Hostname()
    } 
```

souleb (approved):
LGTM thanks @achetronic

achetronic:
Hello @souleb 
Fixed and tested. It seems to work right and keep the behavior in the same way :)

achetronic:
> LGTM thanks @achetronic

Let's wait the workflows approval. I would like to know, more or less, how long is it expected to be merged? To know when the fix will enter to our production systems :)

Thank you!

cmd/flux/bootstrap_git.go
```diff
@@ -164,7 +164,9 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
...
+		if repositoryURL.Scheme != "https" && repositoryURL.Scheme != "http" {
```
darkowlzz, line 167: Hi, this doesn't look right to me. The line above it sets the scheme and then immediately checks for the same value.
I may be wrong, but the issue seems to be more about not losing the extra port information in the url.
Maybe we can add a functions like:

```go
func hostnameWithPort(u *url.URL) string {
	if u.Port() == "" {
		return u.Hostname()
	}
	return fmt.Sprintf("%s:%s", u.Hostname(), u.Port())
}
```
to ensure that the port is used if it's available and using that to set `repositoryURL.Host` value.

Playground: https://go.dev/play/p/5lhwRw9vFvP

souleb: Sorry I missed that. The `if` case shall be before setting the scheme to `https`. We just want to make sure we don't keep an `ssh` port when we want to sync using `http`.

souleb (approved):

achetronic:
Hello guys @stefanprodan @hiddeco @fluxcdbot 
This merge is pending for your approval to merge :)

Thank you

cmd/flux/bootstrap_git.go
hiddeco: I feel this statement could really use a descriptive comment to help contributors (and ourselves) understand in the future what we are doing here, what effect it has, and why we are doing it. As I even have a hard time to grasp after 2.5 weeks of vacation.

hiddeco (approved):
Thank you @achetronic :bow:

Issue #2273, opened 2022-01-11
Fix envtest setup and usage


aryan9600:
We currently have a a target `setup-envtest`  which installs the `etcd`, `kubectl` , etc for users but we don’t use it implicitly in the `test` target. We also don't seem to set the `KUBEBUILDER_ASSETS` env to the dir where the binaries are installed so that they can be found during tests.

Projects:
* [x] flux2
* [x] pkg
* [x] source-controller
* [x] image-automation-controller
* [x] image-reflector-controller
* [x] notification-controller
* [x] source-watcher
* [x] kustomize-controller

stefanprodan:
Please use https://github.com/fluxcd/helm-controller/blob/main/Makefile as an example for all others. Note that `testbin` must be added to gitignore. Also `go get` has been deprecated for binaries, we should use `go install` starting with Go 1.17.

stefanprodan:
We need this for fluxcd/pkg project as well.

stefanprodan:
Thanks @aryan9600 for taking care of this 🥇

PR #2288, opened 2022-01-13
Fix makefile envtest setup and usage

aryan9600:
Refactor logic to install helper tools into one function in the
Makefile. Add support for envtest to help install tools like kubectl,
etcd which helps users run tests more conveniently.

Ref: https://github.com/fluxcd/flux2/issues/2273

Signed-off-by: Sanskar Jaiswal <jaiswalsanskar078@gmail.com>


stefanprodan (approved):
LGTM

Thanks @aryan9600

Issue #2014, opened 2021-10-27
Conform to the Kubernetes restricted pod security standard


dholbach:
We should make Flux conform to the [Kubernetes restricted pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) and update the documentation to reflect this.

Patch the deployment spec for:
- [x] source-controller
- [x] kustomize-controller
- [x] helm-controller
- [x] notification-controller
- [x] image-reflector-controller
- [x] image-automation-controller

From Ada Logics

> The deployments of the Flux controllers lack container security options that mitigate privilege escalation risks. These are hardening options that we recommend using, and Flux already
makes use of `allowPrivilegeEscalation: false` on all of its controllers. However, in addition to the `allowPrivilegeEscalation` option Flux could harden its containers by:
>
> - Dropping all Linux capabilities and enabling those needed
> - Filtering syscalls by way of Seccomp
>
> Docker drops many Linux capabilities by default but keeps others for convenience. Flux can harden its containers by having Docker drop all privileges a root user’s process can perform on a system and enabling only those needed. See [here](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) for details.
>
> Seccomp filtering is a way of limiting the available system calls and as of v1.19 Kubernetes has support for specifying seccomp policies through the use of `seccompProfile` in the `securityContext` of pods. Please see [here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp) for details.
>
> Recommendation
> Ensure that the pod deployed by Flux has appropriate hardening applied through the use of dropping Linux capabilities and syscall filtering.

gbstan92:
@stefanprodan the solution above does not cover the [registry credentials sync](https://github.com/fluxcd/flux2/tree/main/manifests/integrations/registry-credentials-sync/azure).  

Setting the Pod Security Policy in a similar way to the other components causes the following error: 

![image](https://user-images.githubusercontent.com/35976278/145177431-c1a9e73c-0a4f-4320-975b-236484d9adee.png)

pjbgf:
@kingdonb you can assign this one to me.

pjbgf:
> Yes, shipping a seccomp profile by default is not something that I would consider for Flux. Kubernetes has alpha support for it https://kubernetes.io/blog/2021/08/25/seccomp-default/

[...]

PR #2316, opened 2022-01-20
[security] Enable pod security warnings for flux-system

pjbgf:
Enable warnings for any violation of the `restrict` [Pod Security Standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/) at the `flux-system` namespace. This only works when [Pod Security Admission Controller](https://kubernetes.io/docs/concepts/security/pod-security-admission/) is installed and enabled, which is the default setting on Kubernetes `v1.23+`.

As of version `v0.26.0` all flux components should be compatible with above mentioned policy. This is not being enforced by default to allow backwards compatibility with users that decided deploying non-flux components into the `flux-system` namespace.

### DIFF
--- a/manifests/install/namespace.yaml
+++ b/manifests/install/namespace.yaml
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: flux-system
+  labels:
+    pod-security.kubernetes.io/warn: restricted
+    pod-security.kubernetes.io/warn-version: latest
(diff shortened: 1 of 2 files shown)


manifests/install/namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: flux-system
+    pod-security.kubernetes.io/warn: restricted
```
stefanprodan, line 5: Add these under `labels:`

stefanprodan:
Ref: https://github.com/fluxcd/flux2/issues/2014

manifests/install/namespace.yaml
pjbgf: This should be fixed now, thank you.

stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #2342, opened 2022-01-26
Run the CLI as non-root

stefanprodan:
Run the Flux CLI inside the container under the nobody user and group `65534:65534`.

Closes: #2014


darkowlzz (approved):
Thanks.

Issue #2308, opened 2022-01-19
Prepare Flux v0.26.0 release


stefanprodan:
This is an overarching issue for tracking the Flux v0.26.0 release preparation.

### Catalog and sign artifacts

- [x] https://github.com/fluxcd/flux2/issues/2302
- [x] https://github.com/fluxcd/flux2/issues/2303
- [x] fluxcd.io (document the signature verification of container images, binaries and SBOM)

### Security improvements

- [x] https://github.com/fluxcd/flux2/issues/2340
- [x] https://github.com/fluxcd/flux2/issues/2337
- [x] https://github.com/fluxcd/flux2/issues/2014
- [x] flux2 (drop support for EOL Kubernetes 1.19 in `flux check`)
- [x] fluxcd.io (document the pod security standard and update Kubernetes supported versions)

### Migrate from Debian to Alpine for `libgit2` static build

- [x] https://github.com/fluxcd/source-controller/pull/558
- [x] https://github.com/fluxcd/image-automation-controller/pull/303

### Fuzzing

- [x] https://github.com/fluxcd/pkg/pull/212
- [x] https://github.com/fluxcd/kustomize-controller/pull/536
- [x] https://github.com/fluxcd/helm-controller/pull/396
- [x] https://github.com/fluxcd/notification-controller/pull/306
- [x] https://github.com/fluxcd/image-reflector-controller/pull/221

### New features and improvements 

- [x] https://github.com/fluxcd/flux2/pull/2167
- [x] https://github.com/fluxcd/notification-controller/pull/317
- [x] https://github.com/fluxcd/image-reflector-controller/pull/194
- [x] https://github.com/fluxcd/image-automation-controller/pull/299
- [x] https://github.com/fluxcd/kustomize-controller/pull/538
- [x] https://github.com/fluxcd/kustomize-controller/pull/527

### Documentation

- [ ] https://github.com/fluxcd/website/pull/727
- [ ] https://github.com/fluxcd/website/pull/702
- [ ] https://github.com/fluxcd/website/pull/713

texasbobs:
Is there an ETA for this?  We are kind of stuck right now in our deployment process with Flux2 in test and Flux in prod.  We have been experiencing issues where a commit "works" when deployed in test with Flux2 yet fails in prod with Flux.  We're hoping to test the "revoke kubectl managed fields" change as soon as we can get this release.

texasbobs:
[...]

PR #2343, opened 2022-01-26
Set minimum supported version to Kubernetes 1.20.6

stefanprodan:
Drop support for the Kubernetes versions that have reached end-of-life. Given that Kubernetes between 1.20.0 and 1.20.5 had [a bug](https://github.com/kubernetes/kubernetes/pull/98576) with server-side apply, the `flux check` command constraint has been update to Kubernetes >= 1.20.6.

Part of: https://github.com/fluxcd/flux2/issues/2308

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -27,7 +27,7 @@ jobs:
         uses: engineerd/setup-kind@v0.5.0
         with:
           version: v0.11.1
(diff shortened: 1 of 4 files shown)


darkowlzz (approved):
LGTM!

darkowlzz:
Is there a need to update the contributing doc's kubectl version? https://github.com/fluxcd/flux2/blob/main/CONTRIBUTING.md?plain=1#L71

stefanprodan:
@darkowlzz good point, I also update the Go version.

PR #2370, opened 2022-02-02
Add test case on hpa metrics for flux build/diff

souleb:
If implemented will detect any change on how ssa dry-run return diffs on
hpa metrics

### DIFF
--- a/cmd/flux/testdata/diff-kustomization/diff-with-drifted-hpa.golden
+++ b/cmd/flux/testdata/diff-kustomization/diff-with-drifted-hpa.golden
@@ -1,0 +1,52 @@
+► Deployment/default/podinfo created
+► HorizontalPodAutoscaler/default/podinfo drifted
+
+spec.behavior.scaleDown.policies.0.periodSeconds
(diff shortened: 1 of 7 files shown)


stefanprodan:
Can you please add a custom metric to the test:
```
  metrics:
    - pods:
        metric:
          name: nginx_http_requests_total
        target:
          averageValue: "4"
          type: AverageValue
      type: Pods
```

This was tripping diff in my tests, when other things change this custom metric gets duplicated in the diff output.

cmd/flux/testdata/diff-kustomization/diff-with-drifted-hpa.golden
```diff
@@ -0,0 +1,52 @@
...
+          type: AverageValue
+          averageValue: 3
+    - type: Resource
+      resource:
+        name: cpu
+        target:
+          type: Utilization
+          averageUtilization: 98
+  
+  + two list entries added:
```
stefanprodan, line 34: IMO this shouldn't happen.

PR #2380, opened 2022-02-04
Fix panic on bootstrap when orgRef is not retrieved

souleb:
If implemented, not retrieving an orgRef will always return an error


stefanprodan (approved):
LGTM

Thanks @souleb

PR #2382, opened 2022-02-04
Use `client.Patch` for suspend/resume operations

SomtochiAma:


stefanprodan (approved):
LGTM

This change will prevent managed fields conflicts between Flux the CLI and the Flux controllers.

Thanks @SomtochiAma 🏅

PR #2389, opened 2022-02-04
Fix wrong deletion message on flux diff

souleb:
If implemented, when error happens in dry-run and we have pruning
enforced, the deletion message will be consistent


stefanprodan (approved):
LGTM

Thanks @souleb

PR #2425, opened 2022-02-11
Add a simple spinner when running flux diff kustomization

souleb:
If implemented, users will see a spinner run while the diff is on-going.


go.mod
```diff
@@ -50,6 +50,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/theckman/yacspin v0.13.12
```
stefanprodan, line 53: Please move this to the first block above.

stefanprodan (approved):
LGTM 

Thanks @souleb

PR #2440, opened 2022-02-15
Add e2e sequence diagrams

stefanprodan:
Add end-to-end sequence diagrams made with Mermaid for:
- Cluster sync from Git
- Helm release upgrade from Git
- Image update to Git

Ref: https://github.com/fluxcd/website/pull/725

flux-cluster-sync
flux-helm-upgrade
flux-image-update

### DIFF
--- a/docs/internal/diagrams.md
+++ b/docs/internal/diagrams.md
@@ -1,0 +1,83 @@
+# Flux Diagrams
+
+## Cluster sync from Git
+
+```mermaid
+sequenceDiagram
+    actor me
+    participant git as Git<br><br>repository
+    participant sc as Flux<br><br>source-controller
+    participant kc as Flux<br><br>kustomize-controller
+    participant kube as Kubernetes<br><br>api-server
+    participant nc as Flux<br><br>notification-controller
+    me->>git: 1. git push
+    sc->>git: 2. git pull
+    sc->>sc: 3. build artifact for revision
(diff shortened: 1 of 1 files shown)


makkes (approved):
I love visualizations! Those will help people grasp Flux much quicker. Don't we want to include those in the official docs, too?

stefanprodan:
@makkes I made these for https://github.com/fluxcd/website/pull/725, @kingdonb we'll have to figure out how to enable Mermaid rendering with Hugo.

mbovo:
Sorry to inject in to the converstaion, I think you should add the mermaid.js to the hugo template, I'll try to prepare a PR (as soon as find where that code lives :D )
```
<script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
<script>mermaid.initialize({ startOnLoad: true, securityLevel: 'loose'}});</script>
```

kranurag7:
@stefanprodan These visualizations are really superb.

kingdonb:
Awesome, thank you! This will make a great addition to the E2E doc 👍

stefanprodan:
@kingdonb looks like Hugo has a shortcode for mermaid https://learn.netlify.app/en/shortcodes/mermaid/ So instead of using code blocks like in here we could use `{{< mermaid >}}` for the website.

PR #2443, opened 2022-02-16
Validate that object name adheres to RFC 1123 for `flux create` commands

SomtochiAma:
This pull request validates the name passed to `flux create` commands matches the [RFC 1123 Label Names].(https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)

### DIFF
--- a/cmd/flux/create.go
+++ b/cmd/flux/create.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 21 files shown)


cmd/flux/create.go
```diff
@@ -51,6 +52,18 @@ func init() {
...
 	createCmd.PersistentFlags().StringSliceVar(&createArgs.labels, "label", nil,
 		"set labels on the resource (can specify multiple labels with commas: label1=value1,label2=value2)")
+	createCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			return fmt.Errorf("name is required")
+		}
+
+		name := args[0]
+		if !validateObjectName(name) {
+			return fmt.Errorf("name '%s' is invalid, it should adhere to standard defined in RFC 1123", name)
```
stefanprodan, line 62:
```suggestion
			return fmt.Errorf("name '%s' is invalid, it should adhere to standard defined in RFC 1123. The name can contain only lowercase alphanumeric characters or '-'", name)
```

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🏅

PR #2448, opened 2022-02-17
Use `metadata.Client` for reconcile operations

SomtochiAma:

### DIFF
--- a/cmd/flux/helmrelease.go
+++ b/cmd/flux/helmrelease.go
@@ -25,8 +25,9 @@ import (
 // helmv2.HelmRelease
 
 var helmReleaseType = apiType{
-	kind:      helmv2.HelmReleaseKind,
-	humanKind: "helmreleases",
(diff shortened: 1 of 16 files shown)


cmd/flux/helmrelease.go
```diff
@@ -25,8 +25,9 @@ import (
 // helmv2.HelmRelease
 
 var helmReleaseType = apiType{
-	kind:      helmv2.HelmReleaseKind,
-	humanKind: "helmreleases",
+	kind:         helmv2.HelmReleaseKind,
+	humanKind:    "helmreleases",
```
stefanprodan, line 29: What's the reason this is plural while all the other are singular?

stefanprodan: Also kustomizations

stefanprodan (approved):
LGTM

Thanks @SomtochiAma

PR #2452, opened 2022-02-18
Diff: fix stringData Secret issue

souleb:
This commit migrate to the last version of pkg/ssa v0.14.1 that contains a fix
for stringData secrets. The test case was changed accordingly to
    validate a stringData drift.

A progress-bar flag option has also been added in order to be able to
disable it.


stefanprodan (approved):
LGTM

Thanks @souleb

Issue #2385, opened 2022-02-04
Move Message/Status to the end of kubectl/flux get


schrej:
Currently the Message (flux) and Status (kubectl) columns are displayed in the middle of the output table of the `kubectl/flux get` commands. As they can be very long in error cases, it's often hard to find the remaining columns, especially for resources that don't contain an error.

To improve visibility, I'd suggest to move the Message/Status columns to the end of the output tables. That way the output is a lot easier to read, especially on smaller screens.

stefanprodan:
We are changing the status message for Flux GitRepository objects, in the next version, source-controller will include the commit message making the output even lengthier. Moving the status message to the last column would improve the readability.

PR #2467, opened 2022-02-23
Update `get` subcommand column order

darkowlzz:
- Move `READY` and `MESSAGE` to the end of `get` subcommand output.
- Add test flag `-update` to update the golden files.
- Update the golden files based on the output change using.

Fixes https://github.com/fluxcd/flux2/issues/2385

Associated PRs:
- https://github.com/fluxcd/source-controller/pull/592
- https://github.com/fluxcd/kustomize-controller/pull/578
- https://github.com/fluxcd/helm-controller/pull/425
- https://github.com/fluxcd/notification-controller/pull/337

### DIFF
--- a/cmd/flux/main_test.go
+++ b/cmd/flux/main_test.go
@@ -304,10 +304,15 @@ func assertGoldenTemplateFile(goldenFile string, templateValues map[string]strin
 		if assertErr := assertGoldenValue(expectedOutput)(output, err); assertErr != nil {
 			// Update the golden files if comparision fails and the update flag is set.
 			if *update && output != "" {
-				if err := os.WriteFile(goldenFile, []byte(output), 0644); err != nil {
-					return fmt.Errorf("failed to update golden file '%s': %v", goldenFile, err)
+				// Skip update if there are template values.
+				if len(templateValues) > 0 {
+					fmt.Println("NOTE: -update flag passed but golden template files can't be updated, please update it manually")
+				} else {
+					if err := os.WriteFile(goldenFile, []byte(output), 0644); err != nil {
+						return fmt.Errorf("failed to update golden file '%s': %v", goldenFile, err)
+					}
+					return nil
 				}
-				return nil
 			}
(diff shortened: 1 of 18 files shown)


cmd/flux/main_test.go
```diff
@@ -298,6 +302,13 @@ func assertGoldenTemplateFile(goldenFile string, templateValues map[string]strin
 			expectedOutput = string(goldenFileContents)
 		}
 		if assertErr := assertGoldenValue(expectedOutput)(output, err); assertErr != nil {
+			// Update the golden files if comparision fails and the update flag is set.
+			if *update && output != "" {
+				if err := os.WriteFile(goldenFile, []byte(output), 0644); err != nil {
```
stefanprodan, line 307: Wouldn't this remove the variables declared inside golden files such as https://github.com/fluxcd/flux2/blob/437a7a2852976c037d34915d325d9fb5288ebd77/cmd/flux/testdata/trace/deployment.golden#L9

darkowlzz: Yeah, that'll update the golden template files as well. Didn't realize that the golden files are in template form.
After going through the code again, it looks like we can determine when it's a golden template file based on the `templateValues` passed to `assertGoldenTemplateFile()`.
I've added a check for that. Now when there's a diff in the output **and** `-update` is passed, a note is printed about it.

```console
NOTE: -update flag passed but golden template file can't be updated, please update it manually
    main_test.go:351: Mismatch from golden file 'testdata/trace/helmrelease.golden': Mismatch from expected value (-want +got):
```

Since not all the packages use golden files and won't work with `-update` flag, I made some changes in the Makefile to make it easy to set the target test package and documented how to do it in the CONTRIBUTING.md file.
Does it look good?

stefanprodan: Great! thank you

stefanprodan (approved):
LGTM

Thanks @darkowlzz

PR #2471, opened 2022-02-24
feat: add linux/s390x builds

skuethe:
Signed-off-by: skuethe <56306041+skuethe@users.noreply.github.com>

This PR adds the `s390x` platform to build and release automation. As multi arch builds are setup perfectly already, the necessary changes are minimal.
A (stripped down) successful version of the github action can be seen here:
https://github.com/skuethe/flux2/actions/runs/1893656250

I have thought about also adapting e2e tests for s390x, but that seems (at least at this point) out of the question, as there are no s390x instances on equinix:
https://metal.equinix.com/developers/docs/servers/about/#architecture

I hope opening this PR directly (without a referencing issue) is okay, I did not find anything related to that in your CONTRIBUTING.md

Will look into adding s390x support on the GitOps Toolkit components in the next days.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,6 +14,7 @@ builds:
       - amd64
       - arm64
       - arm
+      - s390x
     goarm:
       - 7
   - <<: *build_defaults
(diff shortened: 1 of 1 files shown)


skuethe:
I have created a discussion about s390x support in general:
https://github.com/fluxcd/flux2/discussions/2475

stefanprodan:
Closing this since we don't have a solution for libgit2.

PR #2512, opened 2022-03-07
Introduce a printer interface for flux resources

souleb:
If implemented, there will a common interface to print flux resource.

We are adding new way to print resource information e.g. diff of objects (and unified diff in the future).

This will make it easier to add new printers for json, jsonpath, etc... when the need arise.

### DIFF
--- a/pkg/printers/interface.go
+++ b/pkg/printers/interface.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


stefanprodan (approved):
LGTM

Thanks @souleb

pkg/printers/interface.go
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package printers
+
+import "io"
+
+// Printer is an interface for printing Flux cmd	outputs.
```
hiddeco, line 21:
```suggestion
// Printer is an interface for printing Flux cmd outputs.
```

PR #2559, opened 2022-03-22
Retry bootstrap operations on Git conflict errors

stefanprodan:
When running bootstrap in-parallel for many clusters that target the same repository, the 2nd commit with the sync files fails with ` non-fast-forward update`. We now detect the conflict, and we retry the operations by creating a fresh clone from upstream.

### DIFF
--- a/internal/bootstrap/bootstrap_plain_git.go
+++ b/internal/bootstrap/bootstrap_plain_git.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	gogit "github.com/go-git/go-git/v5"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


internal/bootstrap/bootstrap_plain_git.go
```diff
@@ -287,10 +288,28 @@ func (b *PlainGitBootstrapper) ReconcileSyncConfig(ctx context.Context, options
...
 		b.logger.Actionf("pushing sync manifests to %q", b.url)
-		if err = b.git.Push(ctx, b.caBundle); err != nil {
+		err = b.git.Push(ctx, b.caBundle)
+		if err != nil {
+			if strings.HasPrefix(err.Error(), gogit.ErrNonFastForwardUpdate.Error()) {
+				b.logger.Warningf("git conflict detected, retrying with a fresh clone")
+				if err := os.RemoveAll(b.git.Path()); err != nil {
+					return fmt.Errorf("failed to remove tmp dir: %w", err)
+				}
+				if err := os.Mkdir(b.git.Path(), 0700); err != nil {
```
darkowlzz, line 302:
```suggestion
				if err := os.Mkdir(b.git.Path(), 0o700); err != nil {
```
since we've started using octal notation in other repos.

stefanprodan: I changed it, but Go itself doesn't not uses octal notation, why are we doing this?

https://github.com/golang/go/blob/334a591a3f4d868368913328b3e81ddf5b0f46fa/src/os/tempfile.go#L94

stefanprodan: Does the octal notation works on Windows?

darkowlzz: I don't have a windows environment, but based on this https://github.com/golang/go/blob/96567fb3ccdc8d8c052606112db7b39abd725b59/src/os/types.go#L23-L28 I'd assume it works.
> // A FileMode represents a file's mode and permission bits.
> // The bits have the same definition on all systems, so that
> // information about files can be moved from one system
> // to another portably. Not all bits apply to all systems.

stefanprodan: Thanks @pjbgf for the explanations

PR #2564, opened 2022-03-23
Use absolute domain name for the events address

stefanprodan:
Add ending dot to the events address to be consistent with source controller address.
This will affect `flux bootstrap` and `flux install` by setting `--events-addr=http://notification-controller.flux-system.svc.cluster.local./`.

### DIFF
--- a/pkg/manifestgen/install/manifests.go
+++ b/pkg/manifestgen/install/manifests.go
@@ -73,7 +73,7 @@ func generate(base string, options Options) error {
 		// traffic from going through http proxy. Without fully specified
 		// domain they need to mention `notifications-controller` explicitly in
 		// `no_proxy` variable after debugging http proxy logs.
-		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.%s/", options.NotificationController, options.Namespace, options.ClusterDomain)
+		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.%s./", options.NotificationController, options.Namespace, options.ClusterDomain)
 	}
 
 	if err := execTemplate(options, namespaceTmpl, path.Join(base, "namespace.yaml")); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/manifestgen/install/manifests.go
```diff
@@ -73,7 +73,7 @@ func generate(base string, options Options) error {
 		// traffic from going through http proxy. Without fully specified
 		// domain they need to mention `notifications-controller` explicitly in
 		// `no_proxy` variable after debugging http proxy logs.
-		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.%s/", options.NotificationController, options.Namespace, options.ClusterDomain)
+		options.EventsAddr = fmt.Sprintf("http://%s.%s.svc.%s./", options.NotificationController, options.Namespace, options.ClusterDomain)
```
hiddeco, line 76: Wouldn't it be better to make the ending `.` part of the `ClusterDomain` option? In this way, people can still control _not_ adding a dot using this option.

stefanprodan: That's breaking change to SC.

hiddeco: How is that breaking to SC? The controller is configured using the full domain as `--storage-adv-addr` string.

stefanprodan: `ClusterDomain` is used by SC, if we remove the dot and we add it to the default value, then we'll break Flux for all those that have set a custom domain, e.g. GKE private.

hiddeco (approved):

pkg/manifestgen/install/manifests.go
pjbgf: I would be for adding the dot in `options.ClusterDomain`. And here a validation check that adds one if there isn't one already - to avoid double dots.

hiddeco: The problem is with people having configured a custom domain, assuming a dot will be added. Which would no longer be added if we change it, which may cause issues (but likely only results in an additional DNS lookup).

PR #2565, opened 2022-03-23
fix: wait for Source objects observed generation

hiddeco:
This ensures the command will wait for the object to report a Ready
Condition with an ObservedGeneration matching the Generation of the
resource. Ensuring that when a "create" is actually a mutation, it waits
instead of prematurely assuming the Source to be Ready.


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #2582, opened 2022-03-28
upgrade kube-prometheus-stack to 34.5.0

kingdonb:
The kube-prometheus-stack chart version 32.2.0 appears to change the way that dashboard configs are labeled by the chart, I do not know what is the purpose of this change yet, however it causes our Flux dashboards to fail to be included.

It should be safe to upgrade to 32.1.0, (although I'd like to find the issue and resolve it here, so we can just be current again)

---

OK, the chart is upgraded here to the latest version, and the custom dashboards we included are tested to still work. 👍

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/cyphar/filepath-securejoin v0.2.3
 	github.com/fluxcd/go-git-providers v0.5.4
 	github.com/fluxcd/helm-controller/api v0.18.2
-	github.com/fluxcd/image-automation-controller/api v0.21.1
+	github.com/fluxcd/image-automation-controller/api v0.21.2
 	github.com/fluxcd/image-reflector-controller/api v0.17.1
 	github.com/fluxcd/kustomize-controller/api v0.22.2
 	github.com/fluxcd/notification-controller/api v0.23.1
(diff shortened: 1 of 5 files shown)


manifests/monitoring/kube-prometheus-stack/release.yaml
```diff
@@ -22,6 +22,8 @@ spec:
     grafana:
       sidecar:
         dashboards:
+          #label: grafana_dashboard
```
stefanprodan, line 25: Please remove these if they are not needed

kingdonb: OK, squashed. 👍

kingdonb:
I'd like to preserve the two commits, so it's clear enough to anyone who reads the history, what broke and how was it fixed.

(I'll add one more that bumps the version to the latest, as soon as I've tested that also still works.)

kingdonb:
👍 LGTM

kingdonb:
We could do that, but this is not the first time we've seen breaking changes in the kube-prometheus-stack chart. I'm not sure if it's better to expose the breakages sooner, or to guarantee the chart is pinned at a working version.

kingdonb:
Not sure why this closed, but I have reopened it as #2587

PR #2597, opened 2022-03-30
[RFC-0002] Flux OCI support for Helm

stefanprodan:
Given that Helm v3.8 supports [OCI](https://helm.sh/docs/topics/registries/) for package distribution, we should extend the Flux Source API to allow fetching Helm charts from container registries.

This proposal is for adding OCI support with minimal API changes to Flux.

### DIFF
--- a/rfcs/helm-oci/README.md
+++ b/rfcs/helm-oci/README.md
@@ -75,7 +75,7 @@ And finally in Flux `HelmReleases`, refer to the ghcr-charts `HelmRepository`:
 apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
-  name: podinfo
+  name: my-app
   namespace: default
 spec:
   interval: 60m
@@ -98,9 +98,18 @@ spec:
 Given that charts are stored in container registries, you can use Flux image automation
 and patch the chart version in Git, in the same way Flux works for updating container image tags.
 
-Define an image policy using semver:
+Define an image registry and a policy for the chart artifact:
 
 ```yaml
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImageRepository
+metadata:
+  name: my-app
+  namespace: default
+spec:
+  image: ghcr.io/my-org/charts/my-app
+  interval: 1m0s
+---
 apiVersion: image.toolkit.fluxcd.io/v1beta1
 kind: ImagePolicy
 metadata:
@@ -120,7 +129,7 @@ Then add the policy marker to the `HelmRelease` manifests in Git:
 apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
-  name: podinfo
+  name: my-app
   namespace: default
 spec:
   interval: 60m
--- a/rfcs/helm-oci/README.md
+++ b/rfcs/helm-oci/README.md
@@ -1,0 +1,110 @@
+# RFC-xxxx Flux OCI support for Helm
+
+**Status:** provisional
+
+**Creation date:** 2022-03-30
+
+**Last update:** 2022-03-30
+
+## Summary
+
+Given that Helm v3.8 supports [OCI](https://helm.sh/docs/topics/registries/) for package distribution,
+we should extend the Flux Source API to allow fetching Helm charts from container registries.
+
+## Motivation
+
+Helm OCI support is one of the most requested feature in Flux
+as seen on this [issue](https://github.com/fluxcd/source-controller/issues/124).
+
+### Goals
+
+- Add support for fetching Helm charts stored as OCI artifacts with minimal API changes to Flux.
+- Make it easy for users to switch from HTTP/S Helm repositories to OCI repositories.
+
+### Non-Goals
+
+- Introduce a new API kind for referencing charts stored as OCI artifacts.
+
+## Proposal
+
+Introduce an optional field called `type` to the `HelmRepository` spec.
+
+When not specified, the `spec.type` field defaults to `Default` which preserve the current `HelmRepository` API behaviour.
+
+When the `spec.type` field is set to `OCI`, the `spec.url` field must be prefixed with `oci://` (to follow the Helm conventions).
+For `oci://` URLs, source-controller will use the Helm SDK and the `oras` library to connect to the OCI remote storage.
+For authentication, the controller will use Kubernetes secrets of `kubernetes.io/dockerconfigjson` type.
+
+### User Stories
+
+#### Story 1
+
+> As a developer I want to use Flux `HelmReleases` that refer to Helm charts stored
+> as OCI artifacts in GitHub Container Registry.
+
+First create a secret using a GitHub token that allows access to GHCR:
+
+```sh
+kubectl create secret docker-registry ghcr-charts \
+    --docker-server=ghcr.io \
+    --docker-username=$GITHUB_USER \
+    --docker-password=$GITHUB_TOKEN
+```
+
+Then define a `HelmRepository` of type `OCI` and reference the `dockerconfig` secret:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: HelmRepository
+metadata:
+  name: ghcr-charts
+  namespace: default
+spec:
+  type: OCI
+  url: oci://ghcr.io/my-org/charts/
+  secretRef:
+    name: ghcr-charts
+```
+
+And finally in Flux `HelmReleases`, refer to the ghcr-charts `HelmRepository`:
+
+```yaml
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: podinfo
+  namespace: default
+spec:
+  interval: 60m
+  chart:
+    spec:
+      chart: my-app
+      version: '1.0.x'
+      sourceRef:
+        kind: HelmRepository
+        name: ghcr-charts
+      interval: 1m # check for new OCI artifacts every minute
+```
+
+### Alternatives
+
+We could use introduce a new API type e.g. `HelmOCIRepository`. That is considered unpractical,
+as there is no benefit for users in having a dedicated kind instead of a `type` filed in the current
+`HelmRepository` API. Adding a `type` filed to the spec follows the Flux Bucket API design, where 
+the same Kind servers different implementations: AWS S3 vs Azure Blob vs Google Storage.
+
+## Design Details
+
+In source-controller we'll add a new predicate for indexing `HelmRepositories` based on the `spec.type` field.
+
+When the `spec.type` field is set to `OCI`, the `HelmRepositoryReconciler`
+will set the `HelmRepository` Ready status to `False` if the URL is not prefixed with `oci://`,
+otherwise the Ready status will be set to `True`.
+
+The current `HelmChartReconciler` will use the `HelmRepositories` with `type: Default`.
+For `type: OCI` we'll introduce a new reconciler `HelmChartOCIReconciler` that uses `oras` to download charts
+and their dependencies.
+
+### Enabling the feature
+
+The feature is enabled by default.
--- a/rfcs/helm-oci/README.md
+++ b/rfcs/helm-oci/README.md
@@ -22,7 +22,8 @@ they do today for container images.
 ### Goals
 
 - Add support for fetching Helm charts stored as OCI artifacts with minimal API changes to Flux.
-- Make it easy for users to switch from HTTP/S Helm repositories to OCI repositories.
+- Make it easy for users to switch from [HTTP/S Helm repositories](https://github.com/helm/helm-www/blob/416fabea6ffab8dc156b6a0c5eb5e8df5f5ef7dc/content/en/docs/topics/chart_repository.md)
+  to OCI repositories.
 
 ### Non-Goals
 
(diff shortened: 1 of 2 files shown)


rfcs/helm-oci/README.md
```diff
@@ -86,9 +89,54 @@ spec:
...
+      sourceRef:
+        kind: HelmRepository
+        name: ghcr-charts
+      interval: 1m
+```
+
 ### Alternatives
 
-We could use introduce a new API type e.g. `HelmOCIRepository`. That is considered unpractical,
+We could introduce a new API type e.g. `HelmOCIRepository`. That is considered unpractical,
```
hiddeco, line 139: I would at least expect an outline of the rejected previous structure, in which an `OCIRegistry` would be introduced, to be included here.

stefanprodan: Add a link to the previous RFC and outlined the proposal there.

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: HelmRepository
+metadata:
+  name: ghcr-charts
+  namespace: default
+spec:
+  type: OCI
+  url: oci://ghcr.io/my-org/charts/
```
wonderflow, line 67: do we still need to type as we can get the inforamtion from url

wonderflow (approved):
LGTM, looking forward to integrating this feature into KubeVela!

rfcs/helm-oci/README.md
stefanprodan: Yes, we need a dedicated field that can be used as a custom index for the controller runtime predicate. This is explained in the "Design Details" section.

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+We could introduce a new API type e.g. `HelmRegistry` to hold the reference to auth secret,
+as proposed in [#2573](https://github.com/fluxcd/flux2/pull/2573).
+That is considered unpractical, as there is no benefit for users in having a dedicated kind instead of
+a `type` filed in the current `HelmRepository` API. Adding a `type` filed to the spec follows the Flux
+Bucket API design, where the same Kind servers different implementations: AWS S3 vs Azure Blob vs Google Storage.
+
+## Design Details
+
+In source-controller we'll add a new predicate for indexing `HelmRepositories` based on the `spec.type` field.
```
hiddeco, line 147: A predicate does not index, but [filters incoming events](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate).

Given this, we would introduce a Predicate that is capable of filtering `HelmRepository` objects based on their `.spec.type`. Allowing a reconciler to handle only `HelmRepository` objects with a single type, allowing a full split between `Default` and `OCI` handling.

With regards to indexing, it might be required to exclude OCI types or handle them differently within the `HelmChart` code while using them to fetch credentials for chart dependencies. But this logic is handled using an indexer func (and can be modified or altered as we deem necessary).

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+When the `spec.type` field is set to `OCI`, the `HelmRepositoryReconciler`
+will set the `HelmRepository` Ready status to `False` if the URL is not prefixed with `oci://`,
+otherwise the Ready status will be set to `True`.
```
hiddeco, line 151: It would be better to introduce a new reconciler for `OCI` types only using the predicate I mentioned earlier, keeping the code (paths) fully separated and simpler to grasp.

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+```yaml
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImagePolicy
+metadata:
+  name: my-app
+  namespace: default
+spec:
+  imageRepositoryRef:
+    name: my-app
```
relu, line 111: A missing implementation detail here: I think this will need to be switched over to `NamespacedObjectKindReference` with the default `kind` attribute set to `ImageRepository`.

stefanprodan: No, people will still need to use image automation as it is.

relu: Of course, but I don't see how it would work otherwise since the IRC reconciler will look only for an `ImageRepository` named `my-app`. I assumed `my-app` in from this example would be a `type: OCI` `HelmRepository`. Would the user need to create a separate `ImageRepository` resource to be able to use this feature? That's not very clear from the current spec details.

stefanprodan: Ok let me add all the YAMLs.

stefanprodan: Please check now and let me know if it's better.

relu: Yes, now it is clear enough. Thank you!

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+        name: ghcr-charts
+      interval: 1m
+```
+
+### Alternatives
+
+We could introduce a new API type e.g. `HelmRegistry` to hold the reference to auth secret,
+as proposed in [#2573](https://github.com/fluxcd/flux2/pull/2573).
+That is considered unpractical, as there is no benefit for users in having a dedicated kind instead of
+a `type` filed in the current `HelmRepository` API. Adding a `type` filed to the spec follows the Flux
```
darkowlzz, line 151:
```suggestion
a `type` field in the current `HelmRepository` API. Adding a `type` field to the spec follows the Flux
```

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Introduce a new API kind for referencing charts stored as OCI artifacts.
+
+## Proposal
+
+Introduce an optional field called `type` to the `HelmRepository` spec.
+
+When not specified, the `spec.type` field defaults to `Default` which preserve the current `HelmRepository` API behaviour.
+
+When the `spec.type` field is set to `OCI`, the `spec.url` field must be prefixed with `oci://` (to follow the Helm conventions).
```
souleb, line 37: If the user have to set the type to `OCI` for a helm repository, then there is no need to prefix the `spec.url` field. Helm is using to see where to pull a chart from, that is the `type` role here.

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+In source-controller we'll add a new predicate for filtering `HelmRepositories` based on the `spec.type` field.
+
+The current `HelmRepositoryReconciler` will be renamed to `HelmRepositoryDefaultReconciler`,
+it's scope remains unchanged, and it will handle only objects with `type: Default`.
+
+We'll introduce a new reconciler named `HelmRepositoryOCIReconciler`, that will handle
+objects with `type: OCI`. This reconciler will set the `HelmRepository` Ready status to
+`False` if the URL is not prefixed with `oci://`, otherwise the Ready status will be set to `True`.
+
+The current `HelmChartReconciler` will be renamed to `HelmChartDefaultReconciler`,
```
souleb, line 165: The current `HelmChartReconciler` reconcile charts from multiple sources (gitrepo, helmrepo), why should we have a different reconciler for `type: OCI`? We could reuse most of the code for fetching charts from a remote helm repository with a check on the `getter` type and then branching to a function that would pull the chart from the  OCI registry.

stefanprodan: I made this change at @hiddeco’s request, IMO dedicated reconcilers are easier to reason with.

rfcs/helm-oci/README.md
stefanprodan: A URL by definition needs a protocol, it makes things easier to spot especially for `flux get` where we don’t want to add yet another column, users will determine the source type from displayed URL.

hiddeco: If this results in too much duplicate code, I am happy to not split on `HelmChart` level. The `HelmRepository` split is however a must, imho.

rfcs/helm-oci/README.md
hiddeco: If this results in too much duplicate code, I am happy to not split on `HelmChart` level. The `HelmRepository` split is however a must, imho.

stefanprodan: I think we can leave this as a suggestion in here and figure out the right way when we review the implementation PR.

rfcs/helm-oci/README.md
RedbackThomson, line 37: Should the type be `OCIRegistry`, rather than just `OCI` - since we want to identify that the URL is reflecting a registry and not just a single image within it?

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+> with the version set in the `HelmRelease` manifests.
+
+Given that charts are stored in container registries, you can use Flux image automation
+and patch the chart version in Git, in the same way Flux works for updating container image tags.
+
+Define an image registry and a policy for the chart artifact:
+
+```yaml
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImageRepository
```
RedbackThomson, line 105: This CRD does not need a `type: OCI`?

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+We could introduce a new API type e.g. `HelmRegistry` to hold the reference to auth secret,
+as proposed in [#2573](https://github.com/fluxcd/flux2/pull/2573).
+That is considered unpractical, as there is no benefit for users in having a dedicated kind instead of
+a `type` field in the current `HelmRepository` API. Adding a `type` field to the spec follows the Flux
+Bucket API design, where the same Kind servers different implementations: AWS S3 vs Azure Blob vs Google Storage.
+
+## Design Details
+
+In source-controller we'll add a new predicate for filtering `HelmRepositories` based on the `spec.type` field.
```
RedbackThomson, line 156: Why does this require we add a new field? Could the predicate not just check for the prefix string `oci://`?

hiddeco: Prefix checking is brittle, and protocol checking would require URL parsing on every queue request. Having a dedicated (enum) field is thus the better option.

rfcs/helm-oci/README.md
stefanprodan: Container images can only be in an OCI registry, this API is not new, see https://fluxcd.io/docs/components/image/imagerepositories/

rfcs/helm-oci/README.md
RedbackThomson: I really feel that inferring the option from the user input would lead to a much simpler user experience (with less chance for users to mismatch the types). Parsing the URL is a fairly cheap string comparison (right?) and shouldn't practically increase the reconciliation time for any resource. If it does become too brittle or has unintended consequences, then it will be easier to add the `spec.type` field at that time - rather than making that decision now and locking into that choice.

rfcs/helm-oci/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Proposal
+
+Introduce an optional field called `type` to the `HelmRepository` spec.
+
+When not specified, the `spec.type` field defaults to `Default` which preserve the current `HelmRepository` API behaviour.
+
+When the `spec.type` field is set to `OCI`, the `spec.url` field must be prefixed with `oci://` (to follow the Helm conventions).
+For `oci://` URLs, source-controller will use the Helm SDK and the `oras` library to connect to the OCI remote storage.
+For authentication, the controller will use Kubernetes secrets of `kubernetes.io/dockerconfigjson` type.
```
makkes, line 39: We should mention [automatic authentication](https://fluxcd.io/docs/components/image/imagerepositories/#automatic-authentication) here which IMHO should be supported the same way as it's documented there.

makkes (approved):
I generally like this RFC, thanks @stefanprodan! I figure a HelmRepository of type OCI wouldn't expose any artifact, right? It would rather server exclusively as a data store for URL and optionally credentials. I think that's absolutely fine, just want to make sure I understand it correctly and that we document that in the RFC.

rfcs/helm-oci/README.md
makkes: Another benefit of a dedicated field is the simpler consumption by external tooling (like kubectl + JSONPath).

hiddeco (approved):
Happy with how this slight shift in the proposal has resulted in a workable solution. LGTM, thanks @stefanprodan :bow:

makkes:
btw, I would love to update the [current PoC implementation](https://github.com/souleb/source-controller/tree/oci-registry) @souleb and I came up with to match what's defined in this RFC.

pjbgf (approved):
@stefanprodan great stuff, LGTM!

rfcs/helm-oci/README.md
stefanprodan: IMO fields can not replace documentation no matter how verbose we make them. The `type` field will be document in code, then it will show up at `kubectl explain helmrepository.spec.type`.

darkowlzz (approved):
LGTM!

relu (approved):
Looks solid! 🙌

phillebaba (approved):
Looks good. Just a small NIT, there may be a better value for the default type than `Default` as it wont really be too descriptive what this means if more types are added in the future.

rashedkvm (approved):
This proposal looks great. Nicely done @stefanprodan

makkes:
According to [our RFC process](https://github.com/fluxcd/flux2/tree/8b989190c4c2b8d9a6ce6c7591681f6610df419b/rfcs#rfc-process) I stepped in as sponsor for this RFC and given we have approval from at least 2 maintainers there's unanimity to merge this.

@stefanprodan this RFC's assigned number is 2.

stefanprodan:
The implementation of this RFC can be tracked here https://github.com/fluxcd/source-controller/issues/669

scottrigby (commented):
Late to the party here. Excited to see this maintainable RFC ❤️

Issue #251, opened 2020-09-22
git clone error: ssh: handshake failed: ssh: unable to authenticate


yawboateng:
am using an existing repo, but getting a handshake failed error even though i see commits with the yaml files and a dir `gitops-system` in the repo.

```
gotk bootstrap github --owner=<org-name> --repository=<private-repo-name> --team=platform --path=bernard

[INFO]  Downloading metadata https://api.github.com/repos/fluxcd/toolkit/releases/latest
[INFO]  Using 0.0.28 as release
[INFO]  Downloading hash https://github.com/fluxcd/toolkit/releases/download/v0.0.28/toolkit_0.0.28_checksums.txt
[INFO]  Downloading binary https://github.com/fluxcd/toolkit/releases/download/v0.0.28/gotk_0.0.28_darwin_amd64.tar.gz
[INFO]  Verifying binary download
[INFO]  Installing gotk to /usr/local/bin/gotk
► connecting to github.com
✔ repository cloned
✚ generating manifests
✔ components manifests pushed
► installing components in gitops-system namespace
namespace/gitops-system unchanged
customresourcedefinition.apiextensions.k8s.io/alerts.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/gitrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmcharts.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmreleases.helm.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/providers.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/receivers.notification.toolkit.fluxcd.io configured
role.rbac.authorization.k8s.io/crd-controller-gitops-system unchanged
rolebinding.rbac.authorization.k8s.io/crd-controller-gitops-system unchanged
clusterrolebinding.rbac.authorization.k8s.io/cluster-reconciler-gitops-system unchanged
service/notification-controller unchanged
service/source-controller unchanged
service/webhook-receiver unchanged
deployment.apps/helm-controller configured
deployment.apps/notification-controller configured
deployment.apps/source-controller configured
networkpolicy.networking.k8s.io/deny-ingress unchanged
deployment "source-controller" successfully rolled out
[...]

PR #2601, opened 2022-03-31
[RFC-0003] Flux OCI support for Kubernetes manifests

stefanprodan:
Proposal for adding OCI support for Kubernetes manifests to Flux.

Flux should be able to distribute and reconcile Kubernetes configuration packaged as OCI artifacts.

On the client-side, the Flux CLI should offer a command for packaging Kubernetes configs into an OCI artifact and pushing the artifact to a container registry using the Docker config file and the Docker credential helpers for authentication.

```sh
flux push artifact oci://ghcr.io/org/my-app-config:v1.0.0 --path ./deploy
```

On the server-side, the Flux source-controller should offer a dedicated API Kind for defining how OCI artifacts are pulled from container registries and how the artifact's authenticity can be verified. Flux source-controller should be able to work with any type of artifact even if it's not created with the Flux CLI.

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: OCIRepository
metadata:
  name: app-config
spec:
  interval: 10m
  url: oci://ghcr.io/org/my-app-config
  ref:
    semver: "1.x"
  secretRef:
    name: my-app-regcred
  verify:
    provider: cosign
    secretRef:
      name: my-app-cosgin-key
```

To apply manifests stored in OCI artifacts, Flux kustomize-controller should accept `OCIRepository` as a source:

```yaml
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
metadata:
  name: app
spec:
  interval: 10m
  serviceAccountName: app-applier
  sourceRef:
    kind: OCIRepository
    name: app-config
  path: ./
  wait: true
  prune: true
```

This RFC is sponsored by @hiddeco

### DIFF
--- a/rfcs/kubernetes-oci/README.md
+++ b/rfcs/kubernetes-oci/README.md
@@ -291,7 +291,9 @@ spec:
 
 ### Alternatives
 
-TODO
+An alternative solution is to introduce an OCI artifact type especially made for Kubernetes configuration.
+That is considered unpractical, as introducing an OCI type has to go through the
+IANA process and Flux is not the owner of those type as Helm is for Helm artifact for example.
 
 ## Design Details
 
--- a/rfcs/kubernetes-oci/README.md
+++ b/rfcs/kubernetes-oci/README.md
@@ -110,7 +110,25 @@ spec:
 The `secretRef` points to a Kubernetes secret in the same namespace as the `OCIRepository`,
 the secret type must be `kubernetes.io/dockerconfigjson`.
 
-When Flux runs on EKS or GKE, an IAM role (that grants read-only access to ACR, ECR or GCR)
+For private repositories which require a certificate to authenticate,
+the client certificate, private key and the CA certificate (if self-signed), can be provided with:
+
+```yaml
+spec:
+  certSecretRef:
+    name: regcert
+```
+
+The `certSecretRef` points to a Kubernetes secret in the same namespace as the `OCIRepository`:
+
+```shell
+kubectl create secret generic regcert \
+  --from-file=certFile=client.crt \
+  --from-file=keyFile=client.key \
+  --from-file=caFile=ca.crt
+```
+
+When Flux runs on AKS, EKS or GKE, an IAM role (that grants read-only access to ACR, ECR or GCR)
 can be used to bind the `source-controller` to the IAM role.
 
 Similar to image-reflector-controller
@@ -125,6 +143,19 @@ source-controller will expose dedicated flags for each cloud provider:
 
 We should extract the flags and the AWS, Azure and GCP auth implementations from image-reflector-controller into 
 `fluxcd/pkg/oci/auth` to reuses the code in source-controller.
+
+### Pull artifacts from self-hosted repositories
+
+For self-hosted Docker registries where the API is exposed with a self-signed TLS certificate,
+the CA certificate and private key can be provided with: 
+
+```yaml
+spec:
+  secretRef:
+    name: regcred
+
+```
+
 
 ### Reconcile artifacts
 
--- a/rfcs/kubernetes-oci/README.md
+++ b/rfcs/kubernetes-oci/README.md
@@ -83,7 +83,23 @@ spec:
     semver: "6.0.x"
 ```
 
-For private repositories, the credentials can be supplied with:
+To verify the authenticity of an artifact, the Sigstore cosign public key can be supplied with:
+
+```yaml
+spec:
+  verify:
+    provider: cosign
+    secretRef:
+      name: cosign-key
+```
+
+### Pull artifacts from private repositories
+
+For authentication purposes, Flux users can choose between supplying static credentials with Kubernetes secrets
+and cloud-based OIDC using an IAM role binding to the source-controller Kubernetes service account.
+
+For private repositories hosted on DockerHub, GitHub, Quay, self-hosted Docker Registry and others,
+the credentials can be supplied with:
 
 ```yaml
 spec:
@@ -94,15 +110,21 @@ spec:
 The `secretRef` points to a Kubernetes secret in the same namespace as the `OCIRepository`,
 the secret type must be `kubernetes.io/dockerconfigjson`.
 
-To verify the authenticity of an artifact, the Sigstore cosign public key can be supplied with:
+When Flux runs on EKS or GKE, an IAM role (that grants read-only access to ACR, ECR or GCR)
+can be used to bind the `source-controller` to the IAM role.
 
-```yaml
-spec:
-  verify:
-    provider: cosign
-    secretRef:
-      name: cosign-key
+Similar to image-reflector-controller
+[auto-login feature](https://fluxcd.io/docs/guides/image-update/#imagerepository-cloud-providers-authentication),
+source-controller will expose dedicated flags for each cloud provider:
+
+```sh
+--aws-autologin-for-ecr
+--azure-autologin-for-acr
+--gcp-autologin-for-gcr
 ```
+
+We should extract the flags and the AWS, Azure and GCP auth implementations from image-reflector-controller into 
+`fluxcd/pkg/oci/auth` to reuses the code in source-controller.
 
 ### Reconcile artifacts
 
--- a/rfcs/kubernetes-oci/README.md
+++ b/rfcs/kubernetes-oci/README.md
@@ -1,0 +1,140 @@
+# RFC-xxxx Flux OCI support for Kubernetes manifests
+
+**Status:** provisional
+
+**Creation date:** 2022-03-31
+
+**Last update:** 2022-03-31
+
+## Summary
+
+Flux should be able to distribute and reconcile Kubernetes configuration packaged as OCI artifacts.
+
+On the client-side, the Flux CLI should offer a command for packaging Kubernetes configs into
+an OCI artifact and pushing the artifact to a container registry using the Docker config file
+and the Docker credential helpers for authentication.
+
+On the server-side, the Flux source-controller should offer a dedicated API Kind for defining
+how OCI artifacts are pulled from container registries and how the artifact's authenticity can be verified.
+Flux should be able to work with any type of artifact even if it's not created with the Flux CLI.
+
+## Motivation
+
+Given that OCI registries are evolving into a generic artifact storage solution,
+we should extend Flux to allow fetching Kubernetes manifests and related configs
+from container registries similar to how Flux works with Git and Bucket storage.
+
+With OCI support, Flux users can automate artifact updates to Git in the same way
+they do today for container images.
+
+### Goals
+
+- Add support to the Flux CLI for packaging Kubernetes manifests and related configs into OCI artifacts.
+- Add support to Flux source-controller for fetching configs stored as OCI artifacts.
+- Make it easy for users to switch from Git repositories and Buckets to OCI repositories.
+
+### Non-Goals
+
+- Introduce a new OCI media type for artifacts containing Kubernetes manifests.
+
+## Proposal
+
+### Push artifacts
+
+Flux users should be able to package a local directory containing Kubernetes configs into a tarball
+and push the archive to a container registry as an OCI artifact.
+
+```sh
+flux push artifact docker.io/org/app-config:v1.0.0 -f ./deploy
+```
+
+Flux CLI with produce artifacts of type `application/vnd.oci.image.config.v1+json`.
+The directory pointed to by `-f` is archived and compressed in the `tar+gzip` format
+and the layer media type is set to `application/vnd.oci.image.layer.v1.tar+gzip`.
+
+> A proof-of-concept CLI implementation for distributing Kubernetes configs as OCI artifacts
+> is available at [kustomizer.dev](https://github.com/stefanprodan/kustomizer).
+
+### Pull artifacts
+
+Flux users should be able to define a source for pulling manifests inside the cluster from an OCI repository.
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: app-config
+  namespace: flux-system
+spec:
+  interval: 10m
+  url: docker.io/org/app-config
+  ref:
+    tag: v1.0.0
+```
+
+An `OCIRepository` can refer to an artifact by tag, digest or semver range:
+
+```yaml
+spec:
+  ref:
+    # one of
+    tag: "latest"
+    digest: "sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2"
+    semver: "6.0.x"
+```
+
+For private repositories, the credentials can be supplied with:
+
+```yaml
+spec:
+  secretRef:
+    name: regcred
+```
+
+The `secretRef` points to a Kubernetes secret in the same namespace as the `OCIRepository`,
+the secret type must be `kubernetes.io/dockerconfigjson`.
+
+To verify the authenticity of an artifact, the Sigstore cosign public key can be supplied with:
+
+```yaml
+spec:
+  verify:
+    provider: cosign
+    secretRef:
+      name: cosign-key
+```
+
+### Reconcile artifacts
+
+The `OCIRepository` can be used as a drop-in replacement for `GitRepository` and `Bucket` sources.
+For example a Flux Kustomization can refer to an `OCIRepository` and reconcile the manifests found in the OCI artifact:
+
+```yaml
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: app
+  namespace: flux-system
+spec:
+  interval: 10m
+  sourceRef:
+    kind: OCIRepository
+    name: app-config
+  path: ./
+```
+
+### User Stories
+
+TODO
+
+### Alternatives
+
+TODO
+
+## Design Details
+
+TODO
+
+### Enabling the feature
+
+The feature is enabled by default.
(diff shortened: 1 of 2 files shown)


rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+```yaml
+spec:
+  secretRef:
+    name: regcred
+```
+
+The `secretRef` points to a Kubernetes secret in the same namespace as the `OCIRepository`,
+the secret type must be `kubernetes.io/dockerconfigjson`.
+
```
rashedkvm, line 112: One small suggestion. In addition to the repository credentials, an optional secret ref with certificate data is necessary, especially for registries with a self-signed certificate or certificate issued by a non-public CA. This optional field can be similar to `CertSecretRef` defined in type `ImageRepositorySpec`

```
        // CertSecretRef can be given the name of a secret containing
	// either or both of
	//
	//  - a PEM-encoded client certificate (`certFile`) and private
	//  key (`keyFile`);
	//  - a PEM-encoded CA certificate (`caFile`)
	//
	//  and whichever are supplied, will be used for connecting to the
	//  registry. The client cert and key are useful if you are
	//  authenticating with a certificate; the CA cert is useful if
	//  you are using a self-signed server certificate.
	// +optional
	CertSecretRef *meta.LocalObjectReference `json:"certSecretRef,omitempty"`
```

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,251 @@
...
+For authentication purposes, Flux users can choose between supplying static credentials with Kubernetes secrets
+and cloud-based OIDC using an IAM role binding to the source-controller Kubernetes service account.
+
+For private repositories hosted on DockerHub, GitHub, Quay, self-hosted Docker Registry and others,
+the credentials can be supplied with:
+
+```yaml
+spec:
+  secretRef:
```
rashedkvm, line 106: I think a `ServiceAccount` resource can be referenced here, in addition to `secretRef` to access registry credentials. Can the spec look something like this?
```yaml
spec:
  auth:  
    # one of
    secretRef:
      name: regcred
    serviceAccountName: reg
  certSecretRef: regcert
```

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,251 @@
...
+Flux CLI with produce artifacts of type `application/vnd.oci.image.config.v1+json`.
+The directory pointed to by `-f` is archived and compressed in the `tar+gzip` format
+and the layer media type is set to `application/vnd.oci.image.layer.v1.tar+gzip`.
+
+> A proof-of-concept CLI implementation for distributing Kubernetes configs as OCI artifacts
+> is available at [kustomizer.dev](https://github.com/stefanprodan/kustomizer).
+
+### Pull artifacts
+
+Flux users should be able to define a source for pulling manifests inside the cluster from an OCI repository.
```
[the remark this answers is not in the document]
stefanprodan: If we reuse the ORAS implementation from RFC #2597, we will only look at the first layer.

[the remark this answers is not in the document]
stefanprodan: Maybe @souleb can say more about it, he did implemented this RFC for experimentation purposes with ORAS.

rfcs/kubernetes-oci/README.md
stefanprodan: Maybe in the next iteration of the source API `v1beta3`, right now all APIs have `secretRef` in `spec`. If we want to get this into Flux before GA, then we need to conform to the current API standard.

rashedkvm: Sure, we can iterate on this in `v1beta3` and keep it as a suggestion.

rfcs/kubernetes-oci/README.md
stefanprodan: I've added `certSecretRef` to the spec. Thanks!

rfcs/kubernetes-oci/README.md
rashedkvm: @stefanprodan does it make sense to add `spec.serviceAccountName` in this API proposal?

stefanprodan: We need to add it to ImageRepository first, could you contribute that to image reflector?

rashedkvm: Yes @stefanprodan I can add `spec.serviceAccountName` in `ImageRepository` v1beta1 API and implement it in `image-reflector-controller`.  I will put up a PR for the `image-reflector-controller` in the next few days. That said, if you accept the `spec.serviceAccountName` for `OCIRepository` contingent on the PR for the `ImageRepository`, that will be great.

stefanprodan: Yeap I think `serviceAccountName` is a great addition to both image and source APIs, for consistency sake, we'll need this into image v1beta1 API. Thanks for your help.

rashedkvm: @stefanprodan logged an enhancement to include `ServiceAccountName` for `ImageRepository` v1beta1 API in image reflector controller https://github.com/fluxcd/image-reflector-controller/issues/251. I will submit the PR for [#251](https://github.com/fluxcd/image-reflector-controller/issues/251) by next week.

stefanprodan: I've added `serviceAccountName` to the proposal, thanks @rashedkvm for the IRC PR.

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,289 @@
...
+the secret type must be `kubernetes.io/dockerconfigjson`:
+
+```shell
+kubectl create secret docker-registry regcred \
+  --docker-server=<your-registry-server> \
+  --docker-username=<your-name> \
+  --docker-password=<your-pword>
+```
+
+For image pull secrets attached to a service account, the account name can be specified with:
```
rashedkvm, line 122: Thank you @stefanprodan. If you are ok with the OCIRepository API proposed, I am ready start implementation.

rashedkvm (approved):

rashedkvm:
@stefanprodan @hiddeco @pjbgf quick update. I am completing a different work assignment. I expect to have a draft PR for this RFC in the next 2 to 3 weeks. Let me know if this timeline conflicts with the road to GA. Thank you!

hiddeco:
This is absolutely fine I think, there is still quite some work to do for us in other controller areas around GA.

stefanprodan:
@tis-rpage Please read the RFC, cosign verification is there.

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: app-config
+  namespace: flux-system
+spec:
+  interval: 10m
+  url: docker.io/org/app-config
```
scothis, line 98: What is the intent for the `url` field? I find it a bit confusing for the value to not be a url, or in many cases to not even url-like (e.g. `ubuntu` for library images, or `user/repo` for docker hub images).

I see that the `GitRepository` and `HelmRepository` resources both use `url`, but in those cases the value actually is a URL. In the helm case, it may use the http(s) or oci scheme, but the oci scheme is not used in the `OCIRepository` examples. There was a concern raised about a custom OCI artifact media types not being registered with IANA; the `oci` scheme is also [not registered](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml).

The `ImageRepository` resource uses `image` which is what I would tend to expect. 

The behavior isn't called out, but I would expect a user to be able to specify a tag or digest as part of the image and optionally override that value with the `ref` struct. Is that the intent?

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,403 @@
...
+and push the archive to a container registry as an OCI artifact.
+
+```sh
+flux push artifact docker.io/org/app-config:v1.0.0 \
+  --path="./deploy" \
+  --source="$(git config --get remote.origin.url)" \
+  --revision="$(git branch --show-current)/$(git rev-parse HEAD)"
+```
+
+The Flux CLI with produce artifacts of type `"application/vnd.docker.distribution.manifest.v2+json`
```
darkowlzz, line 54:
```suggestion
The Flux CLI will produce artifacts of type `"application/vnd.docker.distribution.manifest.v2+json`
```

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,403 @@
...
+### Alternatives
+
+An alternative solution is to introduce an OCI artifact type especially made for Kubernetes configuration.
+That is considered unpractical, as introducing an OCI type has to go through the
+IANA process and Flux is not the owner of those type as Helm is for Helm artifact for example.
+
+## Design Details
+
+Both the Flux CLI and source-controller will use the [go-containerregistry](https://github.com/google/go-containerregistry)
+library for OCI operations such as push, pull, tag, list tags, etc.
```
darkowlzz, line 322: Maybe it's intentionally written this way, but I noticed that some of these commands have not been described in the above sections like `pull`, ` list tags` and `build`. Would it be good to describe them to provide some examples of what to expect from them or in the workflows like in the user stories?

rfcs/kubernetes-oci/README.md
stefanprodan: Thanks @scothis for raising this, I think I've addressed your comments in the latest commit.

scottrigby (approved):
This looks great – I'm pretty excited. 

I left a code suggestion that you could either accept or not, and I think this RFC would be good either way (though I do think we should link to whatever constraint spec we use somewhere in our docs).

Also left a question about this RFC allowing for future support of specific additional media types.

I do not think either of these are blockers for the RFC 👍

stefanprodan:
> But does this proposal exclude supporting additional media types compatible with our implementation in the future, as needed? 

@scottrigby Flux the CLI produces artifacts using the Docker media type to support all container registries. But we have no such requirement in Flux source-controller, Flux can pull any OCI artifact no matter the type.

hiddeco (approved):
This looks good to me. Thanks for explicitly writing out the commands.

It does not have a RFC number assigned yet (nor [an official sponsor](https://github.com/fluxcd/flux2/tree/main/rfcs#rfc-process)). The number should be `0003`.

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+Flux should be able to work with any type of artifact even if it's not created with the Flux CLI.
+
+## Motivation
+
+Given that OCI registries are evolving into a generic artifact storage solution,
+we should extend Flux to allow fetching Kubernetes manifests and related configs
+from container registries similar to how Flux works with Git and Bucket storage.
+
+With OCI support, Flux users can automate artifact updates to Git in the same way
+they do today for container images.
```
pjbgf, line 28: > With OCI support, Flux users can automate artifact updates to Git in the same way
> they do today for container images.

Not sure I understand what workflow we are referring here. Is this referring to updating Git as in what IAC does. Or is this about using OCI to trigger source-controller updates, pretty much as SC does for GitRepository objects?

stefanprodan: It refers to IAC, where Flux patches the `spec.ref.tag` in the same way it does today for `images[].newTag`.

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+  verify:
+    provider: cosign
+    secretRef:
+      name: cosign-key
+```
+
+### Pull artifacts from private repositories
+
+For authentication purposes, Flux users can choose between supplying static credentials with Kubernetes secrets
+and cloud-based OIDC using an IAM role binding to the source-controller Kubernetes service account.
```
pjbgf, line 140: Cloud seems redudant here, as OIDC is not exclusive to it. Users could implement and use this in an airgapped/on-prem environment just as well.

```suggestion
For authentication purposes, Flux users can choose between supplying static credentials as Kubernetes secrets
or OIDC by using an IAM role binding to the source-controller Kubernetes service account.
```

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+source-controller will expose dedicated flags for each cloud provider:
+
+```sh
+--aws-autologin-for-ecr
+--azure-autologin-for-acr
+--gcp-autologin-for-gcr
+```
+
+We should extract the flags and the AWS, Azure and GCP auth implementations from image-reflector-controller into 
+`fluxcd/pkg/oci/auth` to reuses the code in source-controller.
```
pjbgf, line 206:
```suggestion
`fluxcd/pkg/oci/auth` to reuse the code in source-controller.
```

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+### Reconcile artifacts
+
+The `OCIRepository` can be used as a drop-in replacement for `GitRepository` and `Bucket` sources.
+For example a Flux Kustomization can refer to an `OCIRepository` and reconcile the manifests found in the OCI artifact:
```
pjbgf, line 211:
```suggestion
For example, a Flux Kustomization can refer to an `OCIRepository` and reconcile the manifests found in the OCI artifact:
```

pjbgf (commented):
Apart from some small nits, LGTM.

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+and push the archive to a container registry as an OCI artifact.
+
+```sh
+flux push artifact docker.io/org/app-config:v1.0.0 \
+  --source="$(git config --get remote.origin.url)" \
+  --revision="$(git branch --show-current)/$(git rev-parse HEAD)" \
+  --path="./deploy"
+```
+
+The Flux CLI will produce artifacts of type `application/vnd.docker.distribution.manifest.v2+json`
```
souleb, line 54: I think we should produce `application/vnd.oci.image.manifest.v1+json`, manifest types. This is supported by all major container registries, docker included, and is the only one currently registered: https://www.iana.org/assignments/media-types/application/vnd.oci.image.manifest.v1+json

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,428 @@
...
+[auto-login feature](https://fluxcd.io/docs/guides/image-update/#imagerepository-cloud-providers-authentication),
+source-controller will expose dedicated flags for each cloud provider:
+
+```sh
+--aws-autologin-for-ecr
+--azure-autologin-for-acr
+--gcp-autologin-for-gcr
+```
+
+We should extract the flags and the AWS, Azure and GCP auth implementations from image-reflector-controller into
```
souleb, line 205: maybe move this to `design details` section?

rfcs/kubernetes-oci/README.md
scothis: > `repository` or `image` would make it more domain specific. The reason I'm even considering `image` is because I've observed that some providers change the meaning of registry and repositories.

If the field is called `image`, I would expect a full image reference including a tag or digest to work. This would make the `ref` field an optional override of the tag/digest on the `image`. I appreciate that two ways to define the same concept may be confusing. While `image` is my preferred field name (with the semantics), I suggested `repository` as a compromise between something that is not a URL nor a full image reference.

stefanprodan: Using `repository` feels redundant, the kind is called `OCIRepository`. Maybe we should follow Helm and enforce `oci://` in the `url` field in the same way we do for Flux `HelmRepository.spec.url`.

stefanprodan: I have added the `oci://` prefix to the API `spec.url` and to all the `flux <verb> artifact` commands. Thanks everyone for the feedback.

darkowlzz (approved):
LGTM!

pjbgf (approved):
LGTM

rfcs/kubernetes-oci/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+## Proposal
+
+### Push artifacts
+
+Flux users should be able to package a local directory containing Kubernetes configs into a tarball
+and push the archive to a container registry as an OCI artifact.
+
+```sh
+flux push artifact oci://docker.io/org/app-config:v1.0.0 \
```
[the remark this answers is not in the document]
stefanprodan: This is not an option for us, Flux the CLI is following kubectl conventions and changing this means breaking all our users workflows.

davinkevin:
Sorry if I ask this question too late, but is there a possibility to interact with this OCI entry after download by the OCIRepository component? 

As an end user, I would like to use it as the `HelmRelease` component, where I can provide some `values.yaml`. Here, it seems to be really `static`. I have a set of manifests (called `base` in our case) and every env (dev, prod…) would like to consume one (using OCI entry) but with different config, secret, patches and other stuff provided by `kustomize`. 

The main idea would be to replace our `base` by an OCI entry but without extension possible, we are not able to use them.

kingdonb:
I see that `GitRepositorySpec` is where the `includes` feature is defined, so it appears to be correct that OCIRepository cannot currently be "included" in a GitRepository or vice versa.

Should we try to better generalize this part of the spec so those includes would be possible, or is there something blocking?

davinkevin:
For information, I've elaborated my question in [this comment](https://github.com/fluxcd/source-controller/issues/854#issuecomment-1200091330), with some feedbacks on the current OCI specification for manifests:

Issue #2619, opened 2022-04-06
Flux CLI support for parameters from env vars (at least sensitive ones such as password)


canidam:
### Discussed in https://github.com/fluxcd/flux2/discussions/2557

<div type='discussions-op-text'>

<sup>Originally posted by **canidam** March 21, 2022</sup>
Bootstrapping flux with git subcommand using credentials requires the user to provide password over CLI.

```
  # Run bootstrap for a Git repository and authenticate using a password
  flux bootstrap git --url=https://example.com/repository.git --password=<password>
```

This is a bad pattern. I find it weird flux doesn't support provide credentials over environment variables or from a file.

Couple of problems with such pattern:
1. Password is kept in history by default (unless user remembers to start with the command with a space)
2. Other users on the machine can read the password in clear-text with `ps -aux`

Why does the flux CLI doesn't support variables from the environment? Is that something worth opening a PR for?
 </div>

canidam:
I use the git provider for bitbucket server, that's why I need it. I'll open PR to align this with the other providers

PR #2631, opened 2022-04-08
bootstrap git: Allow the password to be specified with GIT_PASSWORD env var

canidam:
Solves issue #2619 

This PR align the generic `bootstrap git` command with `bootstrap github` and `bootstrap gitlab`, allowing the user to provide password using environment var `GIT_PASSWORD` or be prompted to provide it as an input.

Signed-off-by: Chen Anidam <canidam@gmail.com>

### DIFF
--- a/cmd/flux/bootstrap_git.go
+++ b/cmd/flux/bootstrap_git.go
@@ -53,6 +53,9 @@ command will perform an upgrade if needed.`,
   # Run bootstrap for a Git repository and authenticate using a password
   flux bootstrap git --url=https://example.com/repository.git --password=<password>
 
+  # Run bootstrap for a Git repository and authenticate using a password from environment variable
+  GIT_PASSWORD=<password> && flux bootstrap git --url=https://example.com/repository.git
+
   # Run bootstrap for a Git repository with a passwordless private key
   flux bootstrap git --url=ssh://git@example.com/repository.git --private-key-file=<path/to/private.key>
 
@@ -71,6 +74,10 @@ type gitFlags struct {
 	silent   bool
 }
 
+const (
+	gitPasswordEnvVar = "GIT_PASSWORD"
+)
+
 var gitArgs gitFlags
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/flux/bootstrap_git.go
```diff
@@ -85,6 +92,18 @@ func init() {
 }
 
 func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
+	if gitArgs.password == "" {
```
stefanprodan, line 95: This change breaks bootstrap for most users, the password is an option field, usually you would run bootstrap using SSH.

canidam: Because password is optional, I've two suggestions how to fix this:

1. Change the condition to `if gitArgs.username != "" && gitArgs.password == "" {` as password is required when a username is defined. With this option, ssh auth method is not supported (in case password is required for the private-key)
2. If the env var is defined && password is empty, use it. This apply for https and ssh but we won't have a prompt in case the user would like to provide the password as stdin.

IMO, option 2 seems better. Users should not provide the password thru the CLI. 

wdyt?

canidam: I've updated the commit to handle both cases. If the user specifies `--token-auth` and password is not set, they will be prompted for typing it.

stefanprodan (changes requested):
Please remove all .idea files, you can ignore them in your `$HOME/.gitignore`.

cmd/flux/bootstrap_git.go
```diff
@@ -50,9 +50,12 @@ command will perform an upgrade if needed.`,
 	Example: `  # Run bootstrap for a Git repository and authenticate with your SSH agent
   flux bootstrap git --url=ssh://git@example.com/repository.git
 
-  # Run bootstrap for a Git repository and authenticate using a password
+  # Run bootstrap for a Git repository and authenticate using a password CLI
```
stefanprodan, line 53: Please remove CLI from here.

PR #2639, opened 2022-04-13
[RFC] Dedicated API for commit status updates

somtochiama:
Proposal for adding a new API kind `CommitStatus` to the notification-controller.

### DIFF
--- a/rfcs/0008-commitstatus/README.md
+++ b/rfcs/0008-commitstatus/README.md
@@ -1,0 +1,63 @@
+
+# RFC-xxxx Create a new API kind for Commit Status
+
+**Status:** provisional
+
+**Creation date:** 2022-05-12
+
+**Last update:** 2022-05-12
+
+## Summary
+
+There should be a dedicated kind in the notification controller for sending commit status notifications to git providers.
+
+## Motivation
+
+Currently, The Alert type can reference both git providers and chat providers. However, the differences between the two providers and how notifications being sent to them should be handled has continued to diverge. 
+For example, there is a limit on the status message for git provider and sending the same notification twice overwrites the first one.
+
+By creating a sepeerate kind for git commit status, the specific nuances of the provider can be properly taken into account.
+
+### Goals
+
+- Add a new kind `CommitStatus` to Flux notification controller for handling commit status notifications.
+
+### Non-Goals
+
+- Add new `spec` fields to the `Alert` kind.
+
+## Proposal
+
+Introduce a new kind in the notification-controller called `CommitStatus`.
+The `CommitStatus` kind will similar to the `Alert` kind, referencing a provider and including event sources to accept events from. The major difference is that it will only reference git providers. Additionally, it can only have a `Kustomization` as its event source since it requires the `revision` field in the event metadata.
+
+This proposal will also introduce breaking changes as the `Alert` kind will no longer send notifications to git providers.
+
+### User Stories
+
+If a user want to receive status notification on github for a deployment, a `CommitStatus` kind can be used:
+
+```yaml
+apiVersion: notification.toolkit.fluxcd.io/v1beta1
+kind: CommitStatus
+metadata:
+  name: webapp
+  namespace: flux-system
+spec:
+  providerRef: 
+    name: github # The `spec.type` of the provider has to be github, gitlab, or bitbucket
+  eventSeverity: info
+  eventSources:
+    - kind: Kustomization
+      name: webapp
+```
+
+### Alternatives
+
+Alternatively, we could keep using the `Alert` kind for sending notifications to git providers. While this might not be much of a pain now, it would continue to grow as the implementations details of new features might differ for the chat and git providers.
+
+## Implementation History
+
+- [Implement GitHub commit status notifier](https://github.com/fluxcd/notification-controller/pull/27)
+- [Add Gitlab notifier](https://github.com/fluxcd/notification-controller/pull/43)
+- [Add bitbucket notifier](https://github.com/fluxcd/notification-controller/pull/73)
--- a/rfcs/0008-commitstatus/README.md
+++ b/rfcs/0008-commitstatus/README.md
@@ -16,7 +16,7 @@ There should be a dedicated kind in the notification controller for sending comm
 Currently, The Alert type can reference both git providers and chat providers. However, the differences between the two providers and how notifications being sent to them should be handled has continued to diverge. 
 For example, there is a limit on the length of the name/context of the status for git provider.
 
-When one commit triggers more than one deployment, the commit status notification from one overwrites the other as seen in this [issue](https://github.com/fluxcd/notification-controller/issues/195).
+When one commit triggers more than one deployment/reconciliation in different clusters, the commit status notification from one overwrites the other as seen in this [issue](https://github.com/fluxcd/notification-controller/issues/195).
 A new field specific to the git commit status will allow Flux users to specify a prefix that will be prepended to the context to differentiate the statuses.
 
 By creating a separate kind for git commit status, the specific nuances of the provider can be properly taken into account and it would allow more flexibility in adding fields specific to each kind.
--- a/rfcs/0008-commitstatus/README.md
+++ b/rfcs/0008-commitstatus/README.md
@@ -14,13 +14,17 @@ There should be a dedicated kind in the notification controller for sending comm
 ## Motivation
 
 Currently, The Alert type can reference both git providers and chat providers. However, the differences between the two providers and how notifications being sent to them should be handled has continued to diverge. 
-For example, there is a limit on the status message for git provider and sending the same notification twice overwrites the first one.
+For example, there is a limit on the length of the name/context of the status for git provider.
 
-By creating a sepeerate kind for git commit status, the specific nuances of the provider can be properly taken into account.
+When one commit triggers more than one deployment, the commit status notification from one overwrites the other as seen in this [issue](https://github.com/fluxcd/notification-controller/issues/195).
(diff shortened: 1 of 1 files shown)


phillebaba:
@SomtochiAma I would suggest looking at this discussion. I kind of dropped the ball on implementing this last summer...

https://github.com/fluxcd/flux2/discussions/1529#discussioncomment-880161

SomtochiAma:
@phillebaba I have! It was a very good starting point.
I plan on writing another proposal on adding templating (with you as co-author since you have already written a bulk of it in the discussion) and keeping this proposal on just adding the new `CommitStatus` kind. 
Is the proposal too sparse?

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Alternatives
+
+Alternatively, we could keep using the `Alert` kind for sending notifications to git providers. While this might not be much of a pain now, it would continue to grow as the implementations details of new features might differ for the chat and git providers.
+
+## Implementation History
+
+- [Implement GitHub commit status notifier](https://github.com/fluxcd/notification-controller/pull/27)
+- [Add Gitlab notifier](https://github.com/fluxcd/notification-controller/pull/43)
+- [Add bitbucket notifier](https://github.com/fluxcd/notification-controller/pull/73)
```
stefanprodan, line 63: Azure DevOps is missing from here.

stefanprodan (commented):
I think this RFC needs to address the multi-cluster issue that users are running into. @SomtochiAma have you've seen https://github.com/fluxcd/notification-controller/pull/340?

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,63 @@
...
+metadata:
+  name: webapp
+  namespace: flux-system
+spec:
+  providerRef: 
+    name: github # The `spec.type` of the provider has to be github, gitlab, or bitbucket
+  eventSeverity: info
+  eventSources:
+    - kind: Kustomization
+      name: webapp
```
aryan9600, line 52: We should probably have a `summary` field here similar to [`Alert`](https://fluxcd.io/docs/components/notification/alert/#specification). This can be helpful in adding some metadata about the event, which can be helpful, for example when there are multiple clusters. (ref: https://github.com/fluxcd/notification-controller/issues/194)

stefanprodan: The name `summary` doesn't make sense for commit status IMO, it should be `prefix` or something else.

stefanprodan:
@phillebaba @SomtochiAma does it make sense to add templating to CommitStatus if the actual message is limited to a couple of characters ? Maybe adding `prefix` is enough to solve the multi-cluster issue, it's not like the cluster name is something people can template, there is no such metadata in the payload.

rfcs/0008-commitstatus/README.md
SomtochiAma: We could use `context`. 
`context` is the name of the field the GitHub and GitLab API . They both describe it as
```
A string label to differentiate this status from the status of other systems.
```

hiddeco:
It would be good to take into account that [cdevents](https://github.com/cdevents/spec) is shaping.

stefanprodan:
@SomtochiAma we need to take into account things like https://github.com/fluxcd/notification-controller/pull/369 it's not a commit status but it's not an alert either...

phillebaba:
So I have looked this over now and for the most part this looks pretty straight forward. My opinion is that this should only cover CommitStatus as it seems to be a well defined term within most Git providers today. Everything else like Github Workflow dispatch should be covered in another RFC, as it would be to much work dealing with that too.

I would like to push for adding some sort of templating feature to the commit status. Just because there would be so many different needs in the future that we cant really consider. The reasoning for why to have it in the first place is that we want to differ a commit status coming from different environments, but we could consider for example a environment with multiple regions. Does just having a prefix solve this? I am not sure but maybe. One use case which my comment covered was the option to add templating parameters from a configmap which would enable the same commit status to be deployed into multiple clsuters but result in different prefixes. The question is mostly how much of an hassle that work is worth.

stefanprodan:
> My opinion is that this should only cover CommitStatus as it seems to be a well defined term within most Git providers today. 

Agreed, I've merged the GitHub dispatch integration for the Alert API.

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Status:** provisional
+
+**Creation date:** 2022-05-12
+
+**Last update:** 2022-05-12
+
+## Summary
+
+There should be a dedicated kind in the notification controller for sending commit status notifications to git providers.
```
hiddeco, line 12: I would expect the summary to be written as `[This RFC] <summary>`. So in this particular case, something like:

> This RFC proposes to create a dedicated `CommitStatus` kind in the <notification controller API group> to allow configuration of commit status notifications to a (Git) `Provider`.

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+Currently, The Alert type can reference both git providers and chat providers. However, the differences between the two providers and how notifications being sent to them should be handled has continued to diverge.
```
hiddeco, line 16: 1. `Currently, the ...`
2. git -> Git
3. I would provide examples of what Git providers and/or chat providers are to provide further context.
4. The last part of this line is difficult to read ("being sent to them should be handled has continued to diverge").

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+**Last update:** 2022-05-12
+
+## Summary
+
+There should be a dedicated kind in the notification controller for sending commit status notifications to git providers.
+
+## Motivation
+
+Currently, The Alert type can reference both git providers and chat providers. However, the differences between the two providers and how notifications being sent to them should be handled has continued to diverge. 
+For example, there is a limit on the length of the name/context of the status for git provider.
```
hiddeco, line 17: Reference which proofs this?

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+When one commit triggers more than one deployment/reconciliation in different clusters, the commit status notification from one overwrites the other as seen in this [issue](https://github.com/fluxcd/notification-controller/issues/195).
+A new field specific to the git commit status will allow Flux users to specify a prefix that will be prepended to the context to differentiate the statuses.
+
+By creating a separate kind for git commit status, the specific nuances of the provider can be properly taken into account and it would allow more flexibility in adding fields specific to each kind.
```
hiddeco, line 22: I would restructure the argumentation here to gradually get to the point of introducing a new kind, because the changes to the current kind or not justified.

Paraphrasing:

```
We have observed the issue to be <problem>, this would require the introduction of
<something> because of <reason>. Which is ultimately solved by <introducing new kind>,
because of <argumentation>.
```

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+  prefix: dev
+```
+
+### Alternatives
+
+Alternatively, we could keep using the `Alert` kind for sending notifications to git providers. While this might not be much of a pain now, it would continue to grow as the implementations details of new features might differ for the chat and git providers.
+
+## Implementation History
+
+- [Implement GitHub commit status notifier](https://github.com/fluxcd/notification-controller/pull/27)
```
hiddeco, line 67: I would think none of these entries belong under "implementation history", as they are not related to this RFC but rather to what moved us to creating this RFC. Making it e.g. part of the motivation, which is the section in which you provide historical context to build up your case.

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Add new `spec` fields to the `Alert` kind.
+
+## Proposal
+
+Introduce a new kind in the notification-controller called `CommitStatus`.
+The `CommitStatus` kind will similar to the `Alert` kind, referencing a provider and including event sources to accept events from. The major difference is that it will only reference git providers. Additionally, it can only have a `Kustomization` as its event source since it requires the `revision` field in the event metadata.
+It would include a new field `.spec.prefix` that will be used to differentiate results of multiple deployments on different clusters triggered by the same commit and prevent one from overwriting the other.
+
+This proposal will also introduce breaking changes as the `Alert` kind will no longer send notifications to git providers.
```
hiddeco, line 39: This is very thin and lacks technical details.

rfcs/0008-commitstatus/README.md
```diff
@@ -0,0 +1,71 @@
...
+  eventSeverity: info
+  eventSources:
+    - kind: Kustomization
+      name: webapp
+  prefix: dev
+```
+
+### Alternatives
+
+Alternatively, we could keep using the `Alert` kind for sending notifications to git providers. While this might not be much of a pain now, it would continue to grow as the implementations details of new features might differ for the chat and git providers.
```
hiddeco, line 63: Instead of

> While this might not be much of a pain now, ...

I would phrase it as (something like):

> This would not allow for further growth of the Git and chat providers. As the implementation details of new features might differ, making it difficult to invent an API around this which is self explanatory and user friendly.

hiddeco (commented):
First round of comments to steer towards something I would be comfortable with.

In general, I think the document needs more tidying in terms of style, line wraps, punctuation, etc. which I didn't nitpick on in full.

Other than this, I think what the RFC itself is suggesting is the right direction.

PR #2651, opened 2022-04-19
kustomize: use FS from `fluxcd/pkg`

hiddeco:
This switches to a secure FS implementation in most places, except for
where we can not make changes at this moment because it would break
behavior.

Not handled in this commit:

- Allowing the root for `manifestgen` packages to be configured.
- Allowing the user to define a working root while building locally.
- Defaulting to the secure FS implementation in
  `kustomization.MakeDefaultOptions`. Problem here is that constructing
  the secure FS could result in an error, which we can not surface
  without signature changes to the constructor func.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/fluxcd/kustomize-controller/api v0.24.4
 	github.com/fluxcd/notification-controller/api v0.23.4
 	github.com/fluxcd/pkg/apis/meta v0.12.2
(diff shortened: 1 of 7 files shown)


hiddeco:
We need to postpone this until @stefanprodan is back next week, as I currently can't steer around some of the assumptions (and lack of a root argument) in some of the generating parts of the API.

stefanprodan (approved):
LGTM

pjbgf (approved):
Great work on this @hiddeco! :clap: :clap: 

LGTM

Issue #2699, opened 2022-05-04
cannot upgrade to 0.30.0 on macOS


masterkain:
### Describe the bug

I get an error when trying to upgrade clusters to flux 0.30.0, macOS 12.3.1, flux installed via homebrew

### Steps to reproduce

```
✔ cloned repository
► generating component manifests
✗ component manifest generation failed: fs-security-constraint write /var/folders/ds/rk7m6qr520xfg1v0lprtzkrm0000gn/T/flux-manifests-4129662917/gotk-components.yaml: path '/var/folders/ds/rk7m6qr520xfg1v0lprtzkrm0000gn/T/flux-manifests-4129662917/gotk-components.yaml' is not in or below '/private/var/folders/ds/rk7m6qr520xfg1v0lprtzkrm0000gn/T/flux-manifests-4129662917'
```

### Expected behavior

upgrade as usual

### OS / Distro

macOS 12.3.1

### Flux version

flux: v0.30.0

### Flux check

```
► checking prerequisites
✔ Kubernetes 1.23.6-2+2a84a218e3cd52 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.20.1
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.22.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.17.2
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.23.4
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.24.4
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.24.3
✔ all checks passed
```

### Git provider

GitHub

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

robertpcontreras-ts:
Getting the same error

comjf:
I can confirm that a similar error appears when running `flux bootstrap github`. Same OS & cli version as OP.

hiddeco:
Should be solved by updating to `v0.30.1`.

mrgreyves:
Hi!
Same problem but on flux version 0.30.1
OS macOS 12.3.1
A little more information:
```
flux --version

flux version 0.30.1

```

```
flux check --pre

► checking prerequisites
[...]

PR #2700, opened 2022-05-04
MacOS: fix bootstrap manifest generation

stefanprodan:
Use the OS package to write the generated files on disk instead of Flux  secure FS package which is meant for read operations.

Fix: #2699

### DIFF
--- a/pkg/manifestgen/install/manifests.go
+++ b/pkg/manifestgen/install/manifests.go
@@ -26,7 +26,6 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


aryan9600:
I think a better fix would be to resolve symlinks for temp dirs, refer to: https://github.com/fluxcd/kustomize-controller/pull/630/files

stefanprodan:
@aryan9600 `os.WriteFile` is used for all the other file generators except for this one. Switching to SecureFS for all write operations could be done in a different PR.

hiddeco (approved):

PR #2717, opened 2022-05-09
Fix: Cluster-Dashboard variable missing datasource

mlashley:
Signed-off-by: Malcolm Lashley <mlashley@gmail.com>

### DIFF
--- a/manifests/monitoring/grafana/dashboards/cluster.json
+++ b/manifests/monitoring/grafana/dashboards/cluster.json
@@ -916,7 +916,7 @@
             "$__all"
           ]
         },
-        "datasource": null,
+        "datasource": "$DS_PROMETHEUS",
         "definition": "label_values(gotk_reconcile_condition, exported_namespace)",
         "description": null,
         "error": null,


kingdonb:
Thank you for contributing this!

Can you comment or provide a screenshot about what the effect of this missing data source is to the end user?

kingdonb:
(It looks like this one is superseded by #2727)

mlashley:
> Can you comment or provide a screenshot about what the effect of this missing data source is to the end user?

Sorry - I didn't take a screenshot - and my updated version is already deployed to my cluster - but the 'namespace' dashboard variable  wasn't populated for me without it... there is no 'custom all value' - and the default (at least in Grafana I was using) is to `x|y|z|...` all the values for 'all' - with the net result that all the promql using `gotk_blah{xxx="$namespace"}` failed to match any labels. I see that #2727 has included my change too.

kingdonb:
Thanks for the elaboration, that helps! 👍

I will give this a try later this week if someone else does not get to it first. This should get merged no problem. Thanks for your contribution here.

mlashley:
This has since been fixed in #2727 - we should close this one.

PR #2748, opened 2022-05-18
fix e2e tests

makkes:
fix e2e check test

The output of `kubectl version` has changed with newer kubectl version
from

```
{
  "serverVersion": ...,
  "clientVersion": ...
}
```

to

```
{
  "serverVersion": ...,
  "clientVersion": ...,
  "kustomizeVersion": ...
}
```

So the `kustomizeVersion` field is new which causes the JSON
unmarshaling to fail.

We now just unmarshal it to `map[string]interface{}` and peel the
server git version out of that map manually w/o unmarshalling the JSON
into a custom type.

### DIFF
--- a/cmd/flux/check_test.go
+++ b/cmd/flux/check_test.go
@@ -22,6 +22,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"strings"
 	"testing"
 
 	"github.com/fluxcd/flux2/internal/utils"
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
Thank you @makkes !

LGTM

makkes:
@pjbgf I added a commit that actually fixes the e2e tests.

PR #2727, opened 2022-05-12
grafana: display exported ns, slight resizing, default sorting by state

cr1cr1:
Signed-off-by: Cristian Chiru <cristi.chiru@gmail.com>

Includes #2717 and #2749 and fixes the exported namespace datasource.

image

PR #2749, opened 2022-05-18
grafana: fix the exported namespace filter datasource

mlazowik:
Signed-off-by: Michał Łazowik <m1@qed.ai>


stefanprodan:
This was fixed in #2727

Issue #1522, opened 2021-06-15
Breaking changes in Flux due to Kustomize v4


stefanprodan:
Starting with version 0.15.0, Flux and its controllers have been upgraded to Kustomize v4. While Kustomize v4 comes with many improvements and bug fixes, it introduces a couple of breaking changes.

### Remote archives 

Due to the removal of `hashicorp/go-getter` from Kustomize v4, the set of URLs accepted by Kustomize in the `resources` filed is reduced to file system paths, URLs to plain YAMLs and values compatible with `git clone`.

This means you can no longer use resources from archives (zip, tgz, etc).

No longer works:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://github.com/rook/rook/archive/refs/heads/master.zip//rook-master/cluster/examples/kubernetes/ceph/crds.yaml
```

Works:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://raw.githubusercontent.com/rook/rook/v1.6.0/cluster/examples/kubernetes/ceph/crds.yaml
```

### Non-string YAML keys

Due to a [bug](https://github.com/kubernetes-sigs/kustomize/issues/3446) in Kustomize v4, if you have **non-string keys** in your manifests, the controller will fail to build the final manifest.

The non-string keys bug affects Helm release like the nginx-ingress one, for example:

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: nginx-ingress
spec:
  values:
    tcp:
      2222: "app/server:2222"
```

The above will fail with `{}{2222:"app/server:2222:2222"}}}}: json: unsupported type: map[interface {}]interface {}`.

To fix this issue, you have to make the YAML keys into strings, e.g.:

```yaml
  values:
    tcp:
      "2222": "app/server:2222"
```

### Duplicate YAML keys

Unlike Helm, the Kustomize yaml parser (kyaml) does not accept duplicate keys, while Helm drops the duplicates, Kustomize errors out. This impacts helm-controller as it uses `kustomize/kyaml` to label objects reconciled by a `HelmRelease`.

For example, a chart that adds the `app.kubernetes.io/name` more than once, will result in a HelmRelease install failure:

```
map[string]interface {}(nil): yaml: unmarshal errors:
line 21: mapping key "app.kubernetes.io/name" already defined at line 20
```

### YAML formatting

[...]

PR #2764, opened 2022-05-24
Ensure proper FS root is set while bootstrapping

hiddeco:
Addresses https://github.com/fluxcd/flux2/issues/1522#issuecomment-1135102839


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

Issue #2763, opened 2022-05-24
Add `--ignore-path` to `flux create source`


stefanprodan:
We need to add a new arg (string array) to `flux create source git` and `flux create source bucket` that populates the `ignore` spec filed, each arg would add a new line to the ignore value e.g.:

```sh
flux create source git podinfo \
--url=https://github.com/stefanprodan/podinfo \
--branch=master \
--ignore-path=".git/" \
--ignore-path=".github/" \
--export
```

The above command will generate:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: GitRepository
metadata:
  name: podinfo
  namespace: flux-system
spec:
  interval: 5m
  url: https://github.com/stefanprodan/podinfo
  ref:
    branch: master
  ignore: |
    .git/
    .github/
```

takirala:
Addressed in https://github.com/fluxcd/flux2/pull/2767

PR #2767, opened 2022-05-24
Add `--ignore-paths` flag to `flux create source (git|bucket)`

takirala:
A new --ignore-path flag is added to following commands:

```
flux create source git --ignore-paths ...
flux create source bucket --ignore-paths ...
```

A StringSliceVar is used which supports specifying the flag multiple
times to populate a list or either a comma seperated string value

Fixes https://github.com/fluxcd/flux2/issues/2763

### DIFF
--- a/cmd/flux/create_source_bucket.go
+++ b/cmd/flux/create_source_bucket.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
@@ -63,17 +64,18 @@ For Buckets with static authentication, the credentials are stored in a Kubernet
 }
 
 type sourceBucketFlags struct {
-	name      string
-	provider  flags.SourceBucketProvider
-	endpoint  string
-	accessKey string
-	secretKey string
-	region    string
-	insecure  bool
-	secretRef string
+	name        string
+	provider    flags.SourceBucketProvider
+	endpoint    string
+	accessKey   string
+	secretKey   string
+	region      string
+	insecure    bool
+	secretRef   string
+	ignorePaths []string
 }
 
-var sourceBucketArgs = NewSourceBucketFlags()
+var sourceBucketArgs = newSourceBucketFlags()
 
 func init() {
 	createSourceBucketCmd.Flags().Var(&sourceBucketArgs.provider, "provider", sourceBucketArgs.provider.Description())
@@ -84,11 +86,12 @@ func init() {
 	createSourceBucketCmd.Flags().StringVar(&sourceBucketArgs.region, "region", "", "the bucket region")
 	createSourceBucketCmd.Flags().BoolVar(&sourceBucketArgs.insecure, "insecure", false, "for when connecting to a non-TLS S3 HTTP endpoint")
 	createSourceBucketCmd.Flags().StringVar(&sourceBucketArgs.secretRef, "secret-ref", "", "the name of an existing secret containing credentials")
+	createSourceBucketCmd.Flags().StringSliceVar(&sourceBucketArgs.ignorePaths, "ignore-paths", nil, "set paths to ignore in bucket resource (can specify multiple paths with commas: path1,path2)")
 
 	createSourceCmd.AddCommand(createSourceBucketCmd)
 }
 
-func NewSourceBucketFlags() sourceBucketFlags {
+func newSourceBucketFlags() sourceBucketFlags {
 	return sourceBucketFlags{
 		provider: flags.SourceBucketProvider(sourcev1.GenericBucketProvider),
 	}
@@ -116,6 +119,12 @@ func createSourceBucketCmdRun(cmd *cobra.Command, args []string) error {
 	}
 	defer os.RemoveAll(tmpDir)
 
+	var ignorePaths *string
+	if len(sourceBucketArgs.ignorePaths) > 0 {
+		ignorePathsStr := strings.Join(sourceBucketArgs.ignorePaths, "\n")
+		ignorePaths = &ignorePathsStr
+	}
+
 	bucket := &sourcev1.Bucket{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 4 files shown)


cmd/flux/create_source_git.go
```diff
@@ -189,6 +192,13 @@ func createSourceGitCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var ignorePaths *string
+	if len(sourceGitArgs.ignorePaths) > 0 {
+		ignorePathsStr := strings.Join(sourceGitArgs.ignorePaths, "\n")
+		//ignorePathsStr = fmt.Sprintf("%s\n", ignorePathsStr)
+		ignorePaths = &ignorePathsStr
```
takirala, line 199: Right now, if a user does 

```
./bin/flux create source git podinfo \
--ignore-path=".gitasd/" \
--ignore-path=".github/" \
--export
```

it will result in a manifest that looks like : 

```
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: GitRepository
metadata:
  name: podinfo
  namespace: flux-system
spec:
  ignore: |-
    .gitasd/
    .github/
  interval: 1m0s
  ref:
    branch: master
  url: https://github.com/stefanprodan/podinfo
```

notice the `|-` after the `ignore:` field which means the new line at the end of the multi line string is stripped. if we don't want that, we can add a fake newline at the end of the `ignorePathsStr` after concatenating the list (uncommenting the above line) and then it will result in : 

```
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: GitRepository
metadata:
  name: podinfo
  namespace: flux-system
spec:
  ignore: |
    .gitasd/
    .github/
  interval: 1m0s
  ref:
    branch: master
  url: https://github.com/stefanprodan/podinfo
```
which means no newline chars are stripped. I think this does not make a lot of difference in source-controller and we can leave this as-is (i ll remove the commented out code of course), but i am not very familiar with inner working of source controller. The source controller seems to be using the `ignore` field here - https://github.com/fluxcd/source-controller/blob/55a594a4e13f50f241327015e7e336ad5268b28b/controllers/gitrepository_controller.go#L658 and i think the `ReadPattern` function should work normally. 

Also refer https://yaml-multiline.info/ for an intuitive playground of how `|-` vs `|+` vs `|` work in yaml (i found it handy :))

stefanprodan: I'm Ok with how is now, removing the last line should cause no issues for source controller.

cmd/flux/create_source_bucket.go
```diff
@@ -84,11 +86,12 @@ func init() {
 	createSourceBucketCmd.Flags().StringVar(&sourceBucketArgs.region, "region", "", "the bucket region")
 	createSourceBucketCmd.Flags().BoolVar(&sourceBucketArgs.insecure, "insecure", false, "for when connecting to a non-TLS S3 HTTP endpoint")
 	createSourceBucketCmd.Flags().StringVar(&sourceBucketArgs.secretRef, "secret-ref", "", "the name of an existing secret containing credentials")
+	createSourceBucketCmd.Flags().StringSliceVar(&sourceBucketArgs.ignorePaths, "ignore-path", nil, "set paths to ignore in bucket resource (can specify multiple times)")
```
makkes, line 89:
```suggestion
	createSourceBucketCmd.Flags().StringSliceVar(&sourceBucketArgs.ignorePaths, "ignore-path", nil, "set paths to ignore in bucket resource (can specify multiple paths with commas: path1,path2)")
```
this is in line with the documentation of the `--label` flag.

cmd/flux/create_source_git.go
```diff
@@ -139,6 +141,7 @@ func init() {
 	createSourceGitCmd.Flags().BoolVar(&sourceGitArgs.recurseSubmodules, "recurse-submodules", false,
 		"when enabled, configures the GitRepository source to initialize and include Git submodules in the artifact it produces")
 	createSourceGitCmd.Flags().BoolVarP(&sourceGitArgs.silent, "silent", "s", false, "assumes the deploy key is already setup, skips confirmation")
+	createSourceGitCmd.Flags().StringSliceVar(&sourceGitArgs.ignorePaths, "ignore-path", nil, "set paths to ignore in git resource (can specify multiple times)")
```
makkes, line 144:
```suggestion
	createSourceGitCmd.Flags().StringSliceVar(&sourceGitArgs.ignorePaths, "ignore-path", nil, "set paths to ignore in git resource (can specify multiple paths with commas: path1,path2)")
```

makkes (commented):
Given you can specify multiple paths I think we should call the flag `--ignore-paths` as we do with e.g. `--labels`.

cmd/flux/create_source_git_test.go
```diff
@@ -87,7 +87,7 @@ func (r *reconciler) conditionFunc() (bool, error) {
 
 func TestCreateSourceGit(t *testing.T) {
 	// Default command used for multiple tests
-	var command = "create source git podinfo --url=https://github.com/stefanprodan/podinfo --branch=master --timeout=" + testTimeout.String()
+	var command = "create source git podinfo --url=https://github.com/stefanprodan/podinfo --branch=master --ignore-path .cosign,non-existent-dir/ --timeout=" + testTimeout.String()
```
makkes, line 90: Please add an assertion of the `.spec.ignore` field here.

takirala: i tried to add a `--export` flag to command and add a new case here but since these unit test cases deal with adding a status to the object and then verifying it, they always deal with a valid (and dynamic) namespace. i couldn't assert with a dynamic namespace, so i created a new test case where a gitrepository resource is simply exported (printed to stdout) and then asserted against a file.

stefanprodan:
@takirala the test is failing, you need to set the namespace and the interval in the command. After the test works for you locally please squash all commits into one and force push.

makkes (approved):

makkes:
lgtm module what Stefan said.

takirala:
I have squashed the commits after making necessary changes, tests are :heavy_check_mark: now!

stefanprodan (approved):
LGTM

Thanks @takirala 🏅

Issue #2696, opened 2022-05-03
Move to Go 1.18


hiddeco:
Given Go 1.18 has been out for some time now, and seems to be stable expect for some minor issues with the performance of the [newly introduced Generics](https://go.dev/doc/tutorial/generics), we should start moving to it.

This would also allow us to move to the most recent version of the `github.com/Azure/azure-sdk-for-go/sdk/*` modules, which require Go >=1.18.

### Prerequisites

- [x] `fluxcd/golang-with-libgit2`: needs to pass tests using Go 1.18.
- [x] `fluxcd/pkg`: low level packages all used by controllers.

### Controllers

- [x] `fluxcd/notification-controller` (depends on prerequisites)
- [x] `fluxcd/image-reflector-controller` (depends on prerequisites)
- [x] `fluxcd/source-controller` (depends on prerequisites)
- [x] `fluxcd/kustomize-controller` (depends on `fluxcd/source-controller`)
- [x] `fluxcd/helm-controller` (depends on `fluxcd/source-controller`)
- [x] `fluxcd/image-automation-controller` (depends on `fluxcd/source-controller` and `fluxcd/image-reflector-controller`)

### CLI

- [x] `fluxcd/flux2` (this repository)

### Other `fluxcd/flux2` umbrella projects

- [x] `fluxcd/cues`
- [x] `fluxcd/terraform-provider-flux`
- [x] `fluxcd/source-watcher`
- [x] `fluxcd/go-git-providers`

## :factory_worker:  Steps per project

- Update CI configuration in `.github/` to make use of Go 1.18.x
- Update `Dockerfile` resources to make use of Go 1.18.x
- **Optionally:** if we depend on `github.com/Azure/azure-sdk-for-go/*`, the `go 1.17` in `go.mod` can be updated to `go 1.18`. As there is a hard constraint on this version. For other projects, I would keep this, as it gate-keeps introduction of generics for now.
- **Optionally:** if `go.mod` was changed to include `1.18`, ensure any `-compat=1.17` rules in e.g. the `Makefile` of the project are changed to `-compat=1.18`. While preferably also moving it to a configurable flag option.
- **Optionally:** bump any `github.com/Azure/azure-sdk-for-go/*` dependencies to their latest versions.

phillebaba:
[...]

PR #2769, opened 2022-05-25
Update Go to 1.18 in CI

stefanprodan:
Changes:
- Update Go to 1.18 in CI (ref #2696)
- Update Alpine to v2.16
- Update kubectl to v1.24.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/fluxcd/flux2
 
-go 1.17
+go 1.18
```
hiddeco, line 3: I would not force this to 1.18, but instead keep compatibility at 1.17 for some time. This also guards against people trying to introduce generics in contributions.

hiddeco (approved):

PR #2781, opened 2022-05-28
Add OCI support to `create source helm`

makkes:


cmd/flux/create_source_helm.go
```diff
@@ -60,7 +60,18 @@ For private Helm repositories, the basic authentication credentials are stored i
     --url=https://stefanprodan.github.io/podinfo \
     --cert-file=./cert.crt \
     --key-file=./key.crt \
-    --ca-file=./ca.crt`,
+    --ca-file=./ca.crt
+
+  # Create a source for an OCI-hosted chart
```
souleb, line 65:
```suggestion
  # Create a source for an OCI Helm repository
```

The others one need to be adapted to HTTP(s) Helm repository

cmd/flux/create_source_helm.go
```diff
@@ -60,7 +60,18 @@ For private Helm repositories, the basic authentication credentials are stored i
...
+  flux create source helm podinfo \
+    --url=oci://ghcr.io/stefanprodan/charts/podinfo
+    --username=username \
+    --password=password
+
+  # Create a source for an OCI-hosted chart using an existing secret
```
souleb, line 71:
```suggestion
  # Create a source for an OCI Helm repository using an existing secret
```

souleb:
we use `stringData` in the underlying secret... We should use `data` format.

makkes:
rebased and addressed comments

stefanprodan (approved):
LGTM

Thanks @makkes 🏅

Issue #2786, opened 2022-05-30
Allow flux bootstrap with http git URLs


vipulnewaskar7:
**Context**
A git URL is necessary for flux bootstrap. Currently allowed schemes for bootstrapping are https and ssh.
If http URL is used, it fails with - 
"scheme _http_ is not supported"
For some reason, if user is not able to change http git URL to https, 
user just needs to set something like `--allow-insecure-http=true` 
and user will be able to bootstrap flux with _http_ URLs.

PR #2787, opened 2022-05-30
Add `--allow-insecure-http` to `bootstrap git`

vipulnewaskar7:
This change will allow the end-user to bootstrap with HTTP git URLs
But the user must explicitly set --allow-insecure-http=true
Fix: https://github.com/fluxcd/flux2/issues/2786


vipulnewaskar7:
Hi @stefanprodan 
I have updated the misleading error message and rebased master on my branch.

stefanprodan (approved):
LGTM

Thanks @vipulnewaskar7 🏅

Issue #2780, opened 2022-05-27
Handle multi-doc `kustomize.toolkit.fluxcd.io/v1beta2` YAML within flux build


anthr76:
### Describe the bug

When using on-disk local flux Kustomization file that is a multi-doc yaml patches within it aren't "built" on `flux build`

### Steps to reproduce

1. Consider a `flux-kustomize.yaml` like:

```yaml
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
metadata:
  name: infra-namespace
  namespace: flux-system
  labels:
    component.kutara.io/part-of: definitions
spec:
  path: "./k8s/base/infra"
  prune: true
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
metadata:
  name: kps
  namespace: flux-system
  labels:
    component.kutara.io/part-of: definitions
spec:
  path: "./k8s/base/infra/kps"
  prune: true
  patches:
    - patch: |
        - op: add
          path: /spec/values/kubeEtcd/endpoints
          value:
            - 192.168.8.40
            - 192.168.8.47
            - 192.168.8.51
      target:
        kind: HelmRelease
        name: kube-prometheus-stack
        namespace: monitoring
```
2. `flux build kustomization kps --path ./k8s/base/infra/kps --kustomization-file ./k8s/clusters/cluster-0/definitions/infra.yaml`

The kustomization will build successfully outputting the multi-doc yaml of `./k8s/base/infra/kps` but **not** include the patches in the `kps` `kustomize.toolkit.fluxcd.io/v1beta2` flux kustomize file.

If you were to take the above example and make it:

`flux-test.yaml`

```yaml
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
metadata:
  name: kps
  namespace: flux-system
  labels:
    component.kutara.io/part-of: definitions
spec:
  path: "./k8s/base/infra/kps"
  prune: true
  patches:
    - patch: |
        - op: add
          path: /spec/values/kubeEtcd/endpoints
          value:
            - 192.168.8.40
            - 192.168.8.47
            - 192.168.8.51
      target:
        kind: HelmRelease
        name: kube-prometheus-stack
        namespace: monitoring
```

and do `flux build kustomization kps --path ./k8s/base/infra/kps --kustomization-file ./flux-test.yaml` it would build correctly with it's patches.

### Expected behavior

[...]

PR #2792, opened 2022-06-01
Handle multi-doc yaml for flux build

SomtochiAma:
Fixes: #2780 

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


stefanprodan:
@SomtochiAma there is a typo in the commit message, should be `check` instead of `cheeck`.

internal/build/build.go
```diff
@@ -250,12 +251,19 @@ func (b *Builder) unMarshallKustomization() (*kustomizev1.Kustomization, error)
...
+	for !(k.Name == b.name && (k.Namespace == b.namespace || k.Namespace == "")) {
+		err = decoder.Decode(k)
+		if err != nil {
+			if err == io.EOF {
+				return nil, fmt.Errorf("failed find kustomization with name '%s' and namespace '%s' in file '%s'",
+					b.name, b.namespace, b.kustomizationFile)
+			} else {
+				return nil, fmt.Errorf("failed to unmarshall kustomization file %s: %w", b.kustomizationFile, err)
+			}
+		}
```
stefanprodan, line 266: We should return here to stop the iteration.

SomtochiAma: It decodes all valid Kubernetes YAML successfully. 
if we find the correct the correct YAML, for loop wouldn't run and the YAML is returned on the next line after the loop

stefanprodan: Got it! thanks

internal/build/testdata/local-kustomization/multi-doc-valid.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: Kustomization
+metadata:
+  name: infra-namespace
+  namespace: flux-system
+  labels:
+    component.kutara.io/part-of: definitions
+spec:
+  path: "./k8s/base/infra"
+  prune: true
+---
```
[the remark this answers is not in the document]
SomtochiAma: Like a second one?

stefanprodan (approved):
LGTM 

Thanks @SomtochiAma please squash and rebase.

PR #2795, opened 2022-06-01
use `data` field in Secrets

makkes:
there are some bugs in stringData, and it's better to just use data


stefanprodan:
While Kubernetes has issues with `stringData` we fixed them in Flux. See https://github.com/fluxcd/kustomize-controller/pull/576

Please raise issues for things like this before PRs. Thanks!

PR #2811, opened 2022-06-07
[RFC-0002] Update status and implementation history

stefanprodan:
Changes:
- Set status to partially implemented
- Add implementation history and TODOs
- Make corrections to the implementation details to match the reality

### DIFF
--- a/rfcs/0002-helm-oci/README.md
+++ b/rfcs/0002-helm-oci/README.md
@@ -1,10 +1,10 @@
 # RFC-0002 Flux OCI support for Helm
 
-**Status:** implementable
+**Status:** implemented (partially)
 
 **Creation date:** 2022-03-30
 
-**Last update:** 2022-04-13
+**Last update:** 2022-06-07
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


rfcs/0002-helm-oci/README.md
```diff
@@ -156,19 +156,25 @@ Bucket API design, where the same Kind servers different implementations: AWS S3
...
 
-The current `HelmRepositoryReconciler` will be renamed to `HelmRepositoryDefaultReconciler`,
-it's scope remains unchanged, and it will handle only objects with `type: Default`.
+The current `HelmRepositoryReconciler` will handle only objects with `type: default`,
+it's scope remains unchanged.
 
 We'll introduce a new reconciler named `HelmRepositoryOCIReconciler`, that will handle
-objects with `type: OCI`. This reconciler will set the `HelmRepository` Ready status to
+objects with `type: oci`. This reconciler will set the `HelmRepository` Ready status to
 `False` if the URL is not prefixed with `oci://`, otherwise the Ready status will be set to `True`.
```
hiddeco, line 164: I think this should also mention that it verifies the credentials, which has an impact on status state.

PR #2829, opened 2022-06-09
Accept multiple values for `flux create hr --values-from`

SomtochiAma:
This pull request updates `--values-from` flag for `flux create helmrelease` to accept a comma-separated list of object references.
e.g
```
flux create hr podinfo  --values-from=Secret/my-secret-values,ConfigMap/my-values
```


stefanprodan:
@SomtochiAma pflag does know about multi value flags, it knows how to parse the `,` and also accepts multiple flags. We want `--values-from` to behave like `--value`, to do this `valuesFrom` must be an array set with `StringSliceVar`. In your implementation, setting multiple `-values-from` has no effect, the last one overrides all others.

PR #2851, opened 2022-06-16
logs: write into writer from io.Pipe instead of os.Stdout

TianZong48:
This PR replaces `os.Stdout` with `writer` from `io.Pipe()`, and simplify strings compare with `strings.EqualFold()`

Thanks in advance for your review!

Signed-off-by: TianZong48 [tianzong48@gmail.com](mailto:tianzong48@gmail.com)

### DIFF
--- a/cmd/flux/logs.go
+++ b/cmd/flux/logs.go
@@ -204,12 +204,10 @@ func parallelPodLogs(ctx context.Context, requests []rest.ResponseWrapper) error
 	wg := &sync.WaitGroup{}
 	wg.Add(len(requests))
 
-	var mutex = &sync.Mutex{}
-
 	for _, request := range requests {
 		go func(req rest.ResponseWrapper) {
 			defer wg.Done()
-			if err := logRequest(mutex, ctx, req, os.Stdout); err != nil {
+			if err := logRequest(ctx, req, writer); err != nil {
 				writer.CloseWithError(err)
 				return
 			}
@@ -226,9 +224,8 @@ func parallelPodLogs(ctx context.Context, requests []rest.ResponseWrapper) error
 }
 
 func podLogs(ctx context.Context, requests []rest.ResponseWrapper) error {
-	mutex := &sync.Mutex{}
 	for _, req := range requests {
-		if err := logRequest(mutex, ctx, req, os.Stdout); err != nil {
+		if err := logRequest(ctx, req, os.Stdout); err != nil {
 			return err
 		}
 	}
@@ -246,7 +243,7 @@ func createLabelStringFromMap(m map[string]string) string {
 	return strings.Join(strArr, ",")
 }
 
-func logRequest(mu *sync.Mutex, ctx context.Context, request rest.ResponseWrapper, w io.Writer) error {
+func logRequest(ctx context.Context, request rest.ResponseWrapper, w io.Writer) error {
 	stream, err := request.Stream(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/flux/logs.go
```diff
@@ -266,23 +263,21 @@ func logRequest(mu *sync.Mutex, ctx context.Context, request rest.ResponseWrappe
 			break
 		}
 
-		mu.Lock()
```
TianZong48, line 269: func [io.Pipe](https://cs.opensource.google/go/go/+/go1.18.3:src/io/pipe.go;l=199)
> It is safe to call Read and Write in parallel with each other or with Close. Parallel calls to Read and parallel calls to Write are also safe: the individual calls will be gated sequentially.

SomtochiAma: Hey @TianZong48 , Thanks for your PR⭐️!
Stress testing your PR, without the mutex, The lines of the logs get merged occasionally
I got these by having multiple Kustomization and HelmReleases reconcile from the same source, then do flux reconcile git a couple of times
```
2022-06-18T09:47:16.168Z info HelmRelease2022-06-18T09:45:55.444Z info HelmRelease/podinfo-2./podinfo.default - reconcilation finished in 14.03725ms, next run in 1m0s 
2022-06-18T09:47:16.168Zdefault -  Discarding event, no alerts found for the involved objecterror HelmRelease/podinfo.default - Reconciler error install retries exhausted
 info HelmRelease/podinfo-3.default - Discarding event, no alerts found for the involved object 
```
Putting back the mutex fixes it. I am unsure why this happens the comment says it is safe.

TianZong48: Thanks for the testing🍻! I will try to figure it out

TianZong48: @SomtochiAma It turns out `template.Execute()` will call `w.Wirte()` several times each time. 

I reproduced the problem, and I think It's fixed :D

SomtochiAma: I tested your new changes and it seems to fix it 🎉. 
However, I still think it is possible for two lines to write to the buffer before `Flush` is called.🤔

TianZong48: I believe it won't happen except if one line of log is larger than `4096` bytes. 
Because: only 1 `bufio.Writer` per goroutine. And write into `bufio.Writer` within goroutine is called Sequentially.
However, if a log is larger than 4096(default buff size), It will write unexpectedly. Is that possible?  @SomtochiAma

TianZong48: Replaced `bufio.Writer` with `bytes.Buffer` which:
> A Buffer is a variable-sized buffer of bytes with Read and Write methods.

won't have the problem of `bufio.Writer` which needs a magic const( 4096 as default value) that may or may not work

TianZong48: I squashed these commits btw

SomtochiAma: That's a good point. I think we should keep your earlier implementation. `4096` is a good default.
Now we are calling `io.Copy` twice. which I don't think we need to.

TianZong48: sure 😸. all done @SomtochiAma

SomtochiAma (approved):

TianZong48:
@stefanprodan  Can I get a review on this?  thanks! 
I’m not sure if you’re the right person to ask, appreciate your time. :)

stefanprodan (approved):
LGTM

Thanks @TianZong48 please rebase with upstream main, squash your commits into a single one and for push.

stefanprodan (approved):

PR #2856, opened 2022-06-21
[RFC-0003] Add commands for managing OCI artifacts

stefanprodan:
This PR adds commands to the Flux CLI for managing OCI artifacts as described in the RFC [Flux OCI support for Kubernetes manifests](https://github.com/fluxcd/flux2/tree/main/rfcs/0003-kubernetes-oci).

## OCI artifact commands

```
$ flux push artifact oci://ghcr.io/org/repository/app-config:v0.0.1 \
	--path="./manifests" \
	--revision="$(git branch --show-current)/$(git rev-parse HEAD)"

$ flux tag artifact oci://ghcr.io/org/repository/app-config:v0.0.1 --tag latest --tag production

$ flux list artifacts oci://ghcr.io/org/repository/app-config

$ flux pull artifact oci://ghcr.io/org/repository/app-config:latest --output ./tmp

$ flux build artifact --path ./manifests --output ./tmp/artifact.tgz
```

For authentication purposes, all `flux <verb> artifact` commands are using the `~/.docker/config.json` config file and the Docker credential helpers.

The artifacts create with `flux push artifact` are annotated with the standard [OCI annotation keys](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys):
- `org.opencontainers.image.source`
- `org.opencontainers.image.revision`
- `org.opencontainers.image.created`

## OCI repository commands

```
$ flux create source oci podinfo-oci \
--url oci://ghcr.io/stefanprodan/manifests/podinfo \
--tag 6.1.6 \
--interval 10m

$ flux create kustomization podinfo-oci \
--source=OCIRepository/podinfo-oci \
--path="./kustomize" \
--prune=true \
--interval=5m \
--target-namespace=default \
--wait=true

$ flux get sources oci
$ flux reconcile source oci podinfo-oci
$ flux suspend source oci podinfo-oci
$ flux resume source oci podinfo-oci
$ flux export source oci podinfo-oci
$ flux delete ks podinfo-oci --silent
$ flux delete source oci podinfo-oci --silent
```


internal/oci/build.go
```diff
@@ -0,0 +1,139 @@
...
+		return fmt.Errorf("invalid source dir path: %s", sourceDir)
+	}
+
+	tf, err := os.CreateTemp(filepath.Split(sourceDir))
+	if err != nil {
+		return err
+	}
+	tmpName := tf.Name()
+	defer func() {
+		if err != nil {
```
souleb, line 42: this evaluate the error when the defer is declared, not sure that's what intended. Maybe a better option is to have a named return parameter

stefanprodan: This is taken from there https://github.com/fluxcd/source-controller/blob/main/controllers/storage.go#L345-L349 why is it better to have a named return parameter?

stefanprodan: Ah I get what you're saying, will fix ty.

internal/oci/meta.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	SourceAnnotation   = "source.toolkit.fluxcd.io/url"
-	RevisionAnnotation = "source.toolkit.fluxcd.io/revision"
+	SourceAnnotation   = "org.opencontainers.image.source"
+	RevisionAnnotation = "org.opencontainers.image.revision"
```
hiddeco, line 25: :100:

action/README.md
```diff
@@ -74,6 +74,86 @@ jobs:
...
+on:
+  push:
+    branches:
+      - 'main'
+
+permissions:
+  packages: write # needed for ghcr.io access
+
+env:
+  OCI_REPO: "oci://ghcr.io/my-org/manifests/${{ github.event.repository.name }}"
```
kingdonb, line 93: There is a surprising amount of nuance in the configuration provided here without much commentary. That's probably the right character for this type of doc, but we may want to add more details.

We probably ought to include a link to some docs that elaborate on how to create a free-standing package repo, or users will trip and may give up when they try using it with an error like this:

Screen Shot 2022-08-09 at 1 26 00 PM

kingdonb:
I will look for somewhere better to put this information, but so as to not lose it for now, since I have written it out, I'll put it here:

Link to the docs that show how to create the free-standing "manifests" package repo and associate it with upstreams that can write OCI artifacts in actions runners from a separate repo.

The best way I know how to do this today is (kludgy, hope there is a better way)

Create a GITHUB_TOKEN (PAT) with `packages:write` and use `docker login` and `flux push` to push a first tag, implicitly creating the package repo. Then go to the page for the package which you can find on your GitHub profile, and set the visibility if needed and the permissions there, to allow actions to write from the source repository, as so:
Screen Shot 2022-08-09 at 1 42 12 PM

Each `manifests/podinfo` or `manifests/...` is a separate package repo in GitHub, so you can keep these permissions isolated and granular, rather than free-for-all "everybody write into this bucket" approach.

You still need to set `permissions.packages: write` as above also. We should explain why this is needed in this example and why it is not needed in the following example, since it's not clear without any comment, or at least point to the docs that explain, here is my impression of why:

Pushes to a branch do not confer "write packages" permission by default in the GitHub permission model, because branches are not releases. Pushes to a tag do confer those permissions because tags are used for releasing.

action/README.md
```diff
@@ -74,6 +74,86 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Flux CLI
+        uses: fluxcd/flux2/action@main
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
kingdonb, line 142: Is there some configuration which makes this block unnecessary in the `production` example? I found I got this failure:

https://github.com/kingdonb/podinfo/runs/7751978215?check_suite_focus=true#step:6:11

(Suggesting to add that to the other workflow in a separate comment)

action/README.md
```diff
@@ -74,6 +74,86 @@ jobs:
...
+  kubernetes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Flux CLI
+        uses: fluxcd/flux2/action@main
+      - name: Generate manifests
+        run: |
+          kustomize build ./manifests/staging > ./deploy/app.yaml
```
kingdonb, line 105: This, or perhaps instead Flux could infer the login from `GITHUB_TOKEN` when the target repo to push is `ghcr.io` - wdyt?

```suggestion
          kustomize build ./manifests/staging > ./deploy/app.yaml
      - name: Login to GHCR
        uses: docker/login-action@v1
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

action/README.md
```diff
@@ -74,6 +74,86 @@ jobs:
...
+
+env:
+  OCI_REPO: "oci://ghcr.io/my-org/manifests/${{ github.event.repository.name }}"
+
+jobs:
+  kubernetes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
kingdonb, line 100:
```suggestion
        uses: actions/checkout@v3
```

action/README.md
```diff
@@ -74,6 +74,86 @@ jobs:
...
+
+env:
+  OCI_REPO: "oci://docker.io/my-org/app-config"
+
+jobs:
+  kubernetes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
kingdonb, line 135:
```suggestion
        uses: actions/checkout@v3
```

cmd/flux/push_artifact.go
```diff
@@ -0,0 +1,113 @@
...
+	source   string
+	revision string
+}
+
+var pushArtifactArgs pushArtifactFlags
+
+func init() {
+	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.path, "path", "", "Path to the directory where the Kubernetes manifests are located.")
+	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.source, "source", "", "The source address, e.g. the Git URL.")
+	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.revision, "revision", "", "The source revision in the format '<branch|tag>/<commit-sha>'")
```
darkowlzz, line 62:
```suggestion
	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.path, "path", "", "path to the directory where the Kubernetes manifests are located")
	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.source, "source", "", "the source address, e.g. the Git URL")
	pushArtifactCmd.Flags().StringVar(&pushArtifactArgs.revision, "revision", "", "the source revision in the format '<branch|tag>/<commit-sha>'")
```

darkowlzz (approved):
LGTM!

PR #2877, opened 2022-06-27
Add the `--branch` arg to the basic auth example

sympatheticmoose:
Without a reference specified, the create command will fail. 

Signed-off-by: David Harris <david.harris@weave.works>

Not tested given nature of the change.


stefanprodan (approved):
LGTM

Thanks @sympatheticmoose for the fix 🏅

PR #2945, opened 2022-07-29
Reset flag after executing command in tests

somtochiama:
This pull request adds more flags to `resetCmdArgs`.
This ensure that flags from one test doesn't affect others.


makkes (approved):
💯 Thanks! This has hit me in the past at least once.

Issue #2949, opened 2022-08-01
flux logs does not select the correct resources


kingdonb:
### Describe the bug

When `flux logs` is used with selectors like `--kind` and `--name` the output is incorrect. Those resources are not returned in the selection. It was a change in v0.31.3 that introduced this, I went back to bisect and test which release causes the issue.

It was not present in v0.31.2

### Steps to reproduce

Try to review the logs for a Flux resource with, for example:

```
flux logs -n flux-system --kind=Kustomization --name=my-secrets
```

for a Flux Kustomization in the default `flux-system` namespace, with the name `my-secrets`, and the results are surprising:

```
2022-08-01T14:56:16.815Z info  - Metrics server is starting to listen
2022-08-01T14:56:16.815Z info  - starting manager
2022-08-01T14:56:16.815Z info  - Starting server
2022-08-01T14:56:16.815Z info  - Starting server
2022-08-01T14:56:16.924Z info HelmRelease - Starting EventSource
2022-08-01T14:56:16.924Z info HelmRelease - Starting EventSource
2022-08-01T14:56:16.924Z info HelmRelease - Starting Controller
2022-08-01T14:56:17.025Z info HelmRelease - Starting workers
2022-08-01T14:56:17.025Z info HelmRelease/public-ingress.ingress-nginx - all dependencies are ready, proceeding with release
2022-08-01T14:56:16.878Z info  - Metrics server is starting to listen
2022-08-01T14:56:16.879Z info  - starting manager
2022-08-01T14:56:16.879Z info  - Starting server
2022-08-01T14:56:16.879Z info  - Starting server
2022-08-01T14:56:16.987Z info ImageUpdateAutomation - Starting EventSource
2022-08-01T14:56:16.987Z info ImageUpdateAutomation - Starting EventSource
2022-08-01T14:56:16.987Z info ImageUpdateAutomation - Starting EventSource
2022-08-01T14:56:16.987Z info ImageUpdateAutomation - Starting Controller
2022-08-01T14:56:17.088Z info ImageUpdateAutomation - Starting workers
2022-08-01T14:56:19.334Z info  - Metrics server is starting to listen
2022-08-01T14:56:19.334Z info  - starting manager
2022-08-01T14:56:19.334Z info  - Starting server
2022-08-01T14:56:19.334Z info  - Starting server
2022-08-01T14:56:19.446Z info ImageRepository - Starting EventSource
2022-08-01T14:56:19.446Z info ImageRepository - Starting Controller
[...]

PR #2950, opened 2022-08-01
Restore function of flux logs with filter selectors engaged

kingdonb:
Fix #2949

An unrelated logic change appears to have been accidentally included with #2851 - this reverts the logic change.

This PR reverts the update to use EqualFold, and the discarding of the
ToLower calls so the function reads as it was when it was approved here:

https://github.com/fluxcd/flux2/commit/8abaa39f9775edd97ae5bc6424cbf872a15cd57c

FYI @somtochiama @stefanprodan

I can try to write some tests around this issue, but for now since the release is broken and I have isolated the fix (I have tested it locally and confirmed it works here) I thought I should send it as-is ASAP.

PR #2955, opened 2022-08-02
fix log filter and add tests for `flux logs`

somtochiama:
Closes: #2949 
Supersedes: #2950


kingdonb:
I see EqualFold does indeed take care of the ToLower bit.

I will take a look at these unit tests, but it looks pretty solid 👍 I guess the timestamps are really no impediment to the golden file approach. Thanks for showing me how this is done! This LGTM +1

stefanprodan (approved):
LGTM 

Thanks @somtochiama

PR #2964, opened 2022-08-08
Add validation to namespace flag

pjbgf:


.github/workflows/e2e.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
           kubectl apply -f https://docs.projectcalico.org/v3.20/manifests/calico.yaml
           kubectl -n kube-system set env daemonset/calico-node FELIX_IGNORELOOSERPF=true
       - name: Setup Kustomize
-        uses: fluxcd/pkg//actions/kustomize@main
+        uses: fluxcd/pkg//actions/kustomize@kustomize/v0.5.3
```
stefanprodan, line 37: Please undo this change as we don't keep the pkg in sync with the action.

pjbgf: This was a test, as it seems that actions within `fluxcd/pkg` are now broken: 
https://github.com/fluxcd/flux2/actions/runs/2816258216

stefanprodan (approved):

PR #3014, opened 2022-08-22
[RFC-0003] Select layer by OCI media type

stefanprodan:
This PR adds an optional field to the `OCIRepository` spec for specifying a media type of the OCI artifact layer which contains the Kubernetes configs. This field is meant to make Flux more flexible when dealing with multi-layer artifacts created by other tools than the Flux CLI.

### DIFF
--- a/rfcs/0003-kubernetes-oci/README.md
+++ b/rfcs/0003-kubernetes-oci/README.md
@@ -4,7 +4,7 @@
 
 **Creation date:** 2022-03-31
 
-**Last update:** 2022-08-11
+**Last update:** 2022-08-22
 
 ## Summary
 
@@ -132,6 +132,18 @@ spec:
     provider: cosign
     secretRef:
       name: cosign-key
+```
+
+### Layer selection
+
+By default, Flux assumes that the first layer of the OCI artifact contains the Kubernetes configuration.
+For multi-layer artifacts created by other tools than Flux CLI, users can specify the media type of the layer
+which contains the tarball with Kubernetes manifests.
+
+```yaml
+spec:
+  layerSelector:
+    mediaType: "application/vnd.cncf.flux.content.v1.tar+gzip"
 ```
 
 ### Pull artifacts from private repositories
--- a/rfcs/0003-kubernetes-oci/README.md
+++ b/rfcs/0003-kubernetes-oci/README.md
@@ -137,14 +137,23 @@ spec:
 ### Layer selection
 
 By default, Flux assumes that the first layer of the OCI artifact contains the Kubernetes configuration.
-For multi-layer artifacts created by other tools than Flux CLI, users can specify the media type of the layer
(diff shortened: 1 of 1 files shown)


rfcs/0003-kubernetes-oci/README.md
```diff
@@ -134,6 +134,18 @@ spec:
       name: cosign-key
 ```
 
+### Layer selection
+
+By default, Flux assumes that the first layer of the OCI artifact contains the Kubernetes configuration.
+For multi-layer artifacts created by other tools than Flux CLI, users can specify the media type of the layer
+which contains the tarball with Kubernetes manifests.
```
hiddeco, line 141:
```suggestion
For multi-layer artifacts created by other tools than Flux CLI, users can specify the media type of the layer
which contains the tarball with Kubernetes manifests. When defined, the first layer matching this type will
be used.
```

To clarify how it behaves for ambiguous matches, as theoretically you can have multiple layers with the same media type.

hiddeco: In practice I think that this will not suffice for a minority of projects which may add things in multiple layers, but I think that could be expanded in the future by adding more knobs to allow for composition.

rfcs/0003-kubernetes-oci/README.md
```diff
@@ -134,6 +134,19 @@ spec:
       name: cosign-key
 ```
 
+### Layer selection
+
+By default, Flux assumes that the first layer of the OCI artifact contains the Kubernetes configuration.
+For multi-layer artifacts created by other tools than Flux CLI, users can specify the media type of the layer
```
makkes, line 140: We should add links to such tools as an example here to clarify the context of this feature.

makkes (approved):

rfcs/0003-kubernetes-oci/README.md
hiddeco: I think it might be better (or besides this) to also refer to the OCI spec around media types.

hiddeco: xref: https://github.com/opencontainers/image-spec/blob/v1.0.2/media-types.md

In addition, we might want to be more explicit about the media type expected to be a "compressed layer" (xref: https://github.com/opencontainers/image-spec/blob/v1.0.2/layer.md#gzip-media-types). As this seems to be implied by https://github.com/google/go-containerregistry/blob/main/pkg/v1/stream/layer.go#L131-L136 which we use in the controller code.

stefanprodan: I think I’ve addressed the comments here, please check the last push.

hiddeco: Looks good to me, thanks Stefan.

darkowlzz (approved):
LGTM!

makkes (approved):

PR #3019, opened 2022-08-22
Improve error message in get cmd

somtochiama:
Before 
```
➜  flux get ks me
✗ no Kustomization objects found in flux-system namespace # misleading as there are Kustomizations in the flux-system namespace
```

Now:
```
➜  ./bin/flux get ks me
✗ Kustomization object 'me' not found in `flux-system` namespace
```


stefanprodan (approved):

hiddeco (commented):
Would it still be worth taking note of the namespace for a single object in the error message?

makkes (approved):

Issue #3016, opened 2022-08-22
flux install is using latest version when version is specified without "v" 


Irindul:
### Describe the bug

When running the command `flux install --version=0.29.0 ...` instead of using `v0.29.0`, the installer will use the latest component (`0.32.0` in my test). 

By looking in the [install cmd](https://github.com/fluxcd/flux2/blob/545949c67f490abea7f5629eb2796048c70ba749/cmd/flux/install.go), in the [ExistingVersion()](https://github.com/fluxcd/flux2/blob/main/pkg/manifestgen/install/install.go#L118) function  this is handled, the v is checked for and added but in the [install.Generate()](https://github.com/fluxcd/flux2/blob/545949c67f490abea7f5629eb2796048c70ba749/pkg/manifestgen/install/install.go#L39) function, the version is passed as is (without the v) and the fetch fails. We can see in the [fetch()](https://github.com/fluxcd/flux2/blob/545949c67f490abea7f5629eb2796048c70ba749/pkg/manifestgen/install/manifests.go#L37) method that there is a faulty check. We check that the version has a prefix "v" but there is no case for when the version is defined without a "v". Thus the fetch method defaults to latest. 

I found this elsewhere in the [isEmbeddedVersion](https://github.com/fluxcd/flux2/blob/5346c1cca31261453f1100b3516a77750e46bb17/cmd/flux/version_utils.go#L56) method, where the input is checked raw (without the v) against a value with the v. 

As this version is used many times, it may be best to prefix it with `v` if it's missing (either directly in the [install cmd](https://github.com/fluxcd/flux2/blob/545949c67f490abea7f5629eb2796048c70ba749/cmd/flux/install.go) or perhaps in the [install.Generate()](https://github.com/fluxcd/flux2/blob/545949c67f490abea7f5629eb2796048c70ba749/pkg/manifestgen/install/install.go#L39), not sure which would be best. 

Thanks 

### Steps to reproduce

1. install flux binary 0.29.0 
2. Run `flux install --version=0.29.2` 

### Expected behavior

The version is parsed properly and contains the proper docker version of the different controllers, and not the latest versions. 

### OS / Distro

macOS

### Flux version

v0.29.0

### Flux check

► checking prerequisites
✗ flux 0.29.0 <0.32.0 (new version is available, please upgrade)
✔ Kubernetes 1.23.8-gke.1900 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
[...]

PR #3024, opened 2022-08-24
Add version validation to install commands

stefanprodan:
Fix: #3016
Fix: #1711

### DIFF
--- a/cmd/flux/version_utils.go
+++ b/cmd/flux/version_utils.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/fluxcd/flux2/internal/utils"
 	"github.com/fluxcd/flux2/pkg/manifestgen/install"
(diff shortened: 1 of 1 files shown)


cmd/flux/version_utils.go
```diff
@@ -28,6 +29,10 @@ func getVersion(input string) (string, error) {
 		return rootArgs.defaults.Version, nil
 	}
 
+	if input != install.MakeDefaultOptions().Version && !strings.HasPrefix(input, "v") {
+		return "", fmt.Errorf("targeted version '%s' must be prefix with 'v'", input)
```
pjbgf, line 33:
```suggestion
		return "", fmt.Errorf("targeted version '%s' must be prefixed with 'v'", input)
```

pjbgf (changes requested):
LGTM

thanks @stefanprodan! :bow:

cmd/flux/version_utils.go
stefanprodan: Thanks!

pjbgf (approved):

PR #3025, opened 2022-08-24
[RFC-0002] Add auth specification for Helm OCI

stefanprodan:
Add basic auth and contextual login for AWS, Azure, GCP to the Helm OCI RFC.

### DIFF
--- a/rfcs/0002-helm-oci/README.md
+++ b/rfcs/0002-helm-oci/README.md
@@ -4,7 +4,7 @@
 
 **Creation date:** 2022-03-30
 
-**Last update:** 2022-06-07
+**Last update:** 2022-08-24
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


rfcs/0002-helm-oci/README.md
```diff
@@ -37,7 +37,43 @@ When not specified, the `spec.type` field defaults to `default` which preserve t
...
+the credentials can be supplied with:
+
+```yaml
+spec:
+  secretRef:
+    name: regcred
+```
+
+The `secretRef` points to a Kubernetes secret in the same namespace as the `HelmRepository`.
+The secret type must be `kubernetes.io/dockerconfigjson`:
```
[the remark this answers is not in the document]
hiddeco: or maybe even a specific link to xref https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets

rfcs/0002-helm-oci/README.md
```diff
@@ -37,7 +37,44 @@ When not specified, the `spec.type` field defaults to `default` which preserve t
...
+can be used to bind the `source-controller` to the IAM role.
+
+```yaml
+spec:
+  provider: azure
+```
+
+The provider accepts the following values: `generic`, `aws`, `azure` and `gcp`. When the provider is
+not specified, it defaults to `generic`. When the provider is set to `aws`, `azure` or `gcp`, the
+controller will use a specific cloud SDK for authentication purposes.
```
pjbgf, line 77:
```suggestion
controller will use a specific cloud SDK for the authentication purposes.
```

rfcs/0002-helm-oci/README.md
```diff
@@ -37,7 +37,44 @@ When not specified, the `spec.type` field defaults to `default` which preserve t
...
+
+### Pull charts from private repositories
+
+#### Basic auth
+
+For private repositories hosted on GitHub, Quay, self-hosted Docker Registry and others,
+the credentials can be supplied with:
+
+```yaml
+spec:
```
makkes, line 49:
```suggestion
kind: HelmRepository
[...]
spec:
```
This makes the context of this spec crystal clear.

rfcs/0002-helm-oci/README.md
```diff
@@ -37,7 +37,44 @@ When not specified, the `spec.type` field defaults to `default` which preserve t
...
+  --docker-password=<your-pword>
+```
+
+#### OIDC auth
+
+When Flux runs on AKS, EKS or GKE, an IAM role (that grants read-only access to ACR, ECR or GCR)
+can be used to bind the `source-controller` to the IAM role.
+
+```yaml
+spec:
```
makkes, line 71:
```suggestion
kind: HelmRepository
[...]
spec:
```
We should explicitly call out that this is a new field being added as part of this proposal. We do that at the top of this section for `.spec.type`, too.

rfcs/0002-helm-oci/README.md
stefanprodan: This doesn't sound english to me, maybe we should get a native speaker to decide on this.

pjbgf: I think you are right, let's keep as is.

rfcs/0002-helm-oci/README.md
stefanprodan: Added, please see the latest changes.

makkes (approved):

rfcs/0002-helm-oci/README.md
```diff
@@ -32,12 +32,61 @@ they do today for container images.
...
+metadata:
+  name: <repo-name>
+spec:
+  type: oci
+  provider: azure
+```
+
+The provider accepts the following values: `generic`, `aws`, `azure` and `gcp`. When the provider is
+not specified, it defaults to `generic`. When the provider is set to `aws`, `azure` or `gcp`, the
+controller will use a specific cloud SDK for authentication purposes.
```
darkowlzz, line 89: Can append a similar sentence from OCI Repo RFC here https://github.com/fluxcd/flux2/blob/9f26b09a06678a2e913556734adb4f58c996def1/rfcs/0003-kubernetes-oci/README.md?plain=1#L223-L225

In addition, maybe we need to also make it clear that non-OCI HelmRepository will just ignore the `provider` field.

darkowlzz:
Maybe off-topic but felt like it's related to authentication in a way, the change in https://github.com/fluxcd/source-controller/pull/856 to allow authenticating with the registry domain without path to chart, can we also document that here? Or even separately if it's off-topic. I'm afraid that it's not documented anywhere and may become a hidden feature.

stefanprodan:
>  I'm afraid that it's not documented anywhere and may become a hidden feature.

@darkowlzz I think this should be in the API spec doc.

souleb:
Reading this:
> When not specified, the spec.type field defaults to default which preserve the current HelmRepository API behaviour.

We don't actually have this in the code, we would need to add `// +kubebuilder:default:=default`.

darkowlzz (approved):
LGTM!

stefanprodan:
@souleb we do have the default implemented in code but not in CRD, which is fine.

Issue #2203, opened 2021-12-09
Broken "edit this page" links in Flux CLI section


SamLR:
The "edit this page" links in the right-hand sidebar go to 404 github pages.

I assume this is a hangover of the move to auto-generating these pages from the binary's `help` output

## Steps to reproduce

1. Navigate to any page in the 'Flux CLI' section
2. Click on 'Edit this page' in the right hand side bar

For example, under ['Flux CLI' > 'Flux'](https://fluxcd.io/docs/cmd/flux/) 'Edit this page' links to https://github.com/fluxcd/website/edit/main/content/en/docs/cmd/flux.md which is a 404. 

## Potential solutions

(I have no idea how feasible any of these are, sorry)

* Disable this link on these pages
* Link to the appropriate file in https://github.com/fluxcd/flux2

dholbach:
This
  
```md
importedDoc: true
```

needs to be added in

https://github.com/fluxcd/flux2/blob/37b60666c439dc823af33dc35162de38d5380883/cmd/flux/docgen.go#L59

michaelmwangi:
Hi I would like to work on this if no one has picked it up yet. Thanks

ilanpillemer:
@dholbach if I send a PR would you review it?

dholbach:
@ilanpillemer For sure, yes! <3

PR #3034, opened 2022-08-25
Fix broken "edit this page" links in Flux CLI section

snebel29:
This MR closes https://github.com/fluxcd/flux2/issues/2203

As described in the original issue, the fix to disable the source page related links is pretty simple by adding `importedDoc: true` within the header.

It took me a while to realise where this was exactly needed, I was unable to find documentation about that option and another while to test that the change was good.

### Validation procedure

I did checkout https://github.com/fluxcd/website then I ran
```shell
$ make serve
```
Which generated the site and served it on http://localhost:1313 then navigated to http://localhost:1313/docs/cmd/flux/ and observed the "Edit this page" along with the other source page links.

image

After that edited `content/en/docs/cmd/flux.md` by adding `importedDoc: true` under title.
```markdown
---
title: "flux"
importedDoc: true
---
```
then copy paste the `hugo server` from the previous `make serve` run to avoid overriding my change.
```shell
$ hugo server \
                --bind 127.0.0.1 \
                --buildDrafts \
                --buildFuture \
                --disableFastRender
```
And observed the links related to the source page (view page source, edit this page and create child page) disappear.

image

I think this change is good to go.

### DIFF
--- a/cmd/flux/docgen.go
+++ b/cmd/flux/docgen.go
@@ -28,6 +28,7 @@ import (
 
 const fmTemplate = `---
 title: "%s"
+importedDoc: true
 ---
 `
 


stefanprodan (approved):
LGTM

Thanks @snebel29 🏅

PR #3048, opened 2022-08-29
Update packages in Azure e2e tests

stefanprodan:
Update Flux API packages and the Azure SDK ones. With these changes the Azure conformance tests should be functional again.


darkowlzz (approved):
LGTM!
The changes are similar to what we do in the azure terraform modules in the test-infra repository, refer: https://github.com/fluxcd/test-infra/tree/65e1a901cbb9b3f9f27ffad7f9a32a6366eae1cc/tf-modules/azure/aks.

PR #3049, opened 2022-08-29
Update Kubernetes dependencies to v1.25.0

stefanprodan:
Changes:
- update `k8s.io` packages to match the Kubernetes v1.25.0 release
- update `kubectl` to v1.25.0 in the `flux-cli` container image
- update `go.mod` to Go 1.18
- update `fluxcd/pkg` dependencies


darkowlzz (approved):
LGTM!
Would you also like to update kubectl in github self-hosted runner in https://github.com/fluxcd/flux2/blob/735ebd3336e6904d27d53ee28bc43c2789e69267/.github/runners/prereq.sh#L22 separately or part of this?

stefanprodan:
@darkowlzz for the runners I would do a separate PR as many things are behind latest.

Issue #3051, opened 2022-08-30
Align to Kubernetes structured logging


stefanprodan:
We've been delaying the upgrade to controller-runtime v0.12 due to the changes to structured logging. I think it's time to align the Flux controllers logs with Kubernetes, even if the new logging format is more verbose.

Current format (flux2 v0.33.0):
```json
{
  "level": "info",
  "ts": "2022-08-23T09:35:20.723+0300",
  "msg": "no changes since last reconciliation",
  "logger": "controller.gitrepository",
  "reconciler group": "source.toolkit.fluxcd.io",
  "reconciler kind": "GitRepository",
  "namespace": "default",
  "name": "podinfo"
}
```

New format:
```json
{
  "level": "info",
  "ts": "2022-08-23T09:35:20.723+0300",
  "msg": "no changes since last reconciliation",
  "controller": "gitrepository",
  "controllerGroup": "source.toolkit.fluxcd.io",
  "controllerKind": "GitRepository",
  "gitRepository": {
    "name": "podinfo",
    "namespace": "default"
  },
  "namespace": "default",
  "name": "podinfo",
  "reconcileID": "aef58d3d-c155-4d45-a4b0-af35cce3f978"
}
```

## Breaking changes

These fields have been renamed:

- `logger` -> `controller`
- `reconciler group` -> `controllerGroup`
- `reconciler kind` -> `controllerKind`

### Prerequisites

- [x] `fluxcd/pkg`: update all packages to controller-runtime v0.12
- [x] `fluxcd/pkg/runtime`: release new Flux runtime package based on controller-runtime v0.12

### Controllers

Update `fluxcd/pkg/runtime` and remove our custom `reconcileID` in favour of the upstream one.

- [x] `fluxcd/notification-controller` (depends on prerequisites)
- [x] `fluxcd/image-reflector-controller` (depends on prerequisites)
- [x] `fluxcd/source-controller` (depends on prerequisites)
- [x] `fluxcd/kustomize-controller` (depends on `fluxcd/source-controller`)
- [x] `fluxcd/helm-controller` (depends on `fluxcd/source-controller`)
- [x] `fluxcd/image-automation-controller` (depends on `fluxcd/source-controller` and `fluxcd/image-reflector-controller`)

### CLI

- [x] `fluxcd/flux2`: adapt `flux logs` to the new format in a backwards compatible manner ( support both `reconciler kind` and `controllerKind`)

PR #3063, opened 2022-08-31
Update `flux logs` to accomodate the new format 

somtochiama:
`flux logs` has been updated to support both the new and old log format that comes with controller-runtime v1.12.3.
The output from flux logs is still the same

For more details on the new logging structure please see: #3051.


stefanprodan (approved):
LGTM

Thanks @somtochiama 🏅

PR #3079, opened 2022-09-07
Support autologin when pushing OCI artifacts

adrien-f:
I've noticed during CI, that the current command already expected a configured Docker client to push artifacts to authenticated registries.

Some users might not want to have the Docker client in their process (like a CI job) or build a handcrafted config.json file.

This MR adds `--creds` and `--provider` to the `push artifact` command.

`--creds` can accept either a token that will be used with the `Bearer` keyword, or a username & password which will be used for Basic authentication.

`--provider` will run the provider login logic.

This would allow this kind of behavior:

```
flux push artifact oci://my-registry.dev/foo:v1 \
  --source xxx \
  --revision xxx \
  --path . \
  --creds $TOKEN # Authenticate via "Bearer $TOKEN" Authorization header
```

Or via provider:

```
flux push artifact oci://012345678901.dkr.ecr.us-east-1.amazonaws.com/foo:v1 \
  --source xxx \
  --revision xxx \
  --path . \
  --provider aws
```

This depends on another PR in pkg: https://github.com/fluxcd/pkg/pull/352 

Signed-off-by: Adrien Fillon <adrien.fillon@manomano.com>

### DIFF
--- a/cmd/flux/push_artifact.go
+++ b/cmd/flux/push_artifact.go
@@ -19,8 +19,11 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/spf13/cobra"
 	"os"
+
+	"github.com/fluxcd/flux2/internal/flags"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/spf13/cobra"
 
 	oci "github.com/fluxcd/pkg/oci/client"
 )
@@ -29,7 +32,7 @@ var pushArtifactCmd = &cobra.Command{
 	Use:   "artifact",
 	Short: "Push artifact",
 	Long: `The push artifact command creates a tarball from the given directory and uploads the artifact to an OCI repository.
-The command uses the credentials from '~/.docker/config.json'.`,
+The command can read the credentials from '~/.docker/config.json' but they can also be passed with --creds. It can also login to a supported provider with the --provider flag.`,
 	Example: `  # Push manifests to GHCR using the short Git SHA as the OCI artifact tag
   echo $GITHUB_PAT | docker login ghcr.io --username flux --password-stdin
   flux push artifact oci://ghcr.io/org/config/app:$(git rev-parse --short HEAD) \
@@ -43,6 +46,22 @@ The command uses the credentials from '~/.docker/config.json'.`,
 	--path="./path/to/local/manifests" \
 	--source="$(git config --get remote.origin.url)" \
 	--revision="$(git tag --points-at HEAD)/$(git rev-parse HEAD)"
+
+  # Login directly to the registry provider
+  # You might need to export the following variable if you use local config files for AWS:
+  # export AWS_SDK_LOAD_CONFIG=1
+  flux push artifact oci://<account>.dkr.ecr.<region>.amazonaws.com/foo:v1:$(git tag --points-at HEAD) \
+	--path="./path/to/local/manifests" \
+	--source="$(git config --get remote.origin.url)" \
+	--revision="$(git tag --points-at HEAD)/$(git rev-parse HEAD)" \
+	--provider aws
+
+  # Or pass credentials directly
+  flux push artifact oci://docker.io/org/app-config:$(git tag --points-at HEAD) \
+	--path="./path/to/local/manifests" \
+	--source="$(git config --get remote.origin.url)" \
+	--revision="$(git tag --points-at HEAD)/$(git rev-parse HEAD)" \
+	--creds flux:$DOCKER_PAT
 `,
 	RunE: pushArtifactCmdRun,
 }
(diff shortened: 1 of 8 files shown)


cmd/flux/push_artifact.go
```diff
@@ -101,8 +129,39 @@ func pushArtifactCmdRun(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
 	defer cancel()
 
+	var ociClientOpts []crane.Option
+
+	if pushArtifactArgs.creds != "" {
```
somtochiama, line 134: Hey @adrien-f , Thanks for this pull request🎉!

I think that most of this logic should be in `pkg/oci` so that it can be reused across the other oci commands, maybe pass the token/auth options to `NewLoclaClient` instead. The tests for this would also be added there.
I think the `--auto-login` thing is pretty neat

cc: @stefanprodan

stefanprodan: Yes besides `flux push` there are other commands which make use of the OCI client. I guess @adrien-f didn't run `make build` or `make vet` or `make test` all of those would've fail. Instead of `--auto-login` I would prefer using the `--provider` flag from here https://github.com/fluxcd/flux2/blob/main/cmd/flux/create_source_oci.go#L74

adrien-f: Hi and thanks for the review 👋 

I've pushed a new change that moved that logic to the OCI client, let me know!

Regarding the change in the method arguments, AFAIK, adding variadic arguments does not break the existing usage of the function: https://go.dev/play/p/UTWaaDYbwv-

stefanprodan: What do you think or replacing --auto-login with --provider?

adrien-f: I don't mind either, the auto-login allowed  depending on the end-user flow and the UX of Flux (that I'm still discovering)

Code-wise, in `pkg/oci` that would mean adding a switch over the provider value, then initiating a client for each supported registries, then calling the login function for each of them which would look like this code already present: https://github.com/fluxcd/pkg/blob/main/oci/auth/login/login.go#L97 so I'm not sure if it's worth to duplicate this feature/logic.

```go
func (m *Manager) Login(ctx context.Context, image string, ref name.Reference, opts ProviderOptions) (authn.Authenticator, error) {
	switch ImageRegistryProvider(image, ref) {
	case oci.ProviderAWS:
		return m.ecr.Login(ctx, opts.AwsAutoLogin, image)
	case oci.ProviderGCP:
		return m.gcr.Login(ctx, opts.GcpAutoLogin, image, ref)
	case oci.ProviderAzure:
		return m.acr.Login(ctx, opts.AzureAutoLogin, image, ref)
	}
	return nil, nil
}
```

stefanprodan: It feels really strange to ask for a specific provide everywhere in Flux except for this command.

adrien-f: I've noticed that the provider enum is not the same from the CLI and in the pkg/oci:

* https://github.com/fluxcd/flux2/blob/08401f62b2ad7b4573cb85847a56c35288aa530f/internal/flags/source_oci_provider.go#L28
* https://github.com/fluxcd/source-controller/blob/main/api/v1beta2/ocirepository_types.go#L27

vs

https://github.com/fluxcd/pkg/blob/770524e58984e6165ae7e2fc92da491ec4b1b7c1/oci/constants.go#L23

Which makes interop a bit tricky, I keep on digging 🕵️

adrien-f: I've updated the PR to add a mapping between the two types with additional logic in pkg/oci 👍

stefanprodan: Looks great! Thank you. Can you please add the 2 flags to tag|pull|list artifact(s)

cmd/flux/push_artifact.go
```diff
@@ -43,6 +46,22 @@ The command uses the credentials from '~/.docker/config.json'.`,
 	--path="./path/to/local/manifests" \
 	--source="$(git config --get remote.origin.url)" \
 	--revision="$(git tag --points-at HEAD)/$(git rev-parse HEAD)"
+
+  # Login directly to the registry provider
+  # You might need to export the following variable if you use local config files for AWS:
+  # export AWS_SDK_LOAD_CONFIG=1
+  flux push artifact oci://012345678901.dkr.ecr.us-east-1.amazonaws.com/foo:v1:$(git tag --points-at HEAD) \
```
stefanprodan, line 53:
```suggestion
  flux push artifact oci://<account>.dkr.ecr.<region>.amazonaws.com/foo:$(git tag --points-at HEAD) \
```

cmd/flux/push_artifact.go
```diff
@@ -43,6 +46,22 @@ The command uses the credentials from '~/.docker/config.json'.`,
...
+	--path="./path/to/local/manifests" \
+	--source="$(git config --get remote.origin.url)" \
+	--revision="$(git tag --points-at HEAD)/$(git rev-parse HEAD)" \
+	--provider aws
+
+  # Or pass credentials directly
+  flux push artifact oci://docker.io/org/app-config:v1:$(git tag --points-at HEAD) \
```
stefanprodan, line 60:
```suggestion
  flux push artifact oci://docker.io/org/app-config:$(git tag --points-at HEAD) \
```

stefanprodan:
@adrien-f I've released your changes to the oci package, please update the version in go.mod to` v0.9.0` then run `make tidy`.

adrien-f:
@stefanprodan Hi Stefan, hope the weekend was good!

There you go, I've updated the `go.mod`, it builds properly 🎉

stefanprodan (changes requested):
The auth package broke CI because the Azure SDK requires Go 1.18. To fix this please add these lines https://github.com/fluxcd/source-controller/blob/main/.github/workflows/scan.yaml#L34-L37 to the CodeQL job here https://github.com/fluxcd/flux2/blob/main/.github/workflows/scan.yaml#L51

stefanprodan (approved):
LGTM

Thanks @adrien-f for this great UX improvement to OCI 🥇

PR #3081, opened 2022-09-08
RFC-0004: Block insecure HTTP connections across Flux

### LINKED ISSUE #3076: RFC - Consolidate use of insecure endpoints across controllers
Across Flux controllers the use of non-TLS based connections is not consistent.

Controllers that deals only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
Some Flux objects provide an `.Spec.Insecure` field to enable the use of non-TLS based endpoints, but they don't clearly notify users when the option is not supported (e.g. Azure/GCP Buckets).

A new RFC should be created to establish:
- Controller-level ways to block the use of HTTP endpoints. When enabled, the flag would not allow the use of the non-TLS based connections across all controller-level objects.
- Consistent way to support insecure endpoints and to notify when they are not supported (or it has been blocked at controller level).
- Naming conventions covering flags and object fields.

Related Implementations:
- https://github.com/fluxcd/source-controller/pull/828
- https://github.com/fluxcd/notification-controller/pull/404

aryan9600:
Proposal for blocking HTTP connections across all Flux controllers and objects

TLDR:
* A new boolean flag `--insecure-allow-http` shall be added to all controllers, which would be used to allow/disallow HTTP connections at the controller level. The default value is `true`, which _allows_ HTTP connections
* A new boolean field `.spec.insecure` shall be added to all relevant objects where the specified endpoint doesn't allow for inference of the transport type through the scheme and HTTP connections must be used.
* A new boolean flag `--insecure` shall be added to relevant CLI commands, to configure the value of `.spec.insecure` in the related object definition.

Sponsored by @pjbgf 

Fixes #3076

### DIFF
--- a/rfcs/0004-insecure-http/README.md
+++ b/rfcs/0004-insecure-http/README.md
@@ -83,6 +83,20 @@ that the registry is hosted at a non-TLS endpoint.
 For such objects, we shall introduce a new boolean field `.spec.insecure`, which shall be `false` by default. Users that
 need their object to point to an HTTP endpoint, can set this to `true`.
 
+### CLI
+The Flux CLI offers several commands for creating Flux specific resources. Some of these commands may involve specifying
+an endpoint such as creating an `OCIRepository`:
+
+```sh
+ flux create source oci podinfo \
+    --url=oci://ghcr.io/stefanprodan/manifests/podinfo \
+    --tag=6.1.6 \
+    --interval=10m
+```
+
+Since these commands essentially create object definitions, the CLI should offer a boolean flag `--insecure`
+for relevant objects, which will be used for specifying the value of `.spec.insecure` of such objects.
+
 ### Precedence & Validity
 Objects with `.spec.insecure` as `true ` will only be allowed if HTTP connections are allowed at the controller level.
 Similarly, an object can have `.spec.insecure` as `true` only if the Saas/Cloud provider allows HTTP connections.
@@ -137,9 +151,9 @@ all objects have `.spec.insecure` as `false` and any URLs present in the definit
 as the scheme. This is less attractive, as this would ask users to install another software and prevent
 Flux multi-tenancy from being standalone.
 
-## Design Details 
+## Design Details
 If a controller is started with `--insecure-allow-http=false`, any URL in a Flux object which has `http`
-as the scheme will result in an error and the following condition will be added to the object's 
+as the scheme will result in an error and the following condition will be added to the object's
 `.status.conditions`:
 
 ```yaml
--- a/rfcs/0004-insecure-http/README.md
+++ b/rfcs/0004-insecure-http/README.md
@@ -89,3 +89,50 @@ Similarly, an object can have `.spec.insecure` as `true` only if the Saas/Cloud 
 For example, using a `Bucket` with its `.spec.provider` set to `azure` would be invalid since Azure doesn't allow
 HTTP connections.
 
+
+### User Stories
+
+#### Story 1
+> As a cluster admin of a multi-tenant cluster, I want to ensure all controllers access endpoints using only HTTPS
+> regardless of tenants' object definitions.
+
+Apply a `kustomize` patch which prevents the use of HTTP connections:
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - gotk-components.yaml
+  - gotk-sync.yaml
+patches:
+  - patch: |
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --allow-insecure-http=false
+    target:
+      kind: Deployment
+      name: "(kustomize-controller|helm-controller|source-controller|notification-controller)"
+```
+
+#### Story 2
+> As an application developer, I'm trying to debug a new image pushed to my local registry which
+> is not served over HTTPS.
+
+Modify the object spec to use HTTP connections explicitly:
+```yaml
+apiVersion: image.toolkit.fluxcd.io/v1beta1
+kind: ImageRepository
+metadata:
+  name: podinfo
+  namespace: flux-system
+spec:
+  image: kind-registry:5000/stefanprodan/podinfo
+  interval: 1m0s
+  insecure: true
+```
+
+### Alternatives
+Instead of adding a flag, we can instruct users to make use of Kyverno policies to enforce that
+all objects have `.spec.insecure` as `false` and any URLs present in the definition don't have `http`
+as the scheme. This is less attractive, as this would ask users to install another software and prevent
+Flux multi-tenancy from being standalone.
--- a/rfcs/0004-insecure-http/README.md
+++ b/rfcs/0004-insecure-http/README.md
@@ -5,15 +5,17 @@
 **Creation Date:** 2022-09-08
 
 ## Summary
+
 Flux should have a consistent way of disabling insecure HTTP connections.
 
 At the controller level, a flag should be present which would disable all outgoing HTTP connections.
 At the object level, a field should be provided which would enable the use of non-TLS endpoints.
 
-If the use of a non-TLS endpoint is not supported, it should be made clear to users through the use of
-logs and status conditions.
+If the use of a non-TLS endpoint is not supported, reconciliation will fail and the object will be marked
+as stalled, signalling that human intervention is required.
 
 ## Motivation
+
 Today the use of non-TLS based connections is inconsistent across Flux controllers.
 
 Controllers that deal only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
@@ -21,17 +23,20 @@ Some Flux objects provide a `.spec.insecure` field to enable the use of non-TLS 
 users when the option is not supported (e.g. Azure/GCP Buckets).
 
 ### Goals
+
 * Provide a flag across relevant Flux controllers which disables all outgoing HTTP connections.
 * Add a field which enables the use of non-TLS endpoints to appropriate Flux objects.
 * Provide a way for users to be made aware that their use of non-TLS endpoints is not supported if that is the case.
 
 ### Non-Goals
+
 * Break Flux's current behavior of allowing HTTP connections.
 * Change in behavior of communication between Flux components.
 
 ## Proposal
 
 ### Controllers
+
 Flux users should be able to enforce that controllers are using HTTPS connections only.
 This shall be enabled by adding a new boolean flag `--insecure-allow-http` to the following controllers:
 * source-controller
@@ -60,27 +65,24 @@ spec:
           - --insecure-allow-http=false
 ```
 
-> Note: The flag shall not be added to the following controllers:
-> * kustomize-controller: This flag is excluded from this controller, as the upstream `kubenetes-sigs/kustomize` project
-> does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
-> upstream project adds support for this at a later point in time.
-> * helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
-> Furthermore although both controllers can also do remote applies, serving `kube-apiserver` over plain
-> HTTP is disabled by default. While technically this can be enabled, the option for this configuration was also disabled
-> quite a while back (ref: https://github.com/kubernetes/kubernetes/pull/65830/).
-
-Both kustomize-controller and helm-controller currently have a flag `--insecure-kubeconfig-tls` which makes the controller skip
-TLS verification when connecting to a Kubernetes cluster with an HTTPS connection. This flag shall be renamed to
-`--insecure-skip-tls-verify` to align it with the Flux CLI which offers this command for the same purpose.
+**Note:** The flag shall not be added to the following controllers:
+* kustomize-controller: This flag is excluded from this controller, as the upstream `kubenetes-sigs/kustomize` project
+does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
+upstream project adds support for this at a later point in time.
+* helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
+Furthermore although both controllers can also do remote applies, serving `kube-apiserver` over plain
+HTTP is disabled by default. While technically this can be enabled, the option for this configuration was also disabled
+quite a while back (ref: https://github.com/kubernetes/kubernetes/pull/65830/).
 
 ### Objects
+
 Some Flux objects, like `GitRepository`, provide a field for specifying a URL, and the URL would contain the scheme.
 In such cases, the scheme can be used for inferring the transport type of the connection and consequently,
 whether to use HTTP or HTTPS connections for that object.
 But there are a few objects that don't allow such behavior, for example:
 
-* `ImageRepository`: It provides a field, `.spec.image`, which is used for specifying the URL of the image present on
-a container registry. But any URL containing a scheme is considered invalid and HTTPS is the default transport used.
+* `ImageRepository`: It provides a field, `.spec.image`, which is used for specifying the address of the image present on
+a container registry. But any address containing a scheme is considered invalid and HTTPS is the default transport used.
 This prevents users from using images present on insecure registries.
 * OCI `HelmRepository`: When using an OCI registry as a Helm repository, the `.spec.url` is expected to begin with `oci://`.
 Since the scheme part of the URL is used to specify the type of `HelmRepository`, there is no way for users to specify
@@ -90,6 +92,7 @@ For such objects, we shall introduce a new boolean field `.spec.insecure`, which
 need their object to point to an HTTP endpoint, can set this to `true`.
 
 ### CLI
+
 The Flux CLI offers several commands for creating Flux specific resources. Some of these commands may involve specifying
 an endpoint such as creating an `OCIRepository`:
 
@@ -107,6 +110,7 @@ for the required commands, which will be used for specifying the value of `.spec
 > when using an HTTPS connection.
 
 ### Precedence & Validity
+
 Objects with `.spec.insecure` as `true ` will only be allowed if HTTP connections are allowed at the controller level.
 Similarly, an object can have `.spec.insecure` as `true` only if the Saas/Cloud provider allows HTTP connections.
 For example, using a `Bucket` with its `.spec.provider` set to `azure` would be invalid since Azure doesn't allow
@@ -115,6 +119,7 @@ HTTP connections.
 ### User Stories
 
 #### Story 1
+
 > As a cluster admin of a multi-tenant cluster, I want to ensure all controllers access endpoints using only HTTPS
 > regardless of tenants' object definitions.
 
@@ -134,8 +139,8 @@ patches:
     target:
       kind: Deployment
       name: "(source-controller|notification-controller|image-reflector-controller|image-automation-controller)"
-  # Since this above flag is not available in kustomize-controller for reasons explained in a previous section,
-  # we disable the Kustomize remote build by disallowing use of remote bases. This ensures that kustomize-controller
+  # Since the above flag is not available in kustomize-controller for reasons explained in a previous section,
+  # we disable Kustomize remote builds by disallowing use of remote bases. This ensures that kustomize-controller
   # won't initiate any plain HTTP connections.
   - patch: |
       - op: add
@@ -147,6 +152,7 @@ patches:
 ```
 
 #### Story 2
+
 > As an application developer, I'm trying to debug a new image pushed to my local registry which
 > is not served over HTTPS.
 
@@ -164,14 +170,16 @@ spec:
 ```
 
 ### Alternatives
+
 Instead of adding a flag, we can instruct users to make use of Kyverno policies to enforce that
 all objects have `.spec.insecure` as `false` and any URLs present in the definition don't have `http`
 as the scheme. This is less attractive, as this would ask users to install another software and prevent
 Flux multi-tenancy from being standalone.
 
 ## Design Details
+
 If a controller is started with `--insecure-allow-http=false`, any URL in a Flux object which has `http`
-as the scheme will result in an error and the following condition will be added to the object's
+as the scheme will result in an unsuccessful reconciliation and the following condition will be added to the object's
 `.status.conditions`:
 
 ```yaml
@@ -180,13 +188,13 @@ status:
   - lastTransitionTime: "2022-09-06T09:14:21Z"
     message: "Use of insecure HTTP connections isn't allowed for this controller"
     observedGeneration: 1
-    reason: URLInvalid
+    reason: InsecureConnectionsDisallowed
     status: "True"
-    type: FetchFailedCondition
+    type: Stalled
 ```
 
 Similarly, if an object has `.spec.insecure` as `true` but the Cloud provider doesn't allow HTTP connections,
-the reconciler will error out and add the condition below to the object's `.status.conditions`:
+the reconciliation will fail and the following condition will be added to the object's `.status.conditions`:
 
 ```yaml
 status:
@@ -194,9 +202,9 @@ status:
   - lastTransitionTime: "2022-09-06T09:14:21Z"
     message: "Use of insecure HTTP connections isn't allowed for Azure Storage"
     observedGeneration: 1
-    reason: InsecureConnectionsDisallowed
+    reason: UnsupportedConnectionType
     status: "True"
-    type: FetchFailedCondition
+    type: Stalled
 ```
 
 If an object has `.spec.insecure` as `true`, the registry client or bucket client shall be created with the use
--- a/rfcs/0004-insecure-http/README.md
+++ b/rfcs/0004-insecure-http/README.md
@@ -1,0 +1,91 @@
+# RFC-0004 Block insecure HTTP connections across Flux
+
+**Status:** provisional
+
+**Creation Date:** 2022-09-08
+
+## Summary
+Flux should have a consistent way of disabling insecure HTTP connections.
+
+At the controller level, a flag should be present which would disable all outgoing HTTP connections.
+At the object level, a field should be provided which would enable the use of non-TLS endpoints.
+
+If the use of a non-TLS endpoint is not supported, it should be made clear to users through the use of
+logs and status conditions.
+
+## Motivation
+Today the use of non-TLS based connections is inconsistent across Flux controllers.
+
+Controllers that deal only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
+Some Flux objects provide a `.spec.insecure` field to enable the use of non-TLS based endpoints, but they don't clearly notify users when the option is not supported (e.g. Azure/GCP Buckets).
+
+### Goals
+* Provide a flag across all Flux controllers which disables all outgoing HTTP connections.
+* Add a field which enables the use of non-TLS endpoints to appropriate Flux objects.
+* Provide a way for users to be made aware that their use of non-TLS endpoints is not supported if that is the case.
+
+### Non-Goals
+* Break Flux's current behavior of allowing HTTP connections.
+
+## Proposal
+
+### Controllers
+Flux users should be able to enforce that controllers are using HTTPS connections only.
+This shall be enabled by adding a new boolean flag `--allow-insecure-http` to the following controllers:
+* source-controller
+* notification-controller
+* image-automation-controller
+* image-reflector-controller
+
+> Note: The flag shall not be added to the following controllers:
+> * kustomize-controller: This flag is excluded from this controller, as the upstream `kubenetes-sigs/kustomize` project
+> does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
+> upstream project adds support for this at a later point in time.
+> * helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
+> Furthermore although both controllers can also do remote applies, serving `kube-apiserver` over plain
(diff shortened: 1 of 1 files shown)


stefanprodan (commented):
Please document the use of insecure for Alert Providers, ref: https://github.com/fluxcd/notification-controller/pull/404 

Besides adding the flag to notification-controller, I guess the Provider CRD should have the `insecure` field as well.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,173 @@
...
+  - gotk-components.yaml
+  - gotk-sync.yaml
+patches:
+  - patch: |
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --allow-insecure-http=false
+    target:
+      kind: Deployment
+      name: "(kustomize-controller|helm-controller|source-controller|notification-controller)"
```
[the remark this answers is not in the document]
aryan9600: Agree with you on helm-controller.
Regarding kustomize-controller, the default behavior is to allow remote bases, so admins might want to make sure that only HTTPS is used to access those remote bases.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,173 @@
...
+At the controller level, a flag should be present which would disable all outgoing HTTP connections.
+At the object level, a field should be provided which would enable the use of non-TLS endpoints.
+
+If the use of a non-TLS endpoint is not supported, it should be made clear to users through the use of
+logs and status conditions.
+
+## Motivation
+Today the use of non-TLS based connections is inconsistent across Flux controllers.
+
+Controllers that deals only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
```
hiddeco, line 19:
```suggestion
Controllers that deal only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
```

rfcs/0004-insecure-http/README.md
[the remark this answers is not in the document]
stefanprodan: We'll need to fork kustomize to add this filter, instead I propose we recommend users to disable remote bases all together as they are insecure by default HTTPS or nor.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Since these commands essentially create object definitions, the CLI should offer a boolean flag `--insecure`
+for relevant objects, which will be used for specifying the value of `.spec.insecure` of such objects.
+
+### Precedence & Validitiy
+Objects with `.spec.insecure` as `true `will only be allowed if HTTP connections are allowed at the controller
+level.
+Similarly, an object can have`.spec.insecure` as `true` only if the Saas/Cloud provider allows HTTP connections.
+For example using a `Bucket` with its `.spec.provider` set to `azure` would be invalid, since Azure doesn't allow
+HTTP connections.
```
pjbgf, line 91: The Azure SDK supports HTTP access:
https://github.com/Azure/azure-sdk-for-go/blob/9b91fa8fd891a3364779d8d32f2a13dfa98dc6f4/storage/client.go#L269

But currently we dont use the `.Insecure` field for anything related to that specific case. As part of this RFC we could establish errors that should be observed on such situations.

rfcs/0004-insecure-http/README.md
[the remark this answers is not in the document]
pjbgf: I think this would still be useful in Kustomize Controller to enforce remote applies. In this case, the controller will error if `insecure-skip-tls-verify` inside the `.spec.kubeconfig` is not aligned with the controller level flag.

pjbgf:
As food for thought, what if instead of the new controller level flag being named `--allow-insecure-http` we reused the terminology from the CLI: `--insecure-skip-tls-verify`?

rfcs/0004-insecure-http/README.md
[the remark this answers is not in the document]
pjbgf: They are orthogonal. My point was not related to remote bases, it was related to **remote applies**. The flag is a controller level way to enforce that all outward communications are using TLS, which in this case would be considered when applying reconciliations against remote clusters.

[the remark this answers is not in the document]
stefanprodan: Both kustomize and helm controllers have the flag `--insecure-kubeconfig-tls` witch by default is set to `false` to prevent insecure connections to remote clusters. Are you proposing to drop this flag and replace it with a new one which would allow insecure connections by default?

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```sh
+ flux create source oci podinfo \
+    --url=oci://ghcr.io/stefanprodan/manifests/podinfo \
+    --tag=6.1.6 \
+    --interval=10m
+```
+
+Since these commands essentially create object definitions, the CLI should offer a boolean flag `--insecure`
+for relevant objects, which will be used for specifying the value of `.spec.insecure` of such objects.
```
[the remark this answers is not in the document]
aryan9600: Okay, let me add a note highlighting the difference

aryan9600:
@pjbgf from what I understand, the `--insecure-skip-tls-verify` is used to skip checking the validity of the server's cert, but you can still access an endpoint using an HTTPS connection. I fail to see how that aligns with `--allow-insecure-http` which dictates whether a controller can access an endpoint using an HTTP connection.

stealthybox:
"skip-tls-verify=true" would only suggest relaxing TLS behavior.

It doesn't mention and should not relate to blocking/allowing plain HTTP.

"force-https=false" (with a default to true) might make more sense both disabling TLS verification and permitting HTTP since it at least mentions the protocol.
I don't like this option because plain HTTP and unverified HTTPS are still inherently different. Unverified HTTP is still encrypted. Plain HTTP is not which makes a huge difference securing against third-parties on the wire.

As an SRE/ops person, I would just prefer 2 (or more) flags to control TLS verification and HTTP separately:
```
--insecure-skip-tls-verify
--insecure-allow-http
```

I would put the word "insecure" at the front, so the flags get grouped with other sensitive options.

If you need more flags to differentiate Kustomize remote base fetches from k8s API connections, that seems fine, but maybe not necessary.

I do think this feature is incomplete until it addresses how to get Kustomize to filter out HTTP.
Disabling remote bases completely seems like a temporary workaround if we want this level of configurability at the protocol layer.

Hopefully the Kustomize folks would be open to working this in?

stefanprodan:
Remote bases bypass source-controller so they break many of Flux promises: as we don’t verify nor do we store the remote sources in-cluster as Flux artifacts. Pulling remote bases at every apply makes Flux unpredictable, for example: if external traffic is jammed, Flux can’t detect & correct drift as the last desired state is unknown. Tracing an object to its source is no longer accurate, diffing and any static analysis involves outbound traffic. And finally, there is no API to setup SSH or TLS certs for remote bases, so for private repos you’ll need to mount the private keys and certs in the kustomize-controller pod (breaks multi-tenancy). 

I think Flux should encourage users into using sources which can be verified and trusted. Ideally all the external artifacts which make the desired state should by signed and transported securely, so no remote bases and no insecure protocols. Maybe with OCI we can make this story more appealing, `flux build artifact` could vendor the remote bases and `flux push artifact` could sign them.

aryan9600:
I think whether we should advice Flux users to avoid using remote bases is outside the crux of this proposal. The `allow-insecure-http` flag is meant to disable any usage of HTTP connections despite of whether the feature itself is unsafe. But since we can't feasibly filter out HTTP connections to remote bases, I think the _safe_ thing to do is not add this flag to kustomize-controller. The proposal can also be modified to have a note on why this flag has been excluded from kustomize-controller to clear any doubts over its absence. If at a later time, upstream Kustomize adds support for disabling access to remote bases served over HTTP, we can add this flag to kustomize-controller then.

rfcs/0004-insecure-http/README.md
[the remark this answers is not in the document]
aryan9600: `--inscecure-kubeconfig-tls` would skip TLS verification, but the protocol used would still HTTPS. I don't think adding `--allow-insecure-http` makes sense for remote applies as kube-apiserver by default is served over HTTPS and while technically it can be served over plain HTTP, the option for that configuration was disabled quite a while back: https://github.com/kubernetes/kubernetes/pull/65830

rfcs/0004-insecure-http/README.md
aryan9600: The Azure SDK referenced by you here is deprecated: https://github.com/Azure/azure-sdk-for-go/tree/9b91fa8fd891a3364779d8d32f2a13dfa98dc6f4/storage
We currently use the recommended sdk: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob, which does not support HTTP connections to my knowledege.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,200 @@
...
+> Note: The flag shall not be added to the following controllers:
+> * kustomize-controller: This flag is excluded from this controller, as the upstream `kubenetes-sigs/kustomize` project
+> does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
+> upstream project adds support for this at a later point in time.
+> * helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
+> Furthermore although both controllers also have the ability to do remote applies, serving `kube-apiserver` over plain
+> HTTP is disabled by default. While technically this can be enabled, the option for this configuration was also disabled
+> quite a while back (ref: https://github.com/kubernetes/kubernetes/pull/65830/).
+
+The default value of this flag shall be `false`. This would ensure that there is no breaking change with controllers
```
stefanprodan, line 49:
```suggestion
The default value of this flag shall be `true`. This would ensure that there is no breaking change with controllers
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+    --interval=10m
+```
+
+Since these commands essentially create object definitions, the CLI should offer a boolean flag `--insecure`
+for the required commands, which will be used for specifying the value of `.spec.insecure` of such objects.
+
+> Note: This flag should not be confused with `--insecure-skip-tls-verify` which is meant to skip TLS verification
+> when using an HTTPS connection.
+
+### Precedence & Validitiy
```
stefanprodan, line 103:
```suggestion
### Precedence & Validity
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+## Design Details 
+If a controller is started with `--insecure-allow-http=false`, any URL in a Flux object which has `http`
+as the scheme will result in an error and the following condition will be added to the object's 
+`.status.conditions`:
+
+```yaml
+status:
+  conditions:
+  - lastTransitionTime: "2022-09-06T09:14:21Z"
+    message: "Use of inecure HTTP connection isn't allowed for this controller"
```
stefanprodan, line 175:
```suggestion
    message: "Use of insecure HTTP connection isn't allowed for this controller"
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+Similarly, if an object has `.spec.insecure` as `true` but the Cloud provider doesn't allow HTTP connections,
+the reconciler will error out and add the condition below to the object's `.status.conditions`:
+
+```yaml
+status:
+  conditions:
+  - lastTransitionTime: "2022-09-06T09:14:21Z"
+    message: "Use of inecure HTTP connection isn't allowed for Azure Storage"
```
stefanprodan, line 189:
```suggestion
    message: "Use of insecure HTTP connection isn't allowed for Azure Storage"
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+    observedGeneration: 1
+    reason: InsecureConnectionsDisallowedReason
```
stefanprodan, line 191:
```suggestion
    reason: InsecureConnectionsDisallowed
```

stefanprodan (commented):
@aryan9600 please setup spellcheck in your editor

stefanprodan (approved):
LGTM

pjbgf:
I agree with @stealthybox on the flag name, `insecure` should be used as a prefix like we do for the kubeconfig flags:
```go
	flagInsecureKubeConfigExec = "insecure-kubeconfig-exec"
	flagInsecureKubeConfigTLS  = "insecure-kubeconfig-tls"
```
https://github.com/fluxcd/pkg/blob/87d4c63c2160c70f8557ad4dd7f2dacdf4e613dc/runtime/client/kubeconfig.go#L27-L28

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+* Provide a way for users to be made aware that their use of non-TLS endpoints is not supported if that is the case.
+
+### Non-Goals
+* Break Flux's current behavior of allowing HTTP connections.
+
+## Proposal
+
+### Controllers
+Flux users should be able to enforce that controllers are using HTTPS connections only.
+This shall be enabled by adding a new boolean flag `--allow-insecure-http` to the following controllers:
```
pjbgf, line 34: As per comment above:
```suggestion
This shall be enabled by adding a new boolean flag `--insecure-allow-http` to the following controllers:
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+At the object level, a field should be provided which would enable the use of non-TLS endpoints.
+
+If the use of a non-TLS endpoint is not supported, it should be made clear to users through the use of
+logs and status conditions.
+
+## Motivation
+Today the use of non-TLS based connections is inconsistent across Flux controllers.
+
+Controllers that deal only with `http` and `https` schemes have no way to block use of the `http` scheme at controller-level.
+Some Flux objects provide a `.spec.insecure` field to enable the use of non-TLS based endpoints, but they don't clearly notify users when the option is not supported (e.g. Azure/GCP Buckets).
```
pjbgf, line 20: nit:
```suggestion
Some Flux objects provide a `.spec.insecure` field to enable the use of non-TLS based endpoints, but they don't clearly notify
users when the option is not supported (e.g. Azure/GCP Buckets).
```

Is it worth clarifying that this is used mostly when the scheme is something beyond http/https?

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Goals
+* Provide a flag across relevant Flux controllers which disables all outgoing HTTP connections.
+* Add a field which enables the use of non-TLS endpoints to appropriate Flux objects.
+* Provide a way for users to be made aware that their use of non-TLS endpoints is not supported if that is the case.
+
+### Non-Goals
+* Break Flux's current behavior of allowing HTTP connections.
+
```
pjbgf, line 29: Maybe something to the effect that we are not dealing with inter-controller communications?
```suggestion
* Break Flux's current behavior of allowing HTTP connections.
* Change in behavior for Flux cross-control communications.
```

It may be worth cross referring to the discussions being held around that subject which is orthogonal to this RFC's scope, but they have a really high synergy and I would assume that interested parties would probably need to enforce/relax both.
https://github.com/fluxcd/flux2/discussions/3066

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,198 @@
...
+* image-reflector-controller
+
+> Note: The flag shall not be added to the following controllers:
+> * kustomize-controller: This flag is excluded from this controller, as the upstream `kubenetes-sigs/kustomize` project
+> does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
+> upstream project adds support for this at a later point in time.
+> * helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
+> Furthermore although both controllers can also do remote applies, serving `kube-apiserver` over plain
+> HTTP is disabled by default. While technically this can be enabled, the option for this configuration was also disabled
+> quite a while back (ref: https://github.com/kubernetes/kubernetes/pull/65830/).
```
pjbgf, line 47: We should mention that the enforcement for kube-apiserver communication is [currently handled](https://github.com/fluxcd/pkg/blob/87d4c63c2160c70f8557ad4dd7f2dacdf4e613dc/runtime/client/kubeconfig.go#L28) via `--insecure-kubeconfig-tls`. That setting changes the [TLSClientConfig.Insecure](https://github.com/kubernetes/client-go/blob/2f43d370b32c86d6442be6d0cefd9bd4e307db22/rest/config.go#L217) value, this is used to simply skip TLS verification.

pjbgf: This RFC could also define `--insecure-skip-tls-verification` and as part of the required changes we would update the existing `--insecure-kubeconfig-tls`. With that, we would follow the pattern across the board (and add both helm/kustomize controllers) here.

rfcs/0004-insecure-http/README.md
aryan9600: its not just it about the scheme being something other than http/https, sometimes its invalid to specify a scheme itself,  for eg: an endpoint to a container registry

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,204 @@
...
+* OCI `HelmRepository`: When using an OCI registry as a Helm repository, the `.spec.url` is expected to begin with `oci://`.
+Since the scheme part of the URL is used to specify the type of `HelmRepository`, there is no way for users to specify
+that the registry is hosted at a non-TLS endpoint.
+
+For such objects, we shall introduce a new boolean field `.spec.insecure`, which shall be `false` by default. Users that
+need their object to point to an HTTP endpoint, can set this to `true`.
+
+### CLI
+The Flux CLI offers several commands for creating Flux specific resources. Some of these commands may involve specifying
+an endpoint such as creating an `OCIRepository`:
```
stefanprodan, line 94: This has been implemented in https://github.com/fluxcd/flux2/releases/tag/v0.34.0 Please add it to the "Implementation History" section.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,209 @@
...
+  - gotk-sync.yaml
+patches:
+  - patch: |
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --insecure-allow-http=false
+    target:
+      kind: Deployment
+      name: "(source-controller|notification-controller|image-reflector-controller|image-automation-controller)"
+  # Since this above flag is not available in kustomize-controller for reasons explained in a previous section,
```
darkowlzz, line 137:
```suggestion
  # Since the above flag is not available in kustomize-controller for reasons explained in a previous section,
```

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,209 @@
...
+> does not support disabling HTTP connections while fetching resources from remote bases. We can revisit this if the
+> upstream project adds support for this at a later point in time.
+> * helm-controller: This flag does not serve a purpose in this controller, as the controller does not make any HTTP calls.
+> Furthermore although both controllers can also do remote applies, serving `kube-apiserver` over plain
+> HTTP is disabled by default. While technically this can be enabled, the option for this configuration was also disabled
+> quite a while back (ref: https://github.com/kubernetes/kubernetes/pull/65830/).
+
+Both kustomize-controller and helm-controller currently have a flag `--insecure-kubeconfig-tls` which makes the controller skip
+TLS verification when connecting to a Kubernetes cluster with an HTTPS connection. This flag shall be renamed to
+`--insecure-skip-tls-verify` to align it with the Flux CLI which offers this command for the same purpose.
```
darkowlzz, line 74: Does this belong in this RFC? Skip TLS verification in block insecure HTTP connection RFC. Seems unrelated and may deserve its own RFC or dealt separately.

pjbgf: The main reason why I suggested we covered this here was to take into account cases in which the scheme does not highlight the use of TLS (e.g. `oci://`).  In such cases, the use of the field `spec.insecure=True` would be blocked off when `--insecure-skip-tls-verify` is set to `false`.

The `--insecure-allow-http` would only apply and be relevant when the schemes `http://` and `https://` as being used.

stefanprodan: > The --insecure-allow-http would only apply and be relevant when the schemes http:// and https:// as being used.

This is so confusing, are you suggesting that to block all insecure connections, an admin will have to set 2 flags in source-controller: `--insecure-allow-http` and `--insecure-skip-tls-verify` ?

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,209 @@
+# RFC-0004 Block insecure HTTP connections across Flux
+
+**Status:** provisional
+
+**Creation Date:** 2022-09-08
+
+## Summary
+Flux should have a consistent way of disabling insecure HTTP connections.
```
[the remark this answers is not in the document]
hiddeco: It also makes "scan reading" the document easier :-)

rfcs/0004-insecure-http/README.md
hiddeco: They control separate things: `--insecure-allow-http` allows connections to HTTP endpoints, where as `--insecure-skip-tls-verify` skips the verification of TLS certificates over HTTPS (but does not allow HTTP).

If `--insecure-allow-http` is thus set, HTTP and HTTPS connections can be made but a HTTPS connection would still require a valid TLS certificate. While if `--insecure-skip-tls-verify` is set but `--insecure-allow-http` is not, HTTPS connections are required but the TLS verification is skipped.

stefanprodan: @hiddeco but that's not what @pjbgf is saying:

> the use of the field spec.insecure=True would be blocked off when --insecure-skip-tls-verify is set to false.

The use of `OCIRepository.spec.insecure` is for when the OCI repo is hosted on a plain HTTP non-TLS registry.

hiddeco: That seems wrong (probably due to a misunderstanding of what the field enables), I think the behavior as I described is correct. Skipping TLS verify for at least Helm OCI does not even seem to be an option at present.

pjbgf: My suggestion aligns with what @hiddeco mentioned. `--insecure-allow-http` is for the use of the `http://` scheme from a user-perspective. 

For OCI and Buckets sources, TLS assurances are not scheme bound - at least not from a user PoV. Therefore, the only ability they have to enforce such controls is via `spec.Insecure`. Blocking such use would be via `--insecure-skip-tls-verify`.

stefanprodan: @pjbgf IMO to block OCIRepositories when `spec.insecure=True` admins will set `--insecure-allow-http`, to block OCIRepositories when `spec.certSecretRef`  admins will set `--insecure-skip-tls-verify`.

aryan9600: I fail to see the relationship between `--insecure-skip-tls-verify` and `spec.insecure`. `spec.insecure` is solely for configuring whether the connection uses HTTP or HTTPS. `--insecure-skip-tls-verify` is only considered when the connection being used is an HTTPS one in the first place.
If `spec.insecure=true`,  then the value of `--insecure-skip-tls-verify` is inconsequential automatically because the connection is going to be made using the HTTP protocol, hence there's no TLS involved.
To further clarify any future questions, unlike configuring connections to be HTTP or HTTPS, there is no way to skip TLS verification at an object level, its only possible via the flag `--insecure-skip-tls-verify`.

pjbgf: I was under the impression that `spec.insecure` in Buckets skipped TLS verification. After going through the minio code I realised that that wasn't the case, but it rather just adjust TLSConfig settings, therefore I concur with:
- Removal of `--insecure-skip-tls-verify` from this RFC.
- Use of `--insecure-allow-http` to block `spec.Insecure=True` and the use of `http://`.

rashedkvm (approved):
LGTM. Nicely done @pjbgf

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+At the controller level, a flag should be present which would disable all outgoing HTTP connections.
+At the object level, a field should be provided which would enable the use of non-TLS endpoints.
+
+If the use of a non-TLS endpoint is not supported, it should be made clear to users through the use of
+logs and status conditions.
```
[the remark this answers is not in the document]
aryan9600: What I was trying to drive home through this sentence was that if for whatever reason a non-TLS endpoint cannot be used, users should be made aware of it so they can try to remedy the situation. How the controller behaves in this scenario is an implementation detail, thus irrelevant in this section.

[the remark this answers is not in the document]
aryan9600: I agree with you about the stalled condition, I'll add something accordingly in the design detail section below.

stefanprodan (approved):
LGTM

pjbgf (approved):
LGTM!

Great job putting this together @aryan9600! :bow:

hiddeco (approved):

stefanprodan:
I'm assigning RFC-0004 for this one, @aryan9600 please add `Last update` to today, and mark the PR ready.

stefanprodan:
One more changed need, please set the status to `implementable`

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,219 @@
...
+Similarly, if an object has `.spec.insecure` as `true` but the Cloud provider doesn't allow HTTP connections,
+the reconciler will error out and add the condition below to the object's `.status.conditions`:
+
+```yaml
+status:
+  conditions:
+  - lastTransitionTime: "2022-09-06T09:14:21Z"
+    message: "Use of insecure HTTP connections isn't allowed for Azure Storage"
+    observedGeneration: 1
+    reason: InsecureConnectionsDisallowed
```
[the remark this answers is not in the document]
darkowlzz: But this may just be an example and not exactly the implementation detail that'll be implemented.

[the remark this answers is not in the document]
darkowlzz: Since it's in design details section, maybe we need to clearly define these reasons and what they mean similar to the flags and fields introduced here.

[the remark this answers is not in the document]
darkowlzz: Looked for any existing unsupported reasons used in source-controller and found `UnsupportedSourceKind` in HelmChart.
Maybe we also need to consider a pattern for naming unsupported reason values.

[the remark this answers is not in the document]
aryan9600: what about `UnsupportedConnectionProtocol` or `UnsupportedConnectionType`?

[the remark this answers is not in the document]
darkowlzz: Since the protocol may be invisible, like in the case of bucket, `UnsupportedConnectionType` seems better. User sets `insecure: true`, unsupported connection type would mean that the insecure connection isn't supported.
But I'd like other's opinion about it too.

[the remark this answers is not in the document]
pjbgf: I would concur with `UnsupportedConnectionType`.

rfcs/0004-insecure-http/README.md
```diff
@@ -0,0 +1,219 @@
...
+  - lastTransitionTime: "2022-09-06T09:14:21Z"
+    message: "Use of insecure HTTP connections isn't allowed for this controller"
+    observedGeneration: 1
+    reason: InsecureConnectionsDisallowed
+    status: "True"
+    type: Stalled
+```
+
+Similarly, if an object has `.spec.insecure` as `true` but the Cloud provider doesn't allow HTTP connections,
+the reconciler will error out and add the condition below to the object's `.status.conditions`:
```
darkowlzz, line 199: > the reconciler will error out

This may not be accurate. Stalled conditions don't result in runtime error. The reconciliation just stops until the user makes a change in the configuration.
Maybe update this and the above paragraph accordingly. Maybe "the object will become not ready" or something more appropriate would be better.

aryan9600: i don't understand fully. how can any state that represents an "unsuccessful" reconciliation not result in an error?

hiddeco: Returning an error within a reconciler (as in: letting it bubble up within the `Reconcile` method) would requeue the object. Given this, setting `Stalled` swallows the error and rather emits it in the Condition type (and in e.g. the logs or an Event).

makkes (commented):
I love how this RFC is aligning controller flags, CLI flags and CRD fields. Thanks!

One concern I have is the rather generic term `insecure` in both, the CRDs and the CLI as it's just not specific enough in my book. Calling it `insecure-allow-http` would be much better to understand from a user's perspective. There's already been lots of confusion [in the Helm project around what "insecure" really means](https://github.com/helm/helm/issues/11352).

darkowlzz (approved):
LGTM!

pjbgf:
@makkes I concur that `--insecure` is too generic, and I would also support the CLI to have a more explicit and meaningful flag name instead. This would decrease confusion if we were to introduce other `--insecure-*` flags in the future, which I believe is quite likely.

As for the CRD, although I agree that the same argument applies, I am not sure the trade-off of breaking backwards compatibility vs adding a more meaningful name is worth it on that case. But if that were to happen, we probably better do it ahead of GA. What are your thoughts here?

Do you think there would be other variants within our CRDs that contain a `spec.Insecure` field?

Issue #3086, opened 2022-09-08
Monitoring - Flux Control Plane Dashboard shows incorrect memory consumption


pjbgf:
During a recent investigation we have noticed that the controller's memory consumption was being incorrectly displayed in our official Dashboard, when compared with the data from [Pod Stats & Info](https://grafana.com/grafana/dashboards/10518-pod-stats-info/) or Pixie:

#### Flux Control Plane > Memory
![image](https://user-images.githubusercontent.com/5452977/189129897-9612ec24-49ee-4e2a-975b-0a6e06fdc532.png)
#### Pod Stats & Info
![image](https://user-images.githubusercontent.com/5452977/189130467-b1301060-37e9-4401-8031-1694e0a7519a.png)
#### Pixie PX/Pod
![image](https://user-images.githubusercontent.com/5452977/189130544-af4d9914-f013-4937-b977-bc566e69559f.png)

We need to review that dashboard and update it accordingly.

stefanprodan:
@pjbgf I would argue that we show the correct stats by looking at Go stats (go_memstats_alloc_bytes) instead of the OS.

pjbgf:
Unfortunately that does not account for CGO or C allocated objects (i.e. libgit2) that were not freed. Changing the current panel (left) to only contain source controller, and adding the same metrics as Pod Stats & Info's I get this for the same time range:
![image](https://user-images.githubusercontent.com/5452977/189139297-ae318d1e-928a-4b86-a39d-a4979d913fe0.png)

If we prefer go metrics, we could use `go_memstats_heap_sys_bytes` instead, which seem more accurate for our use case:
![image](https://user-images.githubusercontent.com/5452977/189142647-73b9f9d9-db69-4c75-a0d8-a5cfb5c2e46b.png)

Santosh1176:
Can I take this up? 
Just to reconfirm, the Metrics in `rate(go_memstats_alloc_bytes_total{namespace=\"$namespace\",pod=~\".*-controller-.*\"}[1m])` should be replaced with the one mentioned by @stefanprodan to resolve this?

stefanprodan:
@Santosh1176 Yes, please note that the `go_memstats_alloc_bytes_total` is a counter while the `go_memstats_heap_sys_bytes` is a gauge.

Santosh1176:
> the go_memstats_heap_sys_bytes is a gauge.


Thankyou @stefanprodan 
Does that mean, rate() shouldn’t be used here?

PR #3098, opened 2022-09-12
[Grafana] Use `container_memory_working_set_bytes` to report memory consumption

Santosh1176:
This PR could close [#3086](https://github.com/fluxcd/flux2/issues/3086#issue-1366306801)

Updating current metric of type `go_memstats_alloc_bytes_total` with `container_memory_working_set_bytes` to correctly report memory consumption.

### DIFF
--- a/manifests/monitoring/monitoring-config/dashboards/control-plane.json
+++ b/manifests/monitoring/monitoring-config/dashboards/control-plane.json
@@ -548,7 +548,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "rate(go_memstats_alloc_bytes_total{namespace=\"$namespace\",pod=~\".*-controller-.*\"}[1m])",
+          "expr": "sum(container_memory_working_set_bytes{namespace=\"$namespace\",container!=\"POD\",container!=\"\",pod=~\".*-controller-.*\"}) by (pod)",
           "hide": false,
           "interval": "",
           "legendFormat": "{{pod}}",


stefanprodan:
@Santosh1176 can you post a print screen with the two graphs, how it looked before and after? Thanks

Santosh1176:
@stefanprodan Apologies for not testing it before raising a PR. I shall keep that in mind for the future.
I referred the [Prometheus docs](https://prometheus.io/docs/practices/instrumentation/#counter-vs-gauge-summary-vs-histogram), which recomended not using `rate` on Guage metrics. But, this wasn't working when I tried it locally:

error-monitor1

When I applied `rate` to the metric. Its showing as below:

with-rate-monitor

stefanprodan:
> this wasn't working when I tried it locally

It doesn’t work because the query is wrong, remove the `()[1m]`

Santosh1176:
Thank you @stefanprodan 
The updated graph:

time

stefanprodan:
Looks good now! Thank you, can you please squash the commits into a single one, rebase with upstream and force push.

aryan9600:
@Santosh1176 i have tested your changes and while using `go_memstats_heap_sys_bytes` does make the metrics much more accurate, i don't think its the right one to go for. `container_memory_working_set_bytes` is the query that we should use as its the standard query used in other dashboards such as [Pod Stats & Info](https://grafana.com/grafana/dashboards/10518-pod-stats-info/) and its also what the scheduler tracks before it OOMKills the container (ref: https://faun.pub/how-much-is-too-much-the-linux-oomkiller-and-used-memory-d32186f29c9d).

Santosh1176:
> `container_memory_working_set_bytes` is the query that we should use

@aryan9600 @stefanprodan Thanks for this valuable information, so for this has been a good learning experience.
I tried applying the said metric to the dashboard, results are below:

container

I had a doubt, as in other metric types I could not select pods to show their memory usage stats. Am I doing something wrong here? Or its the behavior specific to this metric?

aryan9600:
> I had a doubt, as in other metric types I could not select pods to show their memory usage stats. Am I doing something wrong here?

Since we are using `sum()` and the filter passed to the query selects all pods in the namespace with "controller" in its name, you're seeing the memory being consumed by all these pods. You need to use a group by like:
```
sum(container_memory_working_set_bytes{namespace="$namespace",container!="POD",container!="",pod=~".*-controller-.*"}) by (pod)
```

stefanprodan:
@Santosh1176 please rebase instead of merging, The [0540722](https://github.com/fluxcd/flux2/pull/3098/commits/0540722ba5954f6d50921b506336c85e566fc7ea) commit has nothing to do with this PR.

Santosh1176:
> 0540722 commit has nothing to do with this PR.

Sorry! my bad, I accidentally added the unrelated commit while squashing with rebase, apologies. Can I now use the `drop` option while rebasing, or that would further complicate the mess.
Thanks

stefanprodan:
Please squash the commits into a single one and it's ready to go. Thanks!

Santosh1176:
> Please squash the commits into a single one and it's ready to go. Thanks!

Done. Request review @stefanprodan @aryan9600

aryan9600:
@Santosh1176 you need to force push the squashed commit as well

stefanprodan (approved):
LGTM

Thanks @Santosh1176 🏅

Santosh1176:
Thank you @stefanprodan @aryan9600

PR #3158, opened 2022-09-29
[RFC] Gating Flux reconciliation

stefanprodan:
Add proposal for introducing a gating mechanism to Flux that would offer a way for cluster admins and other teams involved in the release process to manually approve the rollout of changes onto clusters.

In addition, Flux could offer a way to define maintenance time windows and other time-based gates, to allow a better control of applications and infrastructure changes to critical system.

**Note** that this proposal is WIP, we need to cover more use-cases from https://github.com/fluxcd/flux2/discussions/870

### DIFF
--- a/rfcs/XXXX-gating/README.md
+++ b/rfcs/XXXX-gating/README.md
@@ -4,7 +4,7 @@
 
 **Creation date:** 2022-09-28
 
-**Last update:** 2022-09-28
+**Last update:** 2022-10-04
 
 ## Summary
 
@@ -61,6 +61,8 @@ The `Gate` API would replace Flagger's current
 [manual gating mechanism](https://docs.flagger.app/usage/webhooks#manual-gating).
 
 ### User Stories
+
+#### Story 1
 
 > As a member of the SRE team, I want to allow deployments to happen only
 > in a particular time frame of my own choosing.
@@ -220,13 +222,60 @@ status:
 
 You could also schedule "No Deploy Fridays" with a CronJob that closes the `maintenance` gate at `0 0 * * FRI`.
 
+#### Story 2
+
+> As a member of the SRE team, I want existing deployments to still be
+> reconciled during a change freeze.
+
+Gates can be used to block Flux sources from being refreshed, resulting in Flux
+to continue to reconcile existing approved desired states, whislt new changes
+are held at a Flux source gate.
+
+Example:
+
+```yaml
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: GitRepository
+metadata:
+  name: flux-system
+  namespace: flux-system
+spec:
+  gates:
+    - name: change-freeze # gate that enforces a change freeze time window
+status:
+  conditions:
+    - lastTransitionTime: "2022-05-26T01:12:22Z"
+      message: "Reconciliation is blocked as gate 'flux-system/change-freeze' is closed."
+      reason: GateClosed
+      status: "True"
+      type: Blocked
+```
+
+This would ensure that Gate changes would not impact the eventual consistency of
+mid-flight reconciliations that were already deployed in the cluster. Flux would also
+continue to re-create Flux managed objects that were manually deleted from the cluster.
+
 ### Alternatives
 
-<!--
-List plausible alternatives to the proposal and explain why the proposal is superior.
+#### Users to implement gating outside of Flux
 
-This is a good place to incorporate suggestions made during discussion of the RFC.
--->
+##### Before Flux source
+
+Users could implement their own gating mechanisms as part of their development processes
+ensuring that their custom rules are applied before the changes reach their Flux sources
+(i.e. the target Git repository). For example, if deployments are not allowed on Fridays,
+no PRs would be merged on those days.
+
+The disadvantage is that some source types may not provide easy ways for users to enforce
+such rules. When using different source types (e.g. Git, OCI, Helm), multiple implementations
+may be required.
+
+##### CronJobs and Flux Suspend
+
+Users can implement a gating mechanism within Kubernetes by leveraging CronJobs and using
+the built-in suspend feature in Flux that allows for a Flux object to stop being reconciled
+until it is resumed. This alternative does not scale well when considering hundreds of Flux
+objects.
 
 ## Design Details
 
--- a/rfcs/XXXX-gating/README.md
+++ b/rfcs/XXXX-gating/README.md
@@ -1,0 +1,252 @@
+# RFC-XXXX Gating Flux reconciliation
+
+**Status:** provisional
+
+**Creation date:** 2022-09-28
+
+**Last update:** 2022-09-28
+
+## Summary
+
+Flux should offer a mechanism for cluster admins and other teams involved in the release process
+to manually approve the rollout of changes onto clusters. In addition, Flux should offer 
+a way to define maintenance time windows and other time-based gates, to allow a better control 
+of applications and infrastructure changes to critical system.
+
+## Motivation
+
+Flux watches sources (e.g. GitRepositories, OCIRepositories, HelmRepositories, S3-compatible Buckets) and
+automatically reconciles the changes onto clusters as described with Flux Kustomizations and HelmReleases.
+The teams involved in the delivery process (e.g. dev, qa, sre) can decide when changes are delivered
+to production by reviewing and approving the proposed changes in a collaborative manner with pull request.
+Once a pull request is merged onto a branch that defines the desired state of the production system,
+Flux kicks off the reconciliation process.
+
+There are situations when users want to have a gating mechanism after the desired state changes are merged in Git:
+
+- Manual approval of container image updates (e.g. https://github.com/fluxcd/flux2/discussions/870)
+- Manual approval of infrastructure upgrades (e.g. https://github.com/fluxcd/flux2/issues/959)
+- Maintenance window (e.g. https://github.com/fluxcd/flux2/discussions/1004)
+- Planned releases
+- No Deploy Friday
+
+### Goals
+
+- Offer a dedicated API for defining time-based gates in a declarative manner.
+- Introduce a `gating-controller` in the Flux suite that manages the `Gate` objects.
+- Extend the current Flux APIs and controllers to support gating.
+
+### Non-Goals
+
+<!--
+What is out of scope for this RFC? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Proposal
+
+In order to support manual gating, Flux could be extended with a dedicated API and controller
+that would allow users to define `Gate` objects and perform operations like `open` and `close`.
+
+A `Gate` object could be referenced in sources (Buckets, Git, Helm, OCI Repositories)
+and syncs (Kustomizations, HelmReleases, ImageUpdateAutomation)
+to block the reconciliation until the gate is opened.
+
+A `Gate` can be opened or closed by annotating the object with a timestamp or by
+calling a specific webhook receiver exposed by notification-controller.
+
+A `Gate` can be configured to automatically close or open based on a time window defined in the `Gate` spec.
+
+The `Gate` API would replace Flagger's current
+[manual gating mechanism](https://docs.flagger.app/usage/webhooks#manual-gating).
+
+### User Stories
+
+> As a member of the SRE team, I want to allow deployments to happen only
+> in a particular time frame of my own choosing.
+
+Define a gate that automatically closes after 1h from the time it has been opened:
+
+```yaml
+apiVersion: gating.toolkit.fluxcd.io/v1alpha1
+kind: Gate
+metadata:
+  name: sre-approval
+  namespace: flux-system
+spec:
+  interval: 30s
+  default: closed
+  window: 1h
+```
+
+When the gate is created in-cluster, the `gating-controller` uses `spec.default` to set the `Opened` condition:
+
+```yaml
+apiVersion: gating.toolkit.fluxcd.io/v1alpha1
+kind: Gate
+metadata:
+  name: sre-approval
+  namespace: flux-system
+status:
+  conditions:
+    - lastTransitionTime: "2021-03-26T10:09:26Z"
+      message: "Gate closed by default"
+      reason: ReconciliationSucceeded
+      status: "False"
+      type: Opened
+```
+
+While the gate is closed, all the objects that reference it will wait for an approval:
+
+```yaml
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: my-app
+  namespace: flux-system
+spec:
+  gates:
+    - name: sre-approval
+    - name: qa-approval
+status:
+  conditions:
+    - lastTransitionTime: "2021-03-26T10:09:26Z"
+      message: "Reconciliation is waiting approval, gate 'flux-system/sre-approval' is closed."
+      reason: GateClosed
+      status: "False"
+      type: Approved
+```
+
+The SRE team can open the gate either by annotating the gate or by calling the notification-controller webhook:
+
+```sh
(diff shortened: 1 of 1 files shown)


rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+      reason: ReconciliationSucceeded
+      status: "False"
+      type: Opened
+```
+
+While the gate is closed, all the objects that reference it will wait for an approval:
+
+```yaml
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
```
pjbgf, line 105: I propose gates are to be applied at sources only instead of `Kustomization` and `HelmRelease` objects, as otherwise may face the non-trivial challenges below:
- Incompatible desired state when `Kustomization` / `HelmRelease` objects are gated whilst others are not, specially when they depend on the same Flux source that is not gated.
  - Example 1: `Kustomization` objects `app2` and `app1` depends on `GitRepository` `team-source`. A change is merged into the repository and reconciled for object `team-source`, however `app2` has a close Gate and `app1` does not. In this case the desired state at both Git repository and the Flux source in the cluster are valid, but only one Kustomizations is being applied leading to an invalid observed state in the cluster. Same applies if both Kustomizations are behind a Gate which has its state changed before the second object is reconcilied (e.g. example 2 below).
- Enforcing the desired state which was allowed **before the Gate closing** (User Story 2) will be difficult to manage from any point beyond the source, which could lead to the cluster to be in an invalid state.
  - Example 1: user manually deletes Kubernetes object managed in the hope Flux will re-create it.
  - Example 2: two kustomizations (`parent` and `child`) having the same Gate, the latter has a `dependsOn` pointing to the former. If the Gate closes after `parent` but before `child` is applied, the cluster observed state may be invalid (considering the desired state at source).

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+change are understandable. This may include API specs and code snippets.
+
+The design details should address at least the following questions:
+- How can this feature be enabled / disabled?
+- Does enabling the feature change any default behavior?
+- Can the feature be disabled once it has been enabled?
+- How can an operator determine if the feature is in use?
+- Are there any drawbacks when enabling this feature?
+-->
+
```
pjbgf, line 293: A few ideas on design details to be discussed:

#### Multiple Gates

Flux objects that support gating can specify multiple gates. By default,
all gates specified must be open for the reconciliation to go ahead. To
change the behavior `spec.gates.require` can be set to `oneOf` instead:

```yaml
kind: GitRepository
metadata:
  name: flux-system
  namespace: flux-system
spec:
  gates:
    # all (default): all gates must be open for the reconciliation to go ahead.
    # oneOf: at least one of the gates must be open for the reconciliation to go ahead.
    require: oneOf # <all|oneOf>
    refs:
    - change-freeze # gate that enforces a change freeze time window
    - bypass-signoff # gate that allows other gates to be overriden.
```

When `oneOf` is used, a single open Gate is required for reconcilations
to proceed.

#### Recovering from wiped Storage 

The source artifact storage lives in the running source controller instance. If its Pod
is recreated, all the artifacts will need to be reacquired. This process happens
automatically as part of each source reconciliation, however, when sources
are restricted by a gate they will require a new process.

In such situation the controller will need to fetch the version used before the Gate was
closed, and that was in the storage before it being wiped. Not all types will support this
and therefore here are the outcomes expected:

- GitRepository: restores artifact based on the last known revision (commit ID).
- S3 Buckets: fail reconciliation with "artifact not found artifact whilst under closed Gate".
- OCI: TBC.
- Helm Repository: TBC.

stefanprodan: For OCI we have the upstream digest in status, so like with Git, we can use it to pull the data.

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+  conditions:
+    - lastTransitionTime: "2021-03-26T10:00:00Z"
+      message: "Gate scheduled for opening at 2021-03-27T11:00:00Z"
+      reason: ReconciliationSucceeded
+      status: "False"
+      type: Opened
+```
+
+You could also schedule "No Deploy Fridays" with a CronJob that closes the `maintenance` gate at `0 0 * * FRI`.
+
```
hassenius, line 224: In addition to the time dimension, I would also propose a reference dimension, such that a gate can be defined as open to a specific reference, either in spec, or as an annotation

For example

```yaml
apiVersion: gating.toolkit.fluxcd.io/v1alpha1
kind: Gate
metadata:
  name: approved-revisions
  namespace: flux-system
spec:
  openTo: 4f08fc93e31
```

or

```yaml
apiVersion: gating.toolkit.fluxcd.io/v1alpha1
kind: Gate
metadata:
  annotations:
    open.gate.fluxcd.io/revision: 4f08fc93e31
  name: approved-revisions
  namespace: flux-system
spec:
  default: closed
```

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+#### Story 2
+
+> As a member of the SRE team, I want existing deployments to still be
+> reconciled during a change freeze.
+
+Gates can be used to block Flux sources from being refreshed, resulting in Flux
```
hassenius, line 230: To clarify that `block Flux sources from being refreshed` in this case means the gate prevent the source controller from making the new assets available, rather than to suspend the polling in the source controller?

It is desirable for the source to still be able to detect changes, such that in-cluster logic that is tied to the gating mechanisms can be notified of available changes.

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+- Manual approval of container image updates (e.g. https://github.com/fluxcd/flux2/discussions/870)
+- Manual approval of infrastructure upgrades (e.g. https://github.com/fluxcd/flux2/issues/959)
+- Maintenance window (e.g. https://github.com/fluxcd/flux2/discussions/1004)
+- Planned releases
+- No Deploy Friday
+
+### Goals
+
+- Offer a dedicated API for defining time-based gates in a declarative manner.
```
hassenius, line 35: I would suggest that the gates should not be exclusively time based, but also support specific revisions

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+      message: "Reconciliation is blocked as gate 'flux-system/change-freeze' is closed."
+      reason: GateClosed
+      status: "True"
+      type: Blocked
+```
+
+This would ensure that Gate changes would not impact the eventual consistency of
+mid-flight reconciliations that were already deployed in the cluster. Flux would also
+continue to re-create Flux managed objects that were manually deleted from the cluster.
+
```
hassenius, line 257: I would propose an additional use case, bringing in the reference dimension in addition to the time dimension for gates, such that a gate can be defined as open to a specific reference, either in spec, or as an annotation

#### Story 3
> As a multi-cluster operator, I want to control how cluster configuration changes
> are rolled out across the cluster estate from a single configuration source

Gates can be used to control the timing and order of where configurations are reconciled

For example

```yaml
apiVersion: gating.toolkit.fluxcd.io/v1alpha1
kind: Gate
metadata:
  name: approved-revisions
  namespace: flux-system
spec:
  openTo: 4f08fc93e31
```

or

```yaml
apiVersion: gating.toolkit.fluxcd.io/v1alpha1
kind: Gate
metadata:
  annotations:
    open.gate.fluxcd.io/revision: 4f08fc93e31
  name: approved-revisions
  namespace: flux-system
spec:
  default: closed
```

hiddeco:
Have been thinking about the current proposal, and am wondering if the ecosystem would benefit from a simple static `Gate` object which is actually controlled by a `Gatekeeper` (which knows about the mechanics around e.g. a CronJob).

By doing this, it would be much easier for others to write their own "gatekeeper" which e.g. listens to button presses in a UI.

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+##### CronJobs and Flux Suspend
+
+Users can implement a gating mechanism within Kubernetes by leveraging CronJobs and using
+the built-in suspend feature in Flux that allows for a Flux object to stop being reconciled
+until it is resumed. This alternative does not scale well when considering hundreds of Flux
+objects.
+
+## Design Details
+
```
pjbgf, line 281: In highly regulated environments Gating can be used to ensure specific processes were observed before a new change made its way to a given environment. Flux controllers that support such mechanism should extend their log messages to express what Gates were taken into account at the time of a reconciliation, and their states.

rupelu:
Just throwing in another idea. It would be great if a Gate could evaluate a metric. This way we could decline reconciliation on high load or when some long running task is not yet completed. A metrics based Gate would make checks, whether a release can be rolled out, more automatic. Of course as already mentioned by @pjbgf it would be highly appreciated if we would know the reason why a reconciliation did not happen and which Gate was involved - maybe even what the evaluation of the said metric was.

rfcs/XXXX-gating/README.md
pjbgf, line 35:
```suggestion
- Offer a dedicated API for defining gates in a declarative manner.
```

rfcs/XXXX-gating/README.md
```diff
@@ -0,0 +1,301 @@
...
+to block the reconciliation until the gate is opened.
+
+A `Gate` can be opened or closed by annotating the object with a timestamp or by
+calling a specific webhook receiver exposed by notification-controller.
+
+A `Gate` can be configured to automatically close or open based on a time window defined in the `Gate` spec.
+
+The `Gate` API would replace Flagger's current
+[manual gating mechanism](https://docs.flagger.app/usage/webhooks#manual-gating).
+
```
pjbgf, line 62: Additional audit trail for this feature:
- Enriching existing logs so that actions that can be gated should log:
	- What gate(s) were checked (if any).
	- What gate(s) caused the final result of (not) going ahead with the action.
	- If multiple gates can be defined, changes to which gates are being observed and what logical operators (and/or) are being used, should be descriptively logged.
- Logs for the new Gate controller:
	- Descriptive summary of the gate state at creation time.
	- Any change of Gate state should log a descriptive summary.
- The additional logs above would be an "opt-in" feature, so that users that do not require this level of governance don't have to process the extra data.

rfcs/XXXX-gating/README.md
pjbgf: If it stopped polling, it would avoid all the resource consumption (CPU, memory, Network and storage) for an operation that would not cause any side effect to the cluster, which would stream line the controllers to better handle reconciliations that were not gated.

@hassenius is there any specific scenario you have in mind?

pjbgf:
> Have been thinking about the current proposal, and am wondering if the ecosystem would benefit from a simple static `Gate` object which is actually controlled by a `Gatekeeper` (which knows about the mechanics around e.g. a CronJob).
> 
> By doing this, it would be much easier for others to write their own "gatekeeper" which e.g. listens to button presses in a UI.

I agree with @hiddeco, this would make the implementation quite extensible. Users could implement whatever logic they need to toggle the state on top of the static Gate.

In-tree "GateKeeper" implementations could be added based on most popular requests amongst the community.

pjbgf:
> Just throwing in another idea. It would be great if a Gate could evaluate a metric. This way we could decline reconciliation on high load or when some long running task is not yet completed. A metrics based Gate would make checks, whether a release can be rolled out, more automatic. Of course as already mentioned by @pjbgf it would be highly appreciated if we would know the reason why a reconciliation did not happen and which Gate was involved - maybe even what the evaluation of the said metric was.

@rupelu The Gate happens before a new state is applied into the cluster. So I think they are a better fit for defining whether or not sources should be refreshed, for the reasons I mentioned [here](https://github.com/fluxcd/flux2/pull/3158#discussion_r987049960). However, you could have similar results by using [Flagger](https://fluxcd.io/flagger/) to do the progressive roll-out side of things for you.
If the release was broken, no (or few) users would be redirected to the new version. Rolling fixes back/forward, would still require the Gate to be opened though.

pjbgf:
I have worked out an example on what this could look like if we were to follow @hiddeco's suggestion on using static gates. This is what I got to: https://hackmd.io/N-xK-y5fTCiFFnusF22Sfw. I talked with a few users to ascertain whether this would mitigate the key use cases they have, and so far the response has been positive.

The example is not meant to define the implementation details, but rather to flesh out the workflow so folks can raise concerns sooner rather than later.

haggishunk:
*Counterpoint:*

I am fond of the shifting left that flux encourages with its watchful polling of sources like git repos and image registries.  Maybe it's a blocker to widespread community adoption to not have a way to put a human in the loop, but in some cases that human that pushes the deploy button is not the one that merged the app code.  Flux is differentiated from many other imperative pipeline CD tools by this very nature which sets the "human in the loop" to be those that commit artifacts to the watched sources of truth and the policies/controls around them.

Let's take the scenario of putting a gate up on a Friday afternoon, having the source evolve with new commits/pushes.  Some developers finish their projects and update the source of truth, say with refs to new images.  On Monday when the gate is lifted the accumulation of several code changes are released.  If things perform unexpectedly, what is the last good state?  The last commit before the gate went up, sure, but this is cross domain.  I would much rather roll back a merge train that hit Monday morning with reverting commits.  You will have better team engagement with the release if you remove the safety of a gate window.

Promoting the suspend feature to a more first class implementation seems to strike the right balance IMHO.

*Disclaimer:*

Please forgive my rant, but to me flux means "minutes of time difference between my gitops source of truth and my target environment" and I love it that way.

antonmatsiuk:
> _Counterpoint:_
> 
> I am fond of the shifting left that flux encourages with its watchful polling of sources like git repos and image registries. Maybe it's a blocker to widespread community adoption to not have a way to put a human in the loop, but in some cases that human that pushes the deploy button is not the one that merged the app code. Flux is differentiated from many other imperative pipeline CD tools by this very nature which sets the "human in the loop" to be those that commit artifacts to the watched sources of truth and the policies/controls around them.
> 
> Let's take the scenario of putting a gate up on a Friday afternoon, having the source evolve with new commits/pushes. Some developers finish their projects and update the source of truth, say with refs to new images. On Monday when the gate is lifted the accumulation of several code changes are released. If things perform unexpectedly, what is the last good state? The last commit before the gate went up, sure, but this is cross domain. I would much rather roll back a merge train that hit Monday morning with reverting commits. You will have better team engagement with the release if you remove the safety of a gate window.
> 
> Promoting the suspend feature to a more first class implementation seems to strike the right balance IMHO.
> 
> _Disclaimer:_
> 
> Please forgive my rant, but to me flux means "minutes of time difference between my gitops source of truth and my target environment" and I love it that way.

I think your argument is unrelated to what most of the people here ask for. The most valuable use case for the gating feature is to have "maintenance windows" on the environments that reconcile from a stable release branch, and not for having CD-style of deployment that you've described. Imagine you have dozens of environments that have to be updated across the globe within fixed timeframes. Do you want to wake up at night to press the merge button to update the environment? If not, then you have to play around with scheduling of "PR merge" events as CronJobs and creating different branches for different timezones in your version control platform. This will become messy and complex pretty fast.  Gating feature is an elegant solution for this problem, and if you need to update the environment outside the maintenance window, an "override" flag should be there.

stefanprodan:
Closing this in favour of [Flux Operator](https://github.com/controlplaneio-fluxcd/flux-operator) time-based delivery feature. 

Documentation:
-  Blog post: [Time-based deployments with Flux Operator](https://fluxcd.io/blog/2025/07/time-based-deployments/)
- User Guide: [Using ResourceSets for Time-Based Delivery](https://fluxcd.control-plane.io/operator/resourcesets/time-based-delivery/)
- API docs: [ResourceSets Schedule](https://fluxcd.control-plane.io/operator/resourcesetinputprovider/#schedule)

PR #3174, opened 2022-10-05
Update libgit2 version in Azure e2e tests

phillebaba:
Need to update libgit2 library version to keep in line with version used by SC.


phillebaba:
Going to run Azure e2e tests locally to make sure that this fixes the Azure e2e tests before merging.

stefanprodan (approved):
LGTM

Thanks @phillebaba 🏅

PR #3193, opened 2022-10-11
Process getVersion error in bootstrap

bkreitch:
Check and process the error from getVersion() during bootstrap (same as it's done during install in cmd/flux/install.go). Otherwise incorrect flux version (without "v" or not existing) may be silently passed and latest version is used instead for manifest generation.

Signed-off-by: Boris Kreitchman <bkreitch@gmail.com>


bkreitch:
@stefanprodan Hi, do you think this could be merged or anything additional is needed?

stefanprodan:
This looks Ok, sorry for the delay, can you please rebase and force push.

bkreitch:
> This looks Ok, sorry for the delay, can you please rebase and force push.

@stefanprodan Ok, rebased.

stefanprodan (approved):
LGTM

Thanks @bkreitch 🎖️

PR #3197, opened 2022-10-12
Move uninstall code to pkg

phillebaba:
This change moves the uninstall logic from `cmd` to `pkg`. This is done so that it can be shared with the Terraform Provider reducing the need to duplicate code.


pkg/uninstall/uninstall.go
```diff
@@ -0,0 +1,321 @@
+package uninstall
```
hiddeco, line 1: License header is also missing here :-)

pkg/uninstall/uninstall.go
```diff
@@ -0,0 +1,321 @@
...
+					logger.Failuref("CustomResourceDefinition/%s deletion failed: %s", r.Name, err.Error())
+				} else {
+					logger.Successf("CustomResourceDefinition/%s deleted %s", r.Name, dryRunStr)
+				}
+			}
+		}
+	}
+}
+
+func UninstallNamespace(ctx context.Context, logger log.Logger, kubeClient client.Client, namespace string, dryRun bool) {
```
hiddeco, line 291: Would `uninstall.Namespace` be better?

hiddeco:
Same suggestions apply for the other functions.

phillebaba:
@hiddeco I agree that removing the `Uninstall` prefix from each function seems reasonable as it is in the package name. Fixed your comments.

hiddeco:
One additional nit (other than that absolutely perfect): given things move into the public API, it would probably be good to document the (behavior) of the functionalities. Both because of consumer ease, and ensuring the behavior stays the same in terms of backwards compatibility.

hiddeco (approved):

PR #3198, opened 2022-10-12
Add nop logger

phillebaba:
This change adds a discard implementation of the logger interface which is useful for testing purposes.

### DIFF
--- a/pkg/log/nop.go
+++ b/pkg/log/nop.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -30,3 +30,17 @@ type Logger interface {
 	// Failuref logs a formatted failure message.
 	Failuref(format string, a ...interface{})
 }
+
+type DiscardLogger struct{}
```
stefanprodan, line 34: Please move this to a new file e.g. `nop.go` and rename the type to `NopLogger` like zap does.

phillebaba: Done

hiddeco: Wouldn't simply `log.Discard` make more sense?

stefanprodan (approved):

pkg/log/log.go
hiddeco: My above comment wasn't taken into account when this was merged.

PR #3199, opened 2022-10-12
Move bootstrap to pkg

phillebaba:
This change moves the bootstrap package from `internal` to `pkg` to enable sharing code with the Terraform provider.


stefanprodan:
Wasn't this done in https://github.com/fluxcd/flux2/pull/3237?

phillebaba:
Yes I redid it because I wanted to split the work into multiple PRs. Closing this as duplicate.

PR #3224, opened 2022-10-19
Add `diff artifact` command

### LINKED ISSUE #3194: Implement `flux diff artifact`
To avoid pushing artifacts with the same content to the registry, we could add a diff command that compares the locally build tarball checksum with the upstream tarball included in the first layer of the OCI artifact.

Example:

```sh
flux diff artifact oci://ghcr.io/stefanprodan/manifests:podinfo:6.2.0 --path=./kustomize
``` 

The above command will return exit code 0 if no diff was detected or exit code 1 if the contents differs.

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #3194 

/cc @stefanprodan

### DIFF
--- a/cmd/flux/diff_artifact.go
+++ b/cmd/flux/diff_artifact.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/fluxcd/flux2/internal/flags"
+	oci "github.com/fluxcd/pkg/oci/client"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/spf13/cobra"
+)
+
+var diffArtifactCmd = &cobra.Command{
+	Use:   "artifact",
+	Short: "Diff Artifact",
+	Long:  `The diff artifact command computes the diff between the remote OCI artifact and a local directory or file`,
+	Example: `# Check if local files differ from remote
+flux diff artifact oci://ghcr.io/stefanprodan/manifests:podinfo:6.2.0 --path=./kustomize`,
+	RunE: diffArtifactCmdRun,
+}
+
+type diffArtifactFlags struct {
+	path        string
+	creds       string
+	provider    flags.SourceOCIProvider
+	ignorePaths []string
+}
+
+var diffArtifactArgs = newDiffArtifactArgs()
+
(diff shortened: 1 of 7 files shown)


developer-guy:
don't worry about the "replace", I'll remove that before merging this

cmd/flux/diff_artifact.go
```diff
@@ -0,0 +1,111 @@
+/*
+Copyright 2021 The Flux authors
```
stefanprodan, line 2:
```suggestion
Copyright 2022 The Flux authors
```

cmd/flux/diff_artifact.go
```diff
@@ -0,0 +1,111 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/fluxcd/flux2/internal/flags"
```
stefanprodan, line 22: Split the imports in 3 groups please: std lib, 3rd party and fluxcd

cmd/flux/diff_artifact.go
```diff
@@ -0,0 +1,111 @@
...
+			return fmt.Errorf("provider not supported: %w", err)
+		}
+
+		if err := ociClient.LoginWithProvider(ctx, url, ociProvider); err != nil {
+			return fmt.Errorf("error during login with provider: %w", err)
+		}
+	}
+
+	if err := ociClient.Diff(ctx, url, diffArtifactArgs.path, diffArtifactArgs.ignorePaths); err != nil {
+		return fmt.Errorf("error during diff: %w", err)
```
stefanprodan, line 106:
```suggestion
		return err
```

stefanprodan (commented):
Please add tests and golden files to make sure we don't break the behaviour of this command in the future.

stefanprodan (commented):
Please remove the pkg replace and set `oci/v0.14.0` in go.mod, then run `make tidy` and push.

cmd/flux/diff_artifact.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/fluxcd/flux2/internal/flags"
+	oci "github.com/fluxcd/pkg/oci/client"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/spf13/cobra"
+)
+
+var diffArtifactCmd = &cobra.Command{
+	Use:   "artifact",
+	Short: "Diff Artifact",
+	Long: `The diff command does a build, then it performs a server-side dry-run and prints the diff.
```
stefanprodan, line 33: Please update this to reflect what the command actually does.

cmd/flux/diff_artifact.go
```diff
@@ -0,0 +1,111 @@
...
+	oci "github.com/fluxcd/pkg/oci/client"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/spf13/cobra"
+)
+
+var diffArtifactCmd = &cobra.Command{
+	Use:   "artifact",
+	Short: "Diff Artifact",
+	Long:  `The diff artifact command computes the diff between the remote OCI artifact and a local directory or file`,
+	Example: `# Check for differences between the files are included in an OCI image and the local manifests
```
stefanprodan, line 34:
```suggestion
	Example: `# Check if local files differ from remote
```

stefanprodan (approved):
LGTM

Thanks @developer-guy 🎖️

Issue #3212, opened 2022-10-14
Bring CLOMonitor Score to 100%


alinskens:
This repo is signed up as part of the KubeCon [Security Slam](https://events.linuxfoundation.org/kubecon-cloudnativecon-north-america/attend/experiences/#security-slam). I'm bringing to your attention the checklist from the official [CLOMonitor page for Flux](https://clomonitor.io/projects/cncf/flux-project) -- it refreshes every hour, so it should be up-to-date.

### Summary

**Repository**: flux2
**URL**: https://github.com/fluxcd/flux2
**Checks sets**:  `CODE`
**Score**: 86

### Checks passed per category

| Category  |  Score  |
| :----------------- | --------: |
|  Documentation  |  100%  |
|  License  |  75%  |
|  Best Practices  |  94%  |
|  Security  |  80%  |
|  Legal  |  n/a  |

### Documentation [100%]

  - [x] Changelog ([_docs_](https://clomonitor.io/docs/topics/checks/#changelog))
  - [x] [Contributing](https://github.com/fluxcd/flux2/blob/main/CONTRIBUTING.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#contributing))
  - [x] [Maintainers](https://github.com/fluxcd/flux2/blob/main/MAINTAINERS) ([_docs_](https://clomonitor.io/docs/topics/checks/#maintainers))
  - [x] [Readme](https://github.com/fluxcd/flux2/blob/main/README.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#readme))
  
### License [75%]

  - [x] Apache-2.0 ([_docs_](https://clomonitor.io/docs/topics/checks/#license_spdx_id))
  - [x] Approved license ([_docs_](https://clomonitor.io/docs/topics/checks/#license_approved))
  - [ ] License scanning ([_docs_](https://clomonitor.io/docs/topics/checks/#license_scanning))
  
### Best Practices [94%]

  - [ ] Artifact Hub badge ([_docs_](https://clomonitor.io/docs/topics/checks/#artifacthub_badge))
  - [x] Contributor License Agreement ([_docs_](https://clomonitor.io/docs/topics/checks/#cla)) `EXEMPT`
  - [x] Developer Certificate of Origin ([_docs_](https://clomonitor.io/docs/topics/checks/#dco))
[...]

PR #3231, opened 2022-10-20
Adjusted workflow permissions

eddie-knight:
Part of #3212 

Each workflow should be tested prior to merging these changes!

Signed-off-by: Eddie Knight <knight@linux.com>


stefanprodan (approved):
LGTM

Thanks @eddie-knight 🏅

PR #3249, opened 2022-10-25
Remove file reading from bootstrap package

phillebaba:
This change moves all file reading for certificates and PGP from code located in pkg to the CLI command. The purpose for this is to make sharing the code easier where the content is not expected to be in files.

### DIFF
--- a/cmd/flux/bootstrap_gitlab.go
+++ b/cmd/flux/bootstrap_gitlab.go
@@ -215,19 +215,18 @@ func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
 	if bootstrapArgs.tokenAuth {
 		secretOpts.Username = "git"
 		secretOpts.Password = glToken
-
-		if bootstrapArgs.caFile != "" {
-			secretOpts.CAFilePath = bootstrapArgs.caFile
-		}
+		secretOpts.CAFile = caBundle
 	} else {
+		keypair, err := sourcesecret.LoadKeyPairFromPath(bootstrapArgs.privateKeyFile, gitArgs.password)
+		if err != nil {
+			return err
+		}
+		secretOpts.Keypair = keypair
 		secretOpts.PrivateKeyAlgorithm = sourcesecret.PrivateKeyAlgorithm(bootstrapArgs.keyAlgorithm)
 		secretOpts.RSAKeyBits = int(bootstrapArgs.keyRSABits)
 		secretOpts.ECDSACurve = bootstrapArgs.keyECDSACurve.Curve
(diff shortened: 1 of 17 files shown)


pkg/bootstrap/bootstrap_plain_git.go
```diff
@@ -195,7 +196,7 @@ func (b *PlainGitBootstrapper) ReconcileSourceSecret(ctx context.Context, option
 	}
 
 	// Return early if exists and no custom config is passed
-	if ok && len(options.CAFilePath+options.PrivateKeyPath+options.Username+options.Password) == 0 {
```
phillebaba, line 198: Verify that this logic is the same as the previous.

cmd/flux/bootstrap_gitlab.go
```diff
@@ -246,6 +246,19 @@ func bootstrapGitLabCmdRun(cmd *cobra.Command, args []string) error {
...
+	if bootstrapArgs.gpgKeyRingPath != "" {
+		r, err := os.Open(bootstrapArgs.gpgKeyRingPath)
+		if err != nil {
+			return fmt.Errorf("unable to open GPG key ring: %w", err)
+		}
+		entityList, err = openpgp.ReadKeyRing(r)
+		if err != nil {
+			return err
+		}
+	}
```
pjbgf, line 260: This seems to be the same across `bitbucket`, `git`, `github` and `gitlab`. Can we extract as a func and reuse across them?

pkg/manifestgen/sourcesecret/sourcesecret.go
```diff
@@ -65,11 +65,9 @@ func Generate(options Options) (*manifestgen.Manifest, error) {
 	var keypair *ssh.KeyPair
 	switch {
 	case options.Username != "" && options.Password != "":
-		// noop
-	case len(options.PrivateKeyPath) > 0:
-		if keypair, err = loadKeyPair(options.PrivateKeyPath, options.Password); err != nil {
-			return nil, err
-		}
+	// noop
```
pjbgf, line 68: Without indentation it comes across as it relates to the next case, instead of the previous.

cmd/flux/bootstrap_gitlab.go
phillebaba: Good idea. Moved it into a shared function.

stefanprodan (approved):
LGTM

Thanks @phillebaba 🏅

pjbgf (approved):
Apart from the nit, LGTM

Thanks @phillebaba! :bow:

Issue #3273, opened 2022-11-01
Refactor Git operations and introduce go-git support for Azure DevOps and AWS CodeCommit


pjbgf:
This enables Git implementation alignment with source controller, therefore ensuring that anything bootstrapable via CLI can be reconciled by the controller (and vice-versa).

Ref: https://github.com/fluxcd/pkg/issues/245

PR #3289, opened 2022-11-07
Fix committing deleted files

phillebaba:
The bootstrap code makes use of some wrapper functions to commit files. It has a workaround for a go-git bug related to broken symlinks. The issue is that this workaround does not take into account deleted files and will fail if one tries to commit a deleted file. Bootstrap will not delete any committed files from git but the Terraform provider will as it tracks the file in the state. This introduces a check which skips the broken symlink test for deleted files.

### DIFF
--- a/cmd/flux/bootstrap_bitbucket_server.go
+++ b/cmd/flux/bootstrap_bitbucket_server.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/ProtonMail/go-crypto/openpgp"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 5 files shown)


pjbgf:
`go-git` using the default `go-billy` fs has an issue around abs symlinks pointing outside the path of their soft chroot implementation.
We no longer have that issue with our own implementation of go-billy fs: https://github.com/fluxcd/pkg/tree/main/git/gogit/fs.

Based on the above, I am not sure this will be needed once https://github.com/fluxcd/flux2/issues/3273 is finished.
Do you mind giving it a go with the new fs to confirm whether the issues still exists with the Terraform provider?

aryan9600:
@phillebaba i tested deleting committed files using https://github.com/fluxcd/pkg/blob/main/git/gogit/repository_client.go#L241 and its working fine, so i think this can be superseded by #3299 when it gets merged

phillebaba:
Sounds good will close this PR then and switch when that PR gets merged.

PR #3294, opened 2022-11-07
Aggregate errors in uninstall functions

phillebaba:
Adds an error return value to the uninstall functions. The cli ignores the returned error as it relies on the error message being logged to the user. It is however useful for other programs which want the actual returned error.


pkg/uninstall/uninstall.go
```diff
@@ -35,10 +35,12 @@ import (
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
 	notificationv1 "github.com/fluxcd/notification-controller/api/v1beta1"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/hashicorp/go-multierror"
```
stefanprodan, line 38: Let's use `k8s.io/apimachinery/pkg/util/errors` example here https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_controller.go#L181

stefanprodan (approved):

PR #3323, opened 2022-11-16
build: Pin GitHub Actions

pjbgf:
The main benefit of pinning GitHub actions is the determinism it brings in terms of what version of a given action will be executed. This is a step towards having hermetic builds.

Once pinned to a commit, Dependabot will automatically issue PRs to update to newer versions. For the time being this will happen every Monday.

Pinned versions is the only security metric from OpenSSF scorecard that this repository currently have a zero score.

image

---

This PR focus on external actions. Flux actions that live in `pkg/actions` will be dealt with separately.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,5 +3,7 @@ version: 2
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
+    labels: ["area/build"]
     schedule:
+      # by default this will be on a monday.
       interval: "weekly"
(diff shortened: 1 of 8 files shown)


pjbgf:
The comments are not a hard requirement and the official documentation is quite slim on it.

My current understanding is that it can be used to pin to a specific semver. I added the comments across external dependencies, apart from the GitHub official actions, which we won't restrict updates to a given semver.

stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #3324, opened 2022-11-17
Update kubectl and remove nsswitch.conf in flux-cli image

stefanprodan:
Changes:
- Remove nsswitch.conf creation as it's now included in [Alpine 3.16](https://git.alpinelinux.org/aports/commit/?id=348653a9ba0701e8e968b3344e72313a9ef334e4)
- Update kubectl version to 1.25.4


pjbgf (approved):
LGTM

Issue #3278, opened 2022-11-03
build: Fix E2E tests for Azure DevOps


pjbgf:
The E2E tests for Azure DevOps are broken. As part of fixing them we should also try to migrate away from libgit2 and convert to the `fluxcd/pkg` and `fluxcd/go-git` instead.

PR #3342, opened 2022-11-22
Use go-git for Azure e2e tests

somtochiama:
Ref #3278


stefanprodan (approved):
LGTM

Thanks @somtochiama 

Going to merge this so we can run the tests in the target branch.

Issue #3256, opened 2022-10-27
Way to push OCI artifact from stdin


lictw:
### Describe the bug

Hi there!

My kustomize includes other kustomize and I don't want to push whole repo as artifact (mainly due to I don't want path in KS other than ./), I can build kustomize locally and push artifact from file, but how I can achieve that without tmp file?

### Steps to reproduce

```
$ flux push artifact oci://localhost:5000/test --path <(kustomize build .) ..
► pushing artifact to localhost:5000/test
✗ pushing artifact failed: open /dev/fd/633325431970: no such file or directory
```

### Expected behavior

`kustomize build . | flux push artifact ..`

### OS / Distro

N/A

### Flux version

v0.36.0

### Flux check

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3355, opened 2022-11-28
`Push/Build artifacts` from stdin

somtochiama:
Closes:  #3256
Depends on: https://github.com/fluxcd/pkg/pull/418

### DIFF
--- a/cmd/flux/build_artifact.go
+++ b/cmd/flux/build_artifact.go
@@ -17,7 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/flux/build_artifact.go
```diff
@@ -66,18 +72,47 @@ func buildArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
 	return nil
 }
+
+func saveStdinToFile() (string, error) {
+	b, err := io.ReadAll(bufio.NewReader(os.Stdin))
+	if err != nil {
+		return "", err
+	}
+	b = bytes.TrimRight(b, "\r\n")
+	f, err := os.CreateTemp("", "*.yaml")
```
stefanprodan, line 107: How is this file placed in the tarball? Does it contain only the file or it has all the tmp dir path also?

somtochiama: It has all the `tmp` dir path too.

I considered updating the `Build` in `fluxcd/pkg` so that it puts only the file if an absolute path is passed in but that conflicts with the behaviour when a relative path is passed (it has the full dir path too)

stefanprodan: If a single file is passed (either via stdin or file path) the tarball should contain only that file in the root. Please make the necessary changes in pkg. Thanks!

stefanprodan:
@somtochiama please update the package to [oci/v0.16.0](https://github.com/fluxcd/pkg/releases/tag/oci%2Fv0.16.0)

Issue #3078, opened 2022-09-06
RFC - Consolidate validation of remote servers


pjbgf:
Some Flux controllers allow Flux users to trust remote servers on a per-object basis.
For example, Source Controller support CA Bundles to be set at object level, so that TLS connections can be established and trusted when custom PKIs or self-signed certificates are being used. The same applies to `known_hosts` to SSH connections.

A new RFC should be created to consolidate the topic across the controllers and establish:
- Supported (vs recommended) mechanisms that change the CA Store or `known_hosts`.
- Ways for Platform Admins to toggle per-object trust overrides.

PR #3366, opened 2022-12-02
[RFC] Trust Store for TLS and SSH

pjbgf:
Fixes https://github.com/fluxcd/flux2/issues/3078.

### DIFF
--- a/rfcs/0000-trust-store-for-tls-and-ssh/README.md
+++ b/rfcs/0000-trust-store-for-tls-and-ssh/README.md
@@ -43,7 +43,7 @@ security stand-point.
 ## Proposal
 
 For configuring system-wide trust, Flux would rely on the well-established OS-level
-trust store. When dynamically mounting of the trust store is required, it will be
+trust stores. When dynamically mounting of the trust store is required, it will be
 enabled by using Kubernetes `Secret` and `ConfigMap` mounting. When immutable trust
 store is required, users can build their own version of the controllers, with their
 baked-in settings.
@@ -128,6 +128,12 @@ under a `known_hosts` key.
 Known hosts configured this way will be aggregated with the ones defined at both
 system and controller levels.
 
+#### Pre-populated trust store
+
+Flux container images would be pre-populated with [/etc/ssh/ssh_known_hosts] from
+the main Git SaaS providers. As a result, users will only need to update their SSH
+Trust Store for custom or less well known servers.
+
 #### TLS
 
 In TLS, the remote server identity is based on [public key infrastructure] and the
@@ -142,8 +148,8 @@ TLS communications against untrusted remote servers are aborted.
 
 #### Controller-level Trusted Certificates
 
-*NOTE:* this requires no changes on the controllers, as this is based on the ways
-that TLS surface the trust store. This RFC only formalizes it as a supported
+**Note:** this requires no changes to the controllers, as this is based on the ways
+in which TLS surface the trust store. This RFC only formalizes it as a supported
 approach.
 
 To trust CAs that are not part of the root trusted CAs, the OS level trust store
@@ -186,8 +192,8 @@ Patch required on the main `kustomization.yaml`:
 
 A new field is to be introduced into the existing kinds `Bucket`, `GitRepository`,
 `HelmRepository`, `OCIRepository`, `ImageUpdateAutomation`, `Provider` and
-`ImageRepository`, to allow users to expand on the controller-level known hosts for
-SSH operations:
+`ImageRepository`, to allow users to expand on trusted CAs at controller-level for
+HTTPS operations:
 
 ```yaml
 spec:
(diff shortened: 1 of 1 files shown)


rfcs/0000-trust-store-for-tls-and-ssh/README.md
```diff
@@ -0,0 +1,293 @@
...
+or `spec.trustStore.tls.configMapRef`, not both. Either option should contain the
+data under a `caFile` key.
+
+CA bundles configured this way will be aggregated with the ones defined at both
+system and controller levels.
+
+### Enabling Object-Level Trust Store
+
+Object-level trust store expansion is disabled by default. To enable it start
+the controller with:
```
makkes, line 210: I would love to see this expanded a little to provide reasoning for this default setting. With the current Flux version 0.37 the trust store can always be expanded for objects such as HelmRepositories or GitRepositories by defining a `caFile` field in the referenced Secret. Given many admins will unlikely deviate from the defaults, users would no longer be able to do that. Please correct me if I'm wrong here.

rfcs/0000-trust-store-for-tls-and-ssh/README.md
```diff
@@ -0,0 +1,293 @@
...
+The trust store can be expanded by either setting `spec.trustStore.ssh.secretRef` or
+`spec.trustStore.ssh.configMapRef`, not both. Either option should contain the data
+under a `known_hosts` key.
+
+Known hosts configured this way will be aggregated with the ones defined at both
+system and controller levels.
+
+#### TLS
+
+In SSH, the remote server identity is based on [public key infrastructure] and the
```
souleb, line 133:
```suggestion
In TLS, the remote server identity is based on [public key infrastructure] and the
```

rfcs/0000-trust-store-for-tls-and-ssh/README.md
```diff
@@ -0,0 +1,293 @@
...
+trust store. When dynamically mounting of the trust store is required, it will be
+enabled by using Kubernetes `Secret` and `ConfigMap` mounting. When immutable trust
+store is required, users can build their own version of the controllers, with their
+baked-in settings.
+
+TLS and SSH use different techniques to establish the identity of remote servers,
+each relying on its own trust store.
+
+The sections below will dive into the specifics of each one, highlighting their
+details, changes required and example of the propose use.
```
souleb, line 55:
```suggestion
details, changes required and example of the proposed usage.
```

rfcs/0000-trust-store-for-tls-and-ssh/README.md
```diff
@@ -0,0 +1,293 @@
...
+      kind: Deployment
+      name: "(kustomize|image-automation|source|image-reflector|helm|notification)-controller"
+```
+
+#### Object-level Trusted Certificates Expansion
+
+A new field is to be introduced into the existing kinds `Bucket`, `GitRepository`,
+`HelmRepository`, `OCIRepository`, `ImageUpdateAutomation`, `Provider` and
+`ImageRepository`, to allow users to expand on the controller-level known hosts for
+SSH operations:
```
stefanprodan, line 190: Is this about SSH/know hosts or TLS?

pjbgf: Good spot, updated it accordingly.

rfcs/0000-trust-store-for-tls-and-ssh/README.md
pjbgf: This is certainly a point up for debate. The security invariant here was "Flux should only connect to trusted servers" and the criteria used was that most users would expect Flux to be secure by default.

> Given many admins will unlikely deviate from the defaults

This is exactly the point. If something works, users won't look up the documentation and try understand the impact of security sensitive settings. Hence why ideally we should always prioritise security by design and by default.

The idea is that, by pre-populating the controller level SSH trust store with the top SaaS Git providers, a considerable amount of users would not have to worry about this. It will simply securely work - just like for TLS.
Then enterprise customers should configure this once, at controller level and be done with it, for both single and multi-tenancy. The only use case I see for Object-level trust stores to be used **in production**, would be in a Namespace as a Service multi-tenancy model, which would further weaken an already fragile security posture.

rfcs/0000-trust-store-for-tls-and-ssh/README.md
```diff
@@ -0,0 +1,344 @@
...
+    target:
+      kind: Deployment
+      name: "(kustomize|image-automation|source|image-reflector|helm|notification)-controller"
+```
+
+#### Object-level Known Hosts Expansion
+
+A new field is to be introduced into the existing kinds `ImageUpdateAutomation` and
+`GitRepository`, to allow users to expand on the controller-level known hosts for
+SSH operations:
```
stefanprodan, line 115: I find this very problematic, why would we allow for a namespaced object own by a tenant to alter the trust store for all tenants? This also makes Flux bootstrap break, since you can't control which CR is reconciled first, if people add a trust store in one CR we can't reconcile that first to make it available to others.

pjbgf: It does, it expands the trust store for that object only.

stefanprodan: How does this cope with the know_hosts from the `secretRef`? Can you explain the differences in the doc please?

pjbgf: I will rephrase to make that clearer. But basically that's what we already do at present based on the CA bundles provided by secrets, however this would be based on a field instead.

PR #3368, opened 2022-12-04
[RFC] Add build-in TLS support

pjbgf:
Based on the [discussion](https://github.com/fluxcd/flux2/discussions/3066) instigated by @rashedkvm.

More context can be found on the [recordings](https://www.youtube.com/watch?v=s1wb7W3jh3g) from Sept 8th Flux Dev Meeting.

### DIFF
--- a/rfcs/0000-build-in-tls-support/README.md
+++ b/rfcs/0000-build-in-tls-support/README.md
@@ -1,0 +1,137 @@
+# RFC-NNNN Built-in TLS support for Source and Notification Controllers
+
+**Status:** provisional
+
+**Creation date:** 2022-12-05
+
+**Last update:** 2022-12-05
+
+## Summary
+
+Create built-in TLS support for both Source and Notification controllers,
+so that in-cluster cross-component communications are always encrypted.
+
+## Motivation
+
+Due to internal company policies, Data Protection laws and regulations,
+Flux users may be required to enforce encryption of data whilst in-transit.
+As of `v0.37.0`, there is no built-in way to provide TLS for Flux' endpoints.
+
+The existing endpoints being:
+
+1. Source Controller: artifacts storage, used by Kustomize and HelmRelease
+controllers.
+2. Notification Controller: events receivers, used by all other Flux components
+to submit events.
+3. Notification Controller: webhooks for external services to contact Flux.
+
+The webhooks endpoing would be typically used for external use only, and in
+that use case the recommended approach is to enable TLS at ingress.
+However, this proposal covers the first two endpoints to allow always encrypted
+in-cluster communications.
+
+### Goals
+
+- Add TLS support for Notification and Source controllers.
+- Establish a migration process when enabling TLS on existing clusters.
+- Define best practices on enabling the feature.
+
+### Non-Goals
+
+- Change how CA trust is established across Flux containers.
+- Define TLS rotation mechanisms.
+
+## Proposal
+
+Controllers hosting web servers will introduce a new flag to configure the
+[TLS secret]: `--tls-secret`. The Secret type must be `kubernetes.io/tls`.
+
+Invalid secret types, or content, will cause the controller to fail to start.
+
+### Source Controller
+
+When TLS is enabled, the File Storage endpoint will serve both HTTP and HTTPS.
+Existing source objects previously reconciled won't be patched to a new
+`.status.artifact.url`, but all new reconciliations will cause that field to
+start with `https://`.
+
+The HTTP endpoint will only serve a permanent redirect (301) to the HTTPS endpoint,
+enabling a smooth transition when users enable this feature in existing clusters.
+
+The HTTPS endpoint will be served on port `9443` by default, which will be
+configurable via new flag `--storage-https-addr` or environment variable
+`STORAGE_HTTPS_ADDR`. The existing `source-controller` service will bind port
+`443` to `9443`.
+
+### Notification Controller
+
+When TLS is enabled, the HTTP endpoints will be disabled.
+
+#### Events endpoint
+
+The events HTTPS endpoint will be served on port `9443` by default, which will be
+configurable via new flag `--events-https-addr`. The existing `notification-controller`
+service will bind port `443` to `9443`.
+
+All flux components will need to be started with the (existing) flag below, for the
+TLS only endpoint to be used:
+`--events-addr=https://notification-controller.flux-system.svc.cluster.local./`
+
+### User Stories
+
+<!--
+Optional if existing discussions and/or issues are linked in the motivation section.
+-->
+
+### Alternatives
+
+#### Delegate in-trasit data encryption to CNI or Service Mesh
+
+Instead of built-in support, it was considered the option of delegating this to
+external components that support the enablement of end-to-end encryption via TLS
+or mTLS. One of the issues with this approach is that the data would still come
+out of Flux unencrypted, and therefore privileged co-located components in the
+same worker node could be privy to the exchanged plain-text.
+
+Another point considered for not pursuing this alternative was that, being one of
+the first workloads inside a cluster, Flux should be self-reliant in assuring the
+confidentiality and integrity of its communications.
+
+#### Auto-detection of TLS events endpoint
+
+Instead of having to change the `--event-addr` on each Flux component, an auto
+detection mechanism could be implemented to prefer the TLS endpoint whenever the
+notification controller's service has one working.
+
+The challenge with such approach is that the security of cross component
+communication would be non deterministic from a client perspective, and would depend
+on the Notification controller having an active TLS endpoint at the time in which each
+Flux component started. The current approach requires additional configuration,
+but fails safe.
+
+## Design Details
+
+### TLS Versions and Ciphers Supported
+
+For reduced complexity and increased security, only TLS 1.3 is supported.
+
+### Detect TLS Certificate changes
+
+Controllers will watch the TLS secret for changes, once detected it will trigger a
+pod restart to refresh the certificates being used.
+
+### Expand TLS support for the webhooks endpoint
+
+The TLS support could be expanded to the Notification controller webhooks endpoint
+to enable secure in-cluster communications.
+
+## Implementation History
+
+<!--
+Major milestones in the lifecycle of the RFC such as:
+- The first Flux release where an initial version of the RFC was available.
+- The version of Flux where the RFC graduated to general availability.
+- The version of Flux where the RFC was retired or superseded.
+-->
+
+[TLS secret]: https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets


rfcs/0000-build-in-tls-support/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Detect TLS Certificate changes
+
+Controllers will watch the TLS secret for changes, once detected it will trigger a
+pod restart to refresh the certificates being used.
+
+### Expand TLS support for the webhooks endpoint
+
+The TLS support could be expanded to the Notification controller webhooks endpoint
+to enable secure in-cluster communications.
```
pjbgf, line 124: The main issue I see with supporting this is the hybrid scenario in which we need to support in-cluster and external (via ingress). In such scenarios, the ingress would need to trust the CA used by Flux, which can be problematic from a security stand-point.

rfcs/0000-build-in-tls-support/README.md
```diff
@@ -0,0 +1,135 @@
...
+notification controller's service has one working.
+
+The challenge with such approach is that the security of cross component
+communication would be non deterministic from a client perspective, and would depend
+on the Notification controller having an active TLS endpoint at the time in which each
+Flux component started. The current approach requires additional configuration,
+but fails safe.
+
+## Design Details
+
```
pjbgf, line 111: The approach requires a way for the CA used to generate the TLS cert/private key to be trusted by all Flux controllers. That is currently being defined via https://github.com/fluxcd/flux2/pull/3366.

rashedkvm:
Thank you @pjbgf. Will review shortly.

rashedkvm (commented):
@pjbgf thank you for restarting the encrypted communication discussion for Flux components. Am I understanding this correctly that this feature should not be used in a service mesh setup?

pjbgf:
> @pjbgf thank you for restarting the encrypted communication discussion for Flux components. Am I understanding this correctly that this feature should not be used in a service mesh setup?

@rashedkvm The use of CNI or Service Mesh are alternatives, which could still be used regardless of the built-in support. However, I cannot see why users would use a Service Mesh to protect Flux inter-component comms when the built-in TLS is enabled, as IMO they are mutually exclusive. Although I could see cases in which some deployments may have CNI encryption + Flux built-in TLS.

rfcs/0000-build-in-tls-support/README.md
```diff
@@ -0,0 +1,137 @@
...
+### TLS Versions and Ciphers Supported
+
+For reduced complexity and increased security, only TLS 1.3 is supported.
+
+### Detect TLS Certificate changes
+
+Controllers will watch the TLS secret for changes, once detected it will trigger a
+pod restart to refresh the certificates being used.
```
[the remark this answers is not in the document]
pjbgf: I like the idea. This would cause all mid-flight requests to be dropped, which is a minor impact compared to rebuilding the storage. Besides, the mid-fight requests would be automatically retried within seconds anyway.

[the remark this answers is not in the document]
rashedkvm: Yes, definitely that will be a considerable saving over the expensive restarts.

rashedkvm:
> @rashedkvm The use of CNI or Service Mesh are alternatives, which could still be used regardless of the built-in support. However, I cannot see why users would use a Service Mesh to protect Flux inter-component comms when the built-in TLS is enabled, as IMO they are mutually exclusive. 

that is my understanding too. Thanks!

rfcs/0000-build-in-tls-support/README.md
```diff
@@ -0,0 +1,137 @@
...
+- Add TLS support for Notification and Source controllers.
+- Establish a migration process when enabling TLS on existing clusters.
+- Define best practices on enabling the feature.
+
+### Non-Goals
+
+- Change how CA trust is established across Flux containers.
+- Define TLS rotation mechanisms.
+
+## Proposal
```
phoban01, line 44: Would the proposed architecture be compatible with SPIFFE/SPIRE? I can envisage a scenario where I would like to manage identity using SPIFFE SVIDs delivered via SPIRE rather than via Kubernetes TLS secrets. In practice I think this would mean adding a unix domain socket flag to the controllers with an accompanying listener to receive SVID bundles from the SPIRE agent.

rfcs/0000-build-in-tls-support/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+Controllers hosting web servers will introduce a new flag to configure the
+[TLS secret]: `--tls-secret`. The Secret type must be `kubernetes.io/tls`.
```
alukyan, line 47: Why take dependency on a specific K8s secret hosting approach directly in the components? Can you please use a file reference instead. Then it will be easier to support other secret hosting approaches such as cloud key vaults. 
For your information in our enterprise environments hosting any kind of secrets as K8s secrets is prohibited.

stefanprodan: How are you using Flux then? All our APIs rely on Kubernetes Secrets, you can’t connect to Git/OCI/Helm repositories unless you store the auth in a secret.

makkes: > hosting any kind of secrets as K8s secrets is prohibited

So you're using Secrets Store CSI Driver and mounting them directly into containers? I see why this might be a reasonable policy but as Stefan said, Flux doesn't work without Secrets at the moment and likely never will. Otherwise you'd have to mount all the Secrets any user might ever need into the Flux controller containers.

stefanprodan: Consuming secrets as files breaks Flux multi-tenancy. We rely on Kubernetes RBAC to isolate tenant, reading files from the container FS is not subject to RBAC while reading secrets from the API is. Have you considered enabling encryption at rest for etcd?

makkes: > @makkes - correct, we use Secret CSI drivers, or you can create a wrapper/starter script to launch the binaries and this starter script pulls the secrets from key vaults and places them on tempfs volumes.
> 
> @stefanprodan - based on the documentation I see that Flux can use cloud identities to access artifact storages. We can still use K8s secrets as a config store for not secret material - such as public keys, application IDs, etc. But we cannot use K8s secrets for storing secret material like private keys. Most Kubernetes components consume secrets as file references - then if someone wants to use K8s secrets to store secrets they can always mount them as files in the volume.
> 
> Consuming secret material directly from K8s secret resources is a gap that must be fixed for Flux to be used in different deployments that rely on consuming secrets as files.

As I said, if Flux would exclusively consume secret data from files, you as the Flux operator would have to mount all the Secrets any user might ever need into the Flux controller containers which

1. is not exactly a lightweight process for the Flux operator on the cluster
2. is not compatible with resource isolation on multi-tenant clusters (as Stefan noted)

alukyan: > Consuming secrets as files breaks Flux multi-tenancy. We rely on Kubernetes RBAC to isolate tenant, reading files from the container FS is not subject to RBAC while reading secrets from the API is.

I must admit I do not understand the security model for Flux multitenancy.

If Flux service account has RBAC permissions to read tenant K8s secrets via API to put them in memory how it stat different from mounting secrets as files in Flux filesystem? If I have access to the Flux pod I can access the file system directly, and I am one REST call away from accessing secrets via API.

And to avoid confusion I would like to mention that my area of concern is not how Source controller interacts with artifact sources. This is where multitenancy matters.
My concern is about FluxV2 components interact with each other - For example Kustomization controller retrieves artifacts served by Source controller via http.FileServer(). I do not see how multitenancy matters here.

makkes: Take a look at https://fluxcd.io/flux/security/best-practices/#multi-tenancy-lock-down for details on how Flux enables secure multi-tenant scenarios. It makes heavy use of impersonation, leveraging Kubernetes RBAC and that's not possible with files mounted into the container.

alukyan: @makkes - I read that article, service account impersonation with tenant level control over the SA access makes sense to me.

However, I still do not understand how this article relates to the TLS support discussed here. Namely,

> 1. Source Controller: artifacts storage, used by Kustomize and HelmRelease
controllers.

both sides are owned by Flux system itself. Where are the multitenancy here?

pjbgf: There is a lot to unpack here, I am not sure I will be able to do it justice. 

> Why take dependency on a specific K8s secret hosting approach directly in the components? 

That's a great question. Flux must support being the very first thing deployed into a cluster, this proposal aims to support built-in TLS for that scenario without additional dependencies. 

> For your information in our enterprise environments hosting any kind of secrets as K8s secrets is prohibited.

That's a common requirement in highly regulated environments. The best way I can see your requirement being met securely (in the context of this RFC) is with an integration to Cloud Key Management Services via contextual authentication. This way both key material, and the authentication needed for it are only kept in memory. In my opinion that could be handled as a follow-up RFC.

> you can create a wrapper/starter script to launch the binaries and this starter script pulls the secrets from key vaults and places them on tempfs volumes.

I believe that file-based support for sensitive information, such as key material, would just increase the impact of any directory traversal vulnerability in the controllers that were to support this. If that is a hard requirement, I would suggest it not to be enabled by default, and be kept behind a `--insecure-` prefixed flag, so users think through their threat model and requirements before using it.

alukyan: > I believe that file-based support for sensitive information, such as key material, would just increase the impact of any directory traversal vulnerability in the controllers that were to support this.

Thanks for bringing up this perspective. I would like to mention though that many (if not most) of the existing industry solutions that communicate or serve the content over TLS support file-based key material, so it hardly can be marked as `insecure`.

anagg929: I might be confused here. But wouldn't the TLS k8s secret also mounted as a file in the pod. The certwatcher project suggested allows file based certs only. 
https://github.com/kubernetes-sigs/controller-runtime/blob/a784ee78d04bdf216006adfbc313328a0aded88a/pkg/certwatcher/certwatcher.go#L36

Thus not sure how giving an option for a TLS on file-based key inside the pod can be less secure as even with secrets we need to do the same.

I understand going default with secret to not add any additional dependencies for the default secure path. But if the customer is willing to do the work to use a Secret CSI drivers to mount a file directly, shouldn't that be allowed.

makkes: The Secrets aren't mounted into the Flux Pods but read into memory through the Kubernetes API.

stefanprodan: We could add a flag for pointing to a cert on disk, then it's up to the Flux admin to figure out the mounts and so on. With this flag, we'll read the cert at startup only, the Flux admin can bump some pod annotation in Git to restart the controllers to reload the cert. This flag e.g. `--tls-file` should be mutually exclusive with `--tls-secret`, if both are set, the controllers will panic and crash loop.

anagg929: Thanks , this should resolve the cases where admin want to mount the cert on their own.
A bit of query why not use use the certwatcher on the --tls-file also to avoid the restart ?

stefanprodan: > Why not use use the certwatcher on the --tls-file also to avoid the restart ?

Ah I haven't looked into that, if it supports it, then sure.

PR #3389, opened 2022-12-09
Push/Build artifacts from stdin

somtochiama:
Closes: https://github.com/fluxcd/flux2/issues/3256
Depends on: https://github.com/fluxcd/pkg/pull/418

Signed-off-by: Somtochi Onyekwere <somtochionyekwere@gmail.com>

PR #3418, opened 2022-12-17
Fix path on `flux push`

somtochiama:
Fixes bug introduced by https://github.com/fluxcd/flux2/pull/3389 where the path for the build cmd was used in the error message.

Also moves code for saving stdin to file before check for path

### DIFF
--- a/cmd/flux/push_artifact.go
+++ b/cmd/flux/push_artifact.go
@@ -129,8 +129,18 @@ func pushArtifactCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if _, err := os.Stat(pushArtifactArgs.path); err != nil {
-		return fmt.Errorf("invalid path '%s', must point to an existing directory or file", buildArtifactArgs.path)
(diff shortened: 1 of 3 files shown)


cmd/flux/push_artifact.go
```diff
@@ -129,8 +129,18 @@ func pushArtifactCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if _, err := os.Stat(pushArtifactArgs.path); err != nil {
-		return fmt.Errorf("invalid path '%s', must point to an existing directory or file", buildArtifactArgs.path)
+	path := pushArtifactArgs.path
+	if pushArtifactArgs.path == "-" {
+		path, err = saveStdinToFile()
```
pjbgf, line 134: `saveStdinToFile` is leaving the temp file opened. We need to close it before returning, otherwise `os.Remove()` will fail in some systems.

Please add a test for the stdin feature.

pjbgf (approved):
LGTM

Thanks @somtochiama! :bow:

PR #3465, opened 2023-01-06
Fix fluxcd/website#1347

kingdonb:
The docs are missing a `powershell` in `flux completion powershell`


javiermarasco:
Thank you @kingdonb this is exactly what I meant 👍.

Issue #3455, opened 2023-01-01
GitHub Actions unreliable


fritterhoff:
### Describe the bug

I'm trying to automate FluxCD Updates using the recommended workflow (https://github.com/fluxcd/flux2/tree/main/action#automate-flux-updates). Sadly this fails randomly multiple times a day.

### Steps to reproduce

1. Setup workflow https://github.com/fluxcd/flux2/tree/main/action#automate-flux-updates
2. Monitor Success/Failure of GH Action

### Expected behavior

Reliable execution of workflow

### Screenshots and recordings

```
2023-01-01T10:08:13.0895617Z ##[group]Run fluxcd/flux2/action@main
2023-01-01T10:08:13.0895864Z with:
2023-01-01T10:08:13.0896033Z   arch: amd64
2023-01-01T10:08:13.0896219Z ##[endgroup]
2023-01-01T10:08:13.1231060Z ##[group]Run ARCH=amd64
2023-01-01T10:08:13.1231379Z [36;1mARCH=amd64[0m
2023-01-01T10:08:13.1231588Z [36;1mVERSION=[0m
2023-01-01T10:08:13.1231762Z [36;1m[0m
2023-01-01T10:08:13.1231959Z [36;1mif [ -z $VERSION ]; then[0m
2023-01-01T10:08:13.1232361Z [36;1m  VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)[0m
2023-01-01T10:08:13.1232861Z [36;1mfi[0m
2023-01-01T10:08:13.1233034Z [36;1m[0m
2023-01-01T10:08:13.1233348Z [36;1mBIN_URL="https://github.com/fluxcd/flux2/releases/download/v${VERSION}/flux_${VERSION}_linux_${ARCH}.tar.gz"[0m
2023-01-01T10:08:13.1233704Z [36;1mcurl -sL ${BIN_URL} -o /tmp/flux.tar.gz[0m
2023-01-01T10:08:13.1233956Z [36;1mmkdir -p /tmp/flux[0m
2023-01-01T10:08:13.1234211Z [36;1mtar -C /tmp/flux/ -zxvf /tmp/flux.tar.gz[0m
2023-01-01T10:08:13.1292399Z shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
[...]

PR #3474, opened 2023-01-09
Improvements to Flux GitHub Action

ekristen:
## Overview

- Adds auth to curl command for GitHub API (optional)
- Adds example for token passing to documentation
- Updates documentation to not use `set-output` as it has been deprecated.

## Background

I've been witnessing more and more failed flux update action runs and several other workflows I have that make calls to `api.github.com` as best as I have been able to determine while debugging is that due to the workflows running on shared runners, and either an uptick in API calls or a change in the GitHub API backend on API interactions from workers, anonymous API limits are being hit more frequently which results in a 403 and failed `curl` command run.

The fix for me has been to pass a token through using the default `${{ secrets.GITHUB_TOKEN }}` to the curl command such that the API calls  are authenticated, this token gets it's own set of limits.

Fix: https://github.com/fluxcd/flux2/issues/3455

### DIFF
--- a/action/README.md
+++ b/action/README.md
@@ -35,6 +35,18 @@ You can download a specific version with:
           version: 0.32.0
 ```
 
+You can also authentication to the GitHub API using a secret
+
+```yaml
+    steps:
+      - name: Setup Flux CLI
+        uses: fluxcd/flux2/action@main
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+```
+
+This is useful if you are seeing failures on shared runners, those failures are usually API limits being hit.
+
 ### Automate Flux updates
 
 Example workflow for updating Flux's components generated with `flux bootstrap --path=clusters/production`:
--- a/action/README.md
+++ b/action/README.md
@@ -35,7 +35,7 @@ You can download a specific version with:
           version: 0.32.0
 ```
 
-You can also authentication to the GitHub API using a secret
+You can also authentication to the GitHub API using GitHub Action's `GITHUB_TOKEN`. For more information on this secrets, please see [about the github token secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret)
 
 ```yaml
     steps:
--- a/action/README.md
+++ b/action/README.md
@@ -62,7 +62,7 @@ jobs:
             --export > ./clusters/production/flux-system/gotk-components.yaml
 
           VERSION="$(flux -v)"
-          echo "::set-output name=flux_version::$VERSION"
+          echo "flux_version=$VERSION" >> $GITHUB_OUTPUT
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
--- a/action/README.md
+++ b/action/README.md
@@ -64,7 +64,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Flux CLI
         uses: fluxcd/flux2/action@main
       - name: Check for updates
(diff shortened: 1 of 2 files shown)


kingdonb:
Great detective work!

I saw this issue being reported very recently, as:

* #3455
(This likely fixes #3455)

kingdonb:
I've given this a quick test here:
https://github.com/kingdonb/bootstrap-repo/actions/runs/3883751782

It looks like the rate limit is still triggered several times a day, before making this change. 9cc4daf didn't quite implement the whole thing, I see there are some changes needed in the action itself, (so that it knows how to pass an Authorization token.) I made the rest of those changes in the test branch.

I have been able to run it here: https://github.com/kingdonb/bootstrap-repo/actions/runs/3883822670 I think I have implemented all the changes from this branch, and it LGTM (I added some further changes to resolve the outdated action warnings)

We can let it run there for a day or two, if we want to be sure the rate limit issue does not come back anymore, but anyway... paging back through my actions history I can see the last few days it has been reliably hitting the rate limit at least a couple of times a day on average, so this should be a good test.

If there's someone else who already has more confidence that this solves the issue, we can merge it straight away.

kingdonb:
While we are changing this, @ekristen could you please also bump the checkout and create-pull-request actions in the doc

`checkout@v3` and `create-pull-request@v4`

Then there are no deprecation warnings emitted 👍

ekristen:
@kingdonb docs updated to v3 and v4 respectively.

action/action.yml
```diff
@@ -15,6 +15,9 @@ inputs:
   bindir:
     description: "Optional location of the Flux binary. Will not use sudo if set. Updates System Path."
     required: false
+  token:
+    description: "GitHub Token used to authentication against the API (generally only needed to prevent quota limit errors)"
```
relu, line 19: Small typo here.
```suggestion
    description: "GitHub Token used to authenticate against the API (generally only needed to prevent quota limit errors)"
```

relu (changes requested):

relu (approved):
Looks good to me. Thank you @ekristen! :100:

action/README.md
```diff
@@ -35,6 +35,18 @@ You can download a specific version with:
           version: 0.32.0
 ```
 
+You can also authentication to the GitHub API using a secret
```
hiddeco, line 38:
```suggestion
You can also authentication to the GitHub API using an [encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets):
```

Might help a bit with users switching from a Kubernetes to GitHub (Action) context.

relu: I think that might confuse users actually because the value for `secrets.GITHUB_TOKEN` is pre-filled by GitHub Actions. Might be better to link to this instead? https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

hiddeco: That's an even better suggestion indeed.

ekristen: Adjusted the documentation to include the second link and provide a bit more context.

ekristen:
@relu The PR has been rebased.

action/README.md
```diff
@@ -35,6 +35,18 @@ You can download a specific version with:
           version: 0.32.0
 ```
 
+You can also authentication to the GitHub API using GitHub Action's `GITHUB_TOKEN`. For more information on this secrets, please see [about the github token secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret)
```
relu, line 38: Just a nit here and we can merge :)
```suggestion
You can also authenticate against the GitHub API using GitHub Actions' `GITHUB_TOKEN` secret.
For more information, please [read about the GitHub token secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret).
```

relu (changes requested):

kingdonb:
Good news, the repo where I am testing this change shows the issue seems resolved:

(https://github.com/kingdonb/bootstrap-repo/actions/workflows/update-flux.yaml?page=3 -- 9896 is the GHA job build that implemented this change, and you have to go back farther than that to find one of the random failures in this build log. 🎉 )

stefanprodan:
Superseded by https://github.com/fluxcd/flux2/pull/3501

PR #3477, opened 2023-01-10
fix(install-script): support $GITHUB_TOKEN

raffis:
I use the install script in our ci pipeline which happen to be executed quite frequently. 
We started to run into gh ratelimiting because of this script.

```
Run curl -s https://raw.githubusercontent.com/fluxcd/flux2/main/install/flux.sh | sudo bash
[INFO]  Downloading metadata https://api.github.com/repos/fluxcd/flux2/releases/latest
Error: Process completed with exit code 22.
```

This fix adds support for $GITHUB_TOKEN. If its not set it would just be an anonymous request as it was before.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -9,6 +9,10 @@ To install the latest release run:
 curl -s https://raw.githubusercontent.com/fluxcd/flux2/main/install/flux.sh | sudo bash
 ```
 
+**Note**: You may export an env `GITHUB_TOKEN` which is a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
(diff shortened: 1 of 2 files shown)


relu (commented):
Thank you for your contribution @raffis! :1st_place_medal: 

I've tested this change and it seems to work correctly both when GITHUB_TOKEN is set and not set.

Would you mind also updating the [documentation](https://github.com/fluxcd/flux2/blob/main/install/README.md) so that other users will be able to more easily find this?

raffis:
> Would you mind also updating the [documentation](https://github.com/fluxcd/flux2/blob/main/install/README.md) so that other users will be able to more easily find this?

Not at all, I updated the pr 👍🏻

PR #3501, opened 2023-01-21
Add GITHUB_TOKEN  to Flux GitHub Action

kingdonb:
Hello, the PR in #3474 is still blocked

I'm going to rebase and squash, but here's a reference to the original commit I made which adds just one change to the existing PR is:

* fac19820
to make the review easier, you might want to check that one first.

* c7a48769
* is 15385220 rebased against the current main branch without any other changes

I noticed that this doesn't only affect flux update job which runs hourly, but can even impact GitHub actions jobs that embed the "install flux" action inline!

Trying to get this unblocked, I've added the one suggestion from https://github.com/fluxcd/flux2/pull/3474#discussion_r1067178730 rebased and resubmitted this PR so that it can merge.

I'm going to go ahead and pre-emptively squash all the commits, as well, to simplify the merge.

Fix: https://github.com/fluxcd/flux2/issues/3455

### DIFF
--- a/action/README.md
+++ b/action/README.md
@@ -35,6 +35,20 @@ You can download a specific version with:
           version: 0.32.0
 ```
 
+You can also authenticate against the GitHub API using GitHub Actions' `GITHUB_TOKEN` secret.
+
+For more information, please [read about the GitHub token secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret).
+
+```yaml
+    steps:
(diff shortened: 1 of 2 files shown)


kingdonb:
I maintained the original author from #3474 the way I squashed, but it has my signature, hope this is OK – does not seem to bother DCO checker that I did it this way.

Thanks to @ekristen for submitting this, hope this re-submit can quickly push it over the line!

stefanprodan (approved):
LGTM

PR #3507, opened 2023-01-23
Update prometheus-community helm repo due to the suspension of OCI builds

thezanke:
### Proposed changes

This PR updates the `HelmRepository` to use the official non-OCI repository. We also disable chart verification for the "kube-prometheus-stack" `HelmChart` since it requires an OCI repository.

###  Reasoning

The "prometheus-community" developers have disabled OCI builds for now (see https://github.com/prometheus-community/helm-charts/pull/2841).

### DIFF
--- a/manifests/monitoring/kube-prometheus-stack/release.yaml
+++ b/manifests/monitoring/kube-prometheus-stack/release.yaml
@@ -11,8 +11,6 @@ spec:
       sourceRef:
         kind: HelmRepository
         name: prometheus-community
-      verify:
-        provider: cosign
       interval: 60m
   install:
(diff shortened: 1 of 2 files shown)


kingdonb:
Thanks for raising this @thezanke 

I think there are still some things to work out on the helm chart releaser itself, in order to get chart owners unblocked at supporting this. I'm also not sure anyone understood why the chart releaser action fails to publish an OCI chart sometimes.

stefanprodan (approved):
LGTM

Thanks @thezanke hopefully the issues with GH Actions will be solved upstream so we can revert it back.

PR #3508, opened 2023-01-23
Quote tested values in fluxcd/flux GitHub Action

### LINKED ISSUE #3509: The "install Flux" GitHub Action is failing
Fixed by #3508 

kingdonb:
Fix #3509 

This is an issue with improperly quoted variables created by #3474

The `fluxcd/flux2` GitHub Action is failing now, (since that PR was merged about 6h ago)

Also, curl should fail with an error code if the download results in 404 not found

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -28,23 +28,23 @@ runs:
         VERSION=${{ inputs.version }}
         TOKEN=${{ inputs.token }}
 
-        if [ -n $TOKEN ]; then
-          TOKEN=(-H "Authorization: token $TOKEN")
+        if [ -n "${TOKEN}" ]; then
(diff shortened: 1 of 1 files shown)


kingdonb:
My apologies everyone, the authorization header is being passed even when the TOKEN is null, which is something I should have tested for sure. This causes the download to fail unless you pass in a token.

kingdonb:
I made sure to document the workaround as publicly as I could in #3509 so nobody feels pressure to merge this quickly.

It's probably better if we give it a good thorough review. That being said I've had @seh look at it together with me, and we went through the rest of the script looking for issues of the same class, and fixed a few of them.

PR #3511, opened 2023-01-23
Fix fluxcd/flux2/action (finally)

### LINKED ISSUE #3509: The "install Flux" GitHub Action is failing

kingdonb:
Fix #3509 

I have a tester pointed at my local version to be extra sure, sorry there is one more error in that script which we did not see because we were not expecting the context as `set -e` (which I guess all github runners are, and that makes sense.)

Adding quotes here in the original version of this script, when the `TOKEN` variable is empty you get:

```
curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ''
```

(ref:
https://github.com/kingdonb/bootstrap-repo/actions/runs/3991169408/jobs/6845685491#step:3:31 )

That does curl twice, once for the URL provided and again for the empty string, which results in curl returning error code 3 "malformed URL" which results in the script aborting, again whenever the user fails to provide a `token`.

The script should mostly work with occasional rate limiting only when the token isn't present. See #3509 for more details.

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -33,7 +33,7 @@ runs:
         fi
 
         if [ -z "${VERSION}" ]; then
-          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL "${TOKEN[@]}" | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
+          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ${TOKEN[@]} | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
         fi
 
         BIN_URL="https://github.com/fluxcd/flux2/releases/download/v${VERSION}/flux_${VERSION}_linux_${ARCH}.tar.gz"


sjortiz:
our ci/cd process was affected by this, is there a version with can point too that is still functioning ? we'd love to point this to an specific version as we were working of "master" while you get this all clear out

kingdonb:
@sjortiz You can pin your action to any version at all, actually, and that will resolve it temporarily, or just add the github token as described in #3509.

```yaml
      - name: Setup Flux CLI
        uses: fluxcd/flux2/action@v0.38.3
```

This would be the most current tagged version as of right now. 🙏

kingdonb:
Looks like we'll be merging this quickly so you won't have to do anything.

kingdonb:
Maybe seems that our e2e CI itself is also a bit affected by the rate limiting issue! 🤫 

If it's a required check then we can't merge without it, unfortunately, unless there's an admin override. There are various workarounds posted in case this can't be merged right away. Very sorry for the trouble this has caused everyone.

stefanprodan:
@kingdonb this is still broken https://github.com/fluxcd/terraform-provider-flux/actions/runs/3997090660/jobs/6857959801

kingdonb:
@stefanprodan This is some bash issue must be from right here:

```
if [ -z "${VERSION}" ]; then
    VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ${TOKEN[@]} | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
  fi
```

In the last version we quoted `"${TOKEN[@]}"` which fails because it passes an empty string to curl when the tokens array does not contain any -H or headers to pass along with it. This version apparently fails when the token is passed properly.

I'm taking a look right now to see what we can do instead. I was trying to avoid writing a branch or doing separate assignments of the VERSION variable because that would mean breaking up this stream of pipes, and I'm leery to touch it again now given that I've made several ineffective fixes in a row. 😞

kingdonb:
So this is now only broken for folks that have followed the advice and added the token.

https://github.com/fluxcd/flux2/pull/3511#issuecomment-1401134126
(This comment shows how to pin the previous version, for anyone that is now only seeking to avoid further disruption before this is resolved.)

action/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
         fi
 
         if [ -z "${VERSION}" ]; then
-          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL "${TOKEN[@]}" | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
+          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ${TOKEN[@]} | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
```
seh, line 36: Why do we write `TOKEN[@]` instead of just `TOKEN` here? The "TOKEN" variable is not an array.

kingdonb: `TOKEN=(-H "Authorization: token ${TOKEN}")` is this not an array definition in bash?

kingdonb: The failure seems to be related to `set -o pipefail` which is another option I didn't realize is in effect.

(Any idea what "exit code 7" from `cut` means?)

kingdonb: Sorry, the output was misleading, that is not an exit code from `cut -d 2-` which removes the `v` from the front of the version number. It's error code 7 from `curl`.

```
curl -v https://api.github.com/repos/fluxcd/flux2/releases/latest -sL -H "Authorization: token gho_abcdSECRET"
```

This is the structure of the command we need to run, which is why the construction `TOKEN=(-H "Authorization: token ${TOKEN}")` was used in the first place I guess, because it lets us pass those around together.

```
curl -v https://api.github.com/repos/fluxcd/flux2/releases/latest -sL -H Authorization: token gho_abcdSECRET
```
This will fail because the authorization token header string is not all considered an argument to `-H`.

I think the safest thing to do now is to just have two branches, since it's clear enough to me what's failing and clear ho to fix, and I won't be happy with the branch conditional, but I'm open to suggestions if anyone wants to suggest a cleaner refactor.

I will open up another PR and we can continue the discussion there.

kingdonb: * https://github.com/fluxcd/flux2/pull/3513

I believe this fixes it. Either pass a quoted parameter pair, if there is one – or pass no parameter. (Do not pass an empty string, or unquoted parameter.)

If there is a better structure for doing that pattern in bash, I'm all ears.

PR #3513, opened 2023-01-24
Fix fluxcd/flux2/action

### LINKED ISSUE #3509: The "install Flux" GitHub Action is failing

kingdonb:
Add an "else" branch

(https://github.com/fluxcd/flux2/pull/3511 failed to fix #3509)

I'll be testing this version both with the TOKEN set and without TOKEN

I added a note in comment for anyone that encounters the rate limiting behavior, looking for a back-reference to the discussion on how it's solved and why this is necessary. Maybe this is not necessary here?

I'm going to test a few different versions because the first version I came up with here is objectively terrible, we should be able to set the output of an IF statement to a variable, or something slightly more elegant than what I've done here, (which is just copy the whole chain of pipes and repeat it twice.)

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -28,13 +28,17 @@ runs:
         VERSION=${{ inputs.version }}
         TOKEN=${{ inputs.token }}
 
-        if [ -n "${TOKEN}" ]; then
-          TOKEN=(-H "Authorization: token ${TOKEN}")
+        if [ -z "${VERSION}" ]; then
+          if [ -n "${TOKEN}" ]; then
+            VERSION_SLUG=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest --silent --location --header "Authorization: token ${TOKEN}" | grep tag_name)
+          else
+            # With no GITHUB_TOKEN you will experience occasional failures due to rate limiting
+            # Ref: https://github.com/fluxcd/flux2/issues/3509#issuecomment-1400820992
+            VERSION_SLUG=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest --silent --location | grep tag_name)
+          fi
         fi
 
-        if [ -z "${VERSION}" ]; then
-          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ${TOKEN[@]} | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
-        fi
+        VERSION=$(echo "${VERSION_SLUG}" | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
 
         BIN_URL="https://github.com/fluxcd/flux2/releases/download/v${VERSION}/flux_${VERSION}_linux_${ARCH}.tar.gz"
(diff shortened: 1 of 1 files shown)


kingdonb:
This (current) version does work, has been tested:
* with token: https://github.com/kingdonb/bootstrap-repo/actions/runs/3998188042
* without token: https://github.com/kingdonb/bootstrap-repo/actions/runs/3998195092

(it's in [ab76697](https://github.com/fluxcd/flux2/pull/3513/commits/ab76697bd8e392868277f260880325f048759515))

I'm going to iterate on it until it looks nicer structurally, (but FWIW there is a version that would close this issue now.)

kingdonb:
The version in 07b7ae3 is also tested:

* without token: https://github.com/kingdonb/bootstrap-repo/actions/runs/3998259236
* and with token: https://github.com/kingdonb/bootstrap-repo/actions/runs/3998267349

Sorry to make this so difficult. I'm happy to take feedback on ways we could structure this code better before I mark this as ready for review. I would appreciate anyone's opinion for example on whether:

* ab76697
* 07b7ae3

either of these versions is really better than the other one, (or if there's something even better we could do instead.)

kingdonb:
I am OK to have this merged, if there is a CI pass and if a reviewer can sign off on it.

kingdonb:
If someone can, please re-run the tests

(I think this is a persistent failure that we're already familiar with from elsewhere.)

action/action.yml
```diff
@@ -28,14 +28,19 @@ runs:
...
 
-        if [ -n "${TOKEN}" ]; then
-          TOKEN=(-H "Authorization: token ${TOKEN}")
-        fi
-
         if [ -z "${VERSION}" ]; then
-          VERSION=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL ${TOKEN[@]} | grep tag_name | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
+          if [ -n "${TOKEN}" ]; then
+            TOKEN=(-H "Authorization: token ${TOKEN}")
+            VERSION_SLUG=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest -sL "${TOKEN[@]}" | grep tag_name)
```
seh, line 34: Are you comfortable eliminating this intermediate array?
```suggestion
            VERSION_SLUG=$(curl https://api.github.com/repos/fluxcd/flux2/releases/latest --silent --location --header "Authorization: token ${TOKEN}" | grep tag_name)
```
Also, consider hoisting out the common _grep tag_name_ invocation from lines 34 and 38, moving it down into the pipeline at line 42.

seh: I'd find this clearer to read if the URL was the last command-line argument passed to _curl_, with the other flags preceding it.

kingdonb:
I re-tested both scenarios at b82c9be and this has been rebased and signed/squashed again.

action/action.yml
kingdonb: Fixed in 86e9d40

stefanprodan:
@kingdonb please squash all commits into one

stefanprodan (approved):

kingdonb:
Thanks, LGTM now 👍

PR #3517, opened 2023-01-26
Fix broken GitHub Action and handle case where VERSION is provided as an input

jooooel:
A bug was introduced in https://github.com/fluxcd/flux2/commit/ed13067ff2ecbc427f83b969503ad676060448d1, when trying to solve this issue: https://github.com/fluxcd/flux2/issues/3509.

If `VERSION` is provided, `VERSION_SLUG` will never be set and `VERSION` will be overwritten with a bad value.


jooooel:
I realized my commit is marked as "unverified". Thought I had signed it, will update it...

stefanprodan (approved):
LGTM

Thanks @jooooel 🏅

kingdonb:
Thank you @jooooel you beat me to it, I think I'll hang up my Bash shoes for a while. Congrats first-time contributor here!

PR #3519, opened 2023-01-26
Fix GitHub Action

kingdonb:
There is one more issue, I had not tested with the `version:` parameter, and there is a logic inversion.

If `VERSION` is passed in, then we should not overwrite it.


kingdonb:
It is! Closing in favor of that one.

Issue #2409, opened 2022-02-09
Add more scopes to the default k8s ClusterRoles


kingdonb:
If the kubernetes user has viewer access broadly to the cluster, should they also have access to Flux resources?

An example was provided in the discussions from cert-manager who has solved a similar issue.

### Discussed in https://github.com/fluxcd/flux2/discussions/2358

<div type='discussions-op-text'>

<sup>Originally posted by **ilya-git** January 31, 2022</sup>
By default k8s has some ClusterRoles, like e.g. "view" that allows access read to all resources. Right now it does not allow any access to the flux resources as they have to be explicitly added. I am wondering if it's a good idea to add these aggregated roles?

This is how cert-manager does it e.g. (https://app-nl.logz.io/#/dashboard/kibana/doc/logzioCustomerIndex*/logzioCustomerIndex220131_v1.account-395769/doc/?_g=()&id=iI5yr34B-Tg5N-7JdnWO):

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ template "cert-manager.fullname" . }}-view
  labels:
    app: {{ template "cert-manager.name" . }}
    chart: {{ template "cert-manager.chart" . }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
    rbac.authorization.k8s.io/aggregate-to-view: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
rules:
  - apiGroups: ["certmanager.k8s.io"]
    resources: ["certificates", "issuers"]
    verbs: ["get", "list", "watch"]
```

This is how I did it now for myself:

```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: flux-view
  labels:    
    rbac.authorization.k8s.io/aggregate-to-view: "true"
rules:
- apiGroups:
  - notification.toolkit.fluxcd.io
  - source.toolkit.fluxcd.io
  - helm.toolkit.fluxcd.io
  - image.toolkit.fluxcd.io
  - kustomize.toolkit.fluxcd.io
  resources: ["*"]
  verbs: ["get", "list", "watch"]

```</div>

PR #3566, opened 2023-02-07
rbac: Add view and edit aggregated cluster roles

stefanprodan:
This PR adds two cluster roles which grant the Kubernetes `view`, `edit` and `admin` roles access to Flux custom resources. 

Fix: https://github.com/fluxcd/flux2/issues/2409
Fix: #3572
Ref: https://github.com/fluxcd/kustomize-controller/issues/762

### DIFF
--- a/manifests/rbac/edit.yaml
+++ b/manifests/rbac/edit.yaml
@@ -1,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flux-edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+  - apiGroups:
+      - notification.toolkit.fluxcd.io
+      - source.toolkit.fluxcd.io
+      - helm.toolkit.fluxcd.io
+      - image.toolkit.fluxcd.io
+      - kustomize.toolkit.fluxcd.io
+    resources: ["*"]
+    verbs:
+      - create
+      - delete
+      - deletecollection
+      - patch
+      - update
--- a/manifests/bases/helm-controller/kustomization.yaml
+++ b/manifests/bases/helm-controller/kustomization.yaml
@@ -6,7 +6,7 @@ resources:
 - account.yaml
 transformers:
 - labels.yaml
-patchesJson6902:
+patches:
 - target:
     group: apps
     version: v1
--- a/manifests/bases/image-automation-controller/kustomization.yaml
+++ b/manifests/bases/image-automation-controller/kustomization.yaml
@@ -6,7 +6,7 @@ resources:
 - account.yaml
 transformers:
 - labels.yaml
-patchesJson6902:
+patches:
 - target:
     group: apps
     version: v1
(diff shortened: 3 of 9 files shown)


manifests/rbac/edit.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flux-edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```
seh, line 7: Is this still necessary? [Back in August 2018 I learned](https://kubernetes.slack.com/archives/C0EN96KUY/p1533841633000133) from @deads2k that "view" aggregation is supposed to be included by "edit," and "edit" is supposed to be include by "admin" automatically.

stefanprodan: cert-manager and other controllers have these set. I guess it doesn't hurt to make this explicit so people reading the Flux RBAC understand what's going on.

seh (commented):
Is the idea with the "edit" and "admin" mutative rights that a user who has those roles granted in a given namespace could edit the target objects that Flux would reconcile, annotating or labeling them to disable Flux's coverage?

I was thinking at first that an administrator might want to retain full control over the Flux-related objects in a given namespace, in order to ensure that they continue projecting some required set of objects. However, I realized that other users with the "edit" or "admin" roles granted could still fight back against Flux, since [it honors the "kustomize.toolkit.fluxcd.io/reconcile" annotation and label](https://fluxcd.io/flux/components/kustomize/kustomization/#reconciliation).

stefanprodan:
@seh the `admin/edit` are for Flux impersonation, instead of creating a role binding to cluster-admin, the tenant service account could bind to `admin` and still be able to reconcile HelmRelease, Alerts, etc. The actual tenants shouldn't have edit rights to the cluster, all changes should be made in Git and Flux should apply them. I would grand tenants the `view` role only.

seh:
> The actual tenants shouldn't have edit rights to the cluster, all changes should be made in Git and Flux should apply them. I would gran[t] tenants the `view` role only.

I understand that intention. It so happens that in my organization we have a variety of tenants, some of which are engineers maintaining the applications that run within a given namespace. We have to allow them fairly liberal privileges to inspect and manipulate things, even if they like to use Flux as a convenient way to deploy similar changes to many Kubernetes clusters easily. In other words, in this model, Flux serves as assistive automation, as opposed to enforcement of policy.

stefanprodan:
> We have to allow them fairly liberal privileges to inspect and manipulate things, even if they like to use Flux as a convenient way to deploy similar changes to many Kubernetes clusters easily. In other words, in this model, Flux serves as assistive automation, as opposed to enforcement of policy.

Is this RBAC change prevents you from doing that?

seh:
> Does this RBAC change prevents you from doing that?

It makes it more likely that we'd accidentally grant broader privileges than we had intended, but if we _really_ need to prevent that, we can delete these _ClusterRoles_ with a _kustomize_ patch when reconciling Flux's manifests, or patch them to remove the "aggregate-to-..." annotations.

seh:
Thank you for [this part of the documentation](https://fluxcd.io/flux/cheatsheets/bootstrap/#disable-kubernetes-cluster-role-aggregations), @stefanprodan.

PR #3575, opened 2023-02-10
RFC-0003: Introduce Flux OCI media type

stefanprodan:
Given that DockerHub and all the other popular registries now support custom media type, I propose that we use `application/vnd.cncf.flux` for the artifacts created with `flux push`. This change doesn't affect Flux source-controller, the controller will still work with any type of OCI artifact as long as it contains a `tar+gzip` layer.

Instead of using the Docker distribution media types, we'll be using:
- artifact media type `application/vnd.oci.image.manifest.v1+json`
- config media type `application/vnd.cncf.flux.config.v1+json`
- content media type `application/vnd.cncf.flux.content.v1.tar+gzip`

### DIFF
--- a/rfcs/0003-kubernetes-oci/README.md
+++ b/rfcs/0003-kubernetes-oci/README.md
@@ -4,7 +4,7 @@
 
 **Creation date:** 2022-03-31
 
-**Last update:** 2022-09-29
+**Last update:** 2023-02-10
 
 ## Summary
 
@@ -35,7 +35,7 @@ they do today for container images.
 
 ### Non-Goals
 
-- Introduce a new OCI media type for artifacts containing Kubernetes manifests.
+- Enforce a specific OCI media type for artifacts containing Kubernetes manifests or any other configs.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


rfcs/0003-kubernetes-oci/README.md
```diff
@@ -442,7 +435,7 @@ status:
     checksum: d7e924b4882e55b97627355c7b3d2e711e9b54303afa2f50c25377f4df66a83b
     lastUpdateTime: "2022-06-22T09:14:21Z"
     metadata:
-      org.opencontainers.image.created: "2022-08-08T12:31:41+03:00"
+      org.opencontainers.image.created: "2022-08-08T12:31:41+00:00"
       org.opencontainers.image.revision: 6.1.8/b3b00fe35424a45d373bf4c7214178bc36fd7872
```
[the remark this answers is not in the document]
stefanprodan: I've changed the time to UTC, will update here all examples as well.

rfcs/0003-kubernetes-oci/README.md
```diff
@@ -47,25 +47,26 @@ and push the archive to a container registry as an OCI artifact.
...
+and the layer media type is set to `application/vnd.cncf.flux.content.v1.tar+gzip`.
 
-The source URL and revision are added to the OCI artifact as annotations in the format:
+The source and revision are added to the OCI artifact as Open Containers standard annotations:
 
 ```json
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
```
souleb, line 65: Is it new? It feels to me like this is already implemented.

If not, these changes (`manifest mediaType` and `annotations` keys) are quite important for users relying on it. It should at least be in a different commit.

stefanprodan: The annotations are not new, the media type changed to OCI in this proposal.

souleb (approved):
LGTM!

PR #3587, opened 2023-02-13
Support shortening of revision with digest

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

PR #3592, opened 2023-02-15
tests: only provide template values when used

hiddeco:
As otherwise the `.golden` values can not be automatically updated using
`-update` as documented in `CONTRIBUTING.md`.

Also ensure we do not use `defer` but rather `t.Cleanup` in tests, as
this will always be called even if e.g. `t.Fatal` abruptly stops the
test.

Also fixed some other minor things I ran into, and filled issues for the
things that are beyond reach of time I have available.

### DIFF
--- a/cmd/flux/build_kustomization_test.go
+++ b/cmd/flux/build_kustomization_test.go
@@ -171,8 +171,7 @@ spec:
 	if err != nil {
 		t.Fatal(err)
 	}
-
-	defer os.Remove("./testdata/build-kustomization/podinfo.yaml")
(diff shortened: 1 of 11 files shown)


cmd/flux/build_kustomization_test.go
```diff
@@ -171,8 +171,7 @@ spec:
 	if err != nil {
 		t.Fatal(err)
 	}
-
```
hiddeco, line 174: There is more here that could be done, as using `./testdata/` for something temporarily seems wrong.

I think we should instead use `t.TempDir()` as a target for the file that is been written. Which would prevent having to think about garbage collecting it later on.

This is however beyond the time I have available now, as I was merely trying to fix things I ran into while working on #3587.

PR #3654, opened 2023-03-03
Implement `--deploy-token-auth` in GitLab bootstrapping

timofurrer:
This change set implements support for the `--deploy-token-auth` option in the `flux bootstrap gitlab` command.

That option will reconcile a GitLab Project Deploy Token to use for the authentication of the GitLab git repository.
A GitLab Project Deploy Token can be used the same way as a Personal Access Token which is already supported via `--token-auth`. The difference with the GitLab Project Deploy Token is that the token is managed (created, updated, deleted) by Flux and not provided by the user.

This change is transparent to the source-controller.

A prerequisite for this change is the
`fluxcd/go-git-providers` change here:

* https://github.com/fluxcd/go-git-providers/pull/191

See related discussion here: https://github.com/fluxcd/flux2/discussions/3595
GitLab Issue here: https://gitlab.com/gitlab-org/gitlab/-/issues/392605

### DIFF
--- a/pkg/bootstrap/bootstrap_provider.go
+++ b/pkg/bootstrap/bootstrap_provider.go
@@ -59,6 +59,8 @@ type GitProviderBootstrapper struct {
 
 	sshHostname string
 
+	useDeployTokenAuth bool
+
 	provider gitprovider.Client
 }
 
@@ -184,6 +186,16 @@ func (o reconcileOption) applyGitProvider(b *GitProviderBootstrapper) {
 	b.reconcile = true
 }
 
+func WithDeployTokenAuth() GitProviderOption {
+	return deployTokenAuthOption(true)
+}
+
+type deployTokenAuthOption bool
+
+func (o deployTokenAuthOption) applyGitProvider(b *GitProviderBootstrapper) {
+	b.useDeployTokenAuth = true
+}
+
 func (b *GitProviderBootstrapper) ReconcileSyncConfig(ctx context.Context, options sync.Options) error {
 	if b.repository == nil {
 		return errors.New("repository is required")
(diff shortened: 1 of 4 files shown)


timofurrer:
@makkes I think that we can proceed here as well now that https://github.com/fluxcd/go-git-providers/pull/191 has been merged. WDYT?

For the merge, would we require a provider `go-git-providers` release or not?

pkg/bootstrap/bootstrap_provider.go
```diff
@@ -554,6 +612,28 @@ func deployKeyName(namespace, secretName, branch, path string) string {
...
+		if v == "" {
+			continue
+		}
+		if name == "" {
+			name = v
+		} else {
+			name = name + "-" + v
+		}
+	}
+	return name
```
makkes, line 634: How about this instead which is easier to read IMHO:

```suggestion
	var elems []string
	for _, v := range []string{namespace, secretName, branch, path} {
		if v == "" {
			continue
		}
		elems = append(elems, v)
	}
	return strings.Join(elems, "-")
```

makkes (commented):
Looks good to me; ran it locally and works fine. Just 2 minor comments.

timofurrer:
@hiddeco where can I learn more about the timeline of this release, if there is any? When could I expect to have it in ?

hiddeco:
@timofurrer I expected this to be sometime next week. We need to wrap up some other bigger changes like promoting the `GitRepository` API to v1, which will be part of this release as well.

timofurrer:
@hiddeco sounds great! Thanks for the info.

makkes (approved):

Issue #3656, opened 2023-03-04
There is no `flux reconcile source chart`


kingdonb:
When a configuration like this one:

https://github.com/kingdon-ci/fleet-infra/blob/663937f2414a670331c363766cd8df501400db7c/clusters/test/limnocentral/wasm-test/stage/taking-bart-helmrelease.yaml#L12-L15

is linked to a `HelmRepository` like this one:

https://github.com/kingdon-ci/fleet-infra/blob/663937f2414a670331c363766cd8df501400db7c/clusters/test/limnocentral/wasm-test/stage/taking-bart-helmrepo.yaml#L13-L21

and a new chart version is published, it will not be picked up until the `HelmChart` is reconciled. On what interval that gets reconciled, I think you can configure in `spec.chart.spec.interval` but there is no `flux reconcile source chart` to reconcile this on-demand.

I'm not sure that's what is needed exactly (maybe I should set up webhooks, and point the right event at my `HelmChart` with an `Alert` – who wants to reconcile things by hand, or even tweak some random interval for that matter?)

But it seems like a logistical gap that there is no way in the CLI to reconcile the chart by hand, if you may sometimes need to do it.

stefanprodan:
Yes this is needed for OCI as reconciling the Helm repo does nothing.

PR #3660, opened 2023-03-06
feat: reconcile referencing helmchart if helmrelease is reconciled 

raffis:
## Current situation
`flux reconcile hr my-release --with-source` does not update the HelmChart (installed from an oci registry) if the version is a semver expression (like in this example `*`) and a new version of the chart was released meanwhile.
Meaning at the moment one has either to await the configured interval or manually annotate the referenced HelmChart in order to receive the new version.
I would expect that if I manually trigger a reconcile I get the latest version possible which is matched by semver which is not the case currently.

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: my-release
spec:
  chart:
    spec:
      chart: mychart
      interval: 15m0s
      sourceRef:
        kind: HelmRepository
        name: oci-repository
        namespace: flux-system
      version: '*'
  interval: 15m0s
```
## Proposal
`flux reconcile hr my-release --with-source` should actually annotate the referenced helm chart as well.

This pr implements just that:
```
go run . reconcile hr my-release --with-source
► annotating HelmRepository oci-repository in default namespace
✔ HelmRepository annotated
◎ waiting for HelmRepository reconciliation
✔ Helm repository is ready
► annotating HelmRelease my-release in default namespace
✔ HelmRelease annotated
◎ waiting for HelmRelease reconciliation
✔ applied revision 4.0.11+1
► annotating HelmChart default-my-release in flux-system namespace
✔ HelmChart annotated
◎ waiting for HelmChart reconciliation
✔ fetched revision 4.0.12+1
```


stefanprodan:
IMO the chart should be reconciled before the HelmRelease as with all other sources.

raffis:
> IMO the chart should be reconciled before the HelmRelease as with all other sources. In the case a new chart is found, the HelmRelease gets reconciled automatically so now we reconcile it twice which is wrong... cc @hiddeco

Makes sense, will change the order 👍🏻

stefanprodan:
@raffis I suggest you first implement #3656, then call the function in there for the chart.

raffis:
> @raffis I suggest you first implement #3656, then call the reconcileChart function in this PR.

superseded by #3675

Issue #3559, opened 2023-02-03
Update Go to v1.20


stefanprodan:
Update Go to 1.20 in GH Workflows and Dockerfiles for:
- [x] pkg
- [x] source-controller
- [x] source-watcher
- [x] kustomize-controller
- [x] notification-controller
- [x] helm-controller
- [x] image-automation-controller 
- [x] image-reflector-controller
- [x] flux2
- [x] terraform-provider-flux

hiddeco:
Think this is actually scheduled for next MINOR.

PR #3670, opened 2023-03-08
Update Go to 1.20

hiddeco:
Part of https://github.com/fluxcd/flux2/issues/3559

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,7 +67,7 @@ for source changes.
 
 Prerequisites:
 
-* go >= 1.19
+* go >= 1.20
 * kubectl >= 1.20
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -67,7 +67,7 @@ for source changes.
 
 Prerequisites:
 
-* go >= 1.19
```
hiddeco, line 70: Actually think this could just match the version defined in `go.mod`, instead of keeping up with latest.

hiddeco: Problem this creates is of course the `-compat=$ver` which we do in the `Makefile`.

stefanprodan (approved):

PR #3673, opened 2023-03-09
ci: Use latest available images of kindest/node

stefanprodan:


hiddeco (approved):

hiddeco:
Maybe add "latest" to matrix?

stefanprodan:
We have latest in the AMD e2e test. I would leave this here pinned if you agree.

PR #3721, opened 2023-03-23
fix: Avoid printing an extra newline when exporting resources

### LINKED ISSUE #3593: `--export` results at times in double newline at EOF
While looking into #3592, I noticed some of our tests around `--export` functionalities have files which end with a double newline.

The culprit of this seems to be the usage of `rootCmd.Println` to print the rendered YAML, which by itself adds a newline (even if the thing it prints already has this at the end).

To provide a better UX, we should ensure we only print a single one at the end of the file which also makes

```console
$ flux ... --export >> file.yaml
$ flux ... --export >> file.yaml
```

produce prettier files.

yiannistri:
Closes #3593 

Following Hidde's hint, the change in this PR should prevent extra newlines from being printed. This affects both the `flux export ..` command and the `--export` flag used in other Flux commands.

Also tested with
```
flux export .. >> export.yaml
flux export .. >> export.yaml
```
and it no longer produces a newline between resources in the same file.

### DIFF
--- a/cmd/flux/export.go
+++ b/cmd/flux/export.go
@@ -122,5 +122,6 @@ func printExport(export interface{}) error {
 func resourceToString(data []byte) string {
 	data = bytes.Replace(data, []byte("  creationTimestamp: null\n"), []byte(""), 1)
 	data = bytes.Replace(data, []byte("status: {}\n"), []byte(""), 1)
+	data = bytes.TrimSpace(data)
 	return string(data)
 }
--- a/cmd/flux/testdata/create_source_git/export.golden
+++ b/cmd/flux/testdata/create_source_git/export.golden
@@ -12,4 +12,3 @@ spec:
   ref:
     branch: master
   url: https://github.com/stefanprodan/podinfo
-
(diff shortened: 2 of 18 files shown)


cmd/flux/export.go
```diff
@@ -115,7 +115,7 @@ func printExport(export interface{}) error {
 		return err
 	}
 	rootCmd.Println("---")
-	rootCmd.Println(resourceToString(data))
+	rootCmd.Print(resourceToString(data))
```
hiddeco, line 118: Seeing this together with `resourceToString`, it might actually be better to add a `data = bytes.TrimSpace(data)` to the function below and keep this in place.

This would simply ensure that even if the generating function makes a mistake (by either forgetting to add a newline, or adding to many). We always return just a single newline at the end.

hiddeco (commented):
Thanks a lot for picking this up :bow: 

Small suggestion based on second thoughts...

cmd/flux/export.go
yiannistri: Great suggestion 👍

hiddeco (approved):
Thank you @yiannistri :1st_place_medal:

PR #3764, opened 2023-04-06
Add label selector flag to get cmd

somtochiama:
This pull request adds a label selector flag to `get` command to enable users filter objects by label
e.g

```
flux get sources git -l sharding.fluxcd.io/key=shard1
```

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -66,6 +66,7 @@ type GetFlags struct {
 	allNamespaces  bool
 	noHeader       bool
 	statusSelector string
+	labelSelector  string
 	watch          bool
 }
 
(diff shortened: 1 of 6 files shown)


cmd/flux/get.go
```diff
@@ -150,6 +153,21 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
 		listOpts = append(listOpts, client.MatchingFields{"metadata.name": args[0]})
 	}
 
+	if getArgs.labelSelector != "" {
+		label, err := metav1.ParseToLabelSelector(getArgs.labelSelector)
+		if err != nil {
+			return err
```
makkes, line 159: I personally prefer wrapping errors and provide some context before returning them. But that may be personal preference.

makkes (approved):
Looks great.

cmd/flux/get.go
stefanprodan: Yes let's return here something like `invalid label expression, error: %w`

cmd/flux/get.go
```diff
@@ -156,7 +156,7 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
 	if getArgs.labelSelector != "" {
 		label, err := metav1.ParseToLabelSelector(getArgs.labelSelector)
 		if err != nil {
-			return err
+			return fmt.Errorf("unable to parse label selector: error: %w", err)
```
makkes, line 159:
```suggestion
			return fmt.Errorf("unable to parse label selector: %w", err)
```

stefanprodan (approved):
LGTM

Thanks @somtochiama this will help with sharding 🥇

PR #3797, opened 2023-04-13
better messaging for `pull artifact` command

makkes:
- When there's an error stat'ing the output directory flux now prints
  the error:

  Before:
  ```
  ✗ invalid output path ./ro-dir/foo
  ```

  After:
  ```
  ✗ invalid output path "./ro-dir/foo": stat ./ro-dir/foo: permission denied
  ```
- When no output directory is provided flux now explicitly says so in
  the error:

  Before:
  ```
  ✗ invalid output path
  ```

  After:
  ```
  ✗ output path cannot be empty
  ```


stefanprodan (approved):
LGTM

Thanks @makkes

PR #3822, opened 2023-04-20
Outdated Loki Helm values URL

Hey:
The location of this URL was moved: https://github.com/grafana/helm-charts/blob/main/charts/loki/README.md


makkes (approved):
Thanks @Hey

Issue #3840, opened 2023-05-02
Auto-completion for `flux reconcile image repository` shows ImagePolicies


brovoca:
### Describe the bug

The tab-auto-completion for `flux reconcile image repository` shows ImagePolicies

### Steps to reproduce

1. Install Flux2 CLI
2. Install completion: `source <(flux completion zsh) && complete -F _flux f`
3. Type `flux reconcile image repository` and press tab

### Expected behavior

ImageRepositories should be listed

### OS / Distro

N/A

### Flux version

2.0.0-rc.1

### Flux check

► checking prerequisites
✔ Kubernetes 1.24.6 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.32.1
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.32.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.27.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v1.0.0-rc.1
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v1.0.0-rc.1
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v1.0.0-rc.1
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta2
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1
✔ helmcharts.source.toolkit.fluxcd.io/v1beta2
✔ helmreleases.helm.toolkit.fluxcd.io/v2beta1
✔ helmrepositories.source.toolkit.fluxcd.io/v1beta2
✔ imagepolicies.image.toolkit.fluxcd.io/v1beta2
✔ imagerepositories.image.toolkit.fluxcd.io/v1beta2
✔ imageupdateautomations.image.toolkit.fluxcd.io/v1beta1
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1
✔ ocirepositories.source.toolkit.fluxcd.io/v1beta2
✔ providers.notification.toolkit.fluxcd.io/v1beta2
✔ receivers.notification.toolkit.fluxcd.io/v1
✔ all checks passed

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

[...]

PR #3857, opened 2023-05-08
Fix autocompletion for image repository reconcile

talife:
Fix: #3840 

Should we start using a completion test routine ?


stefanprodan:
Hey @talife can you please signoff your commit? Instructions here https://github.com/fluxcd/flux2/pull/3857/checks?check_run_id=13311660095

stefanprodan:
> Should we start using a completion test routine ?

Yes that would be great, please create an issue for this. Thanks!

PR #3866, opened 2023-05-10
Update Alpine to 3.18

onedr0p:
This change would be nice to get in because MUSL finally implemented TCP fallback in their DNS resolver.

alpinelinux.org/posts/Alpine-3.18.0-released.html


hiddeco:
DCO sign-off is required.

hiddeco:
> Whoops.. Should I also do that on the other PRs I opened? Those looked happy at least.

I went and checked, and at least one of them had a sign-off.

hiddeco:
@onedr0p if you feel adventurous, there are some other places where this update is required: https://github.com/search?q=org%3Afluxcd+alpine+language%3ADockerfile&type=code&l=Dockerfile

onedr0p:
Sure I can take a stab at those later today.

Issue #3714, opened 2023-03-21
"flux push" to OCI (AWS ERC) fails with "TOOMANYREQUESTS: Rate exceeded"


xaka:
### Describe the bug

 While running multiple "flux push" commands in parallel against the same AWS ECR registry, but different repos, it ends up crashing with "TOOMANYREQUESTS: Rate exceeded". The expectation is flux would re-try instead and keep going until it succeeds.

### Steps to reproduce

N/A

### Expected behavior

Flux continues to re-try until operation succeeds instead of crashing

### OS / Distro

N/A

### Flux version

v0.40.2

### Flux check

N/A

### Container Registry provider

AWS ECR

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
We could add retries with back-off until the specified `--timeout` expires.

PR #3918, opened 2023-05-25
Retry oci push operations

somtochiama:
Closes: #3714
Depends on: https://github.com/fluxcd/pkg/pull/563
 for retrying on transient errors on `flux push artifact` command.


cmd/flux/push_artifact.go
```diff
@@ -125,6 +136,8 @@ func init() {
 	pushArtifactCmd.Flags().StringArrayVarP(&pushArtifactArgs.annotations, "annotations", "a", nil, "Set custom OCI annotations in the format '<key>=<value>'")
 	pushArtifactCmd.Flags().StringVarP(&pushArtifactArgs.output, "output", "o", "",
 		"the format in which the artifact digest should be printed, can be 'json' or 'yaml'")
+	pushArtifactCmd.Flags().DurationVarP(&pushArtifactArgs.retryTimeout, "retry-timeout", "", 0, "the timeout for retrying temporary errors when pushing to oci registry")
```
stefanprodan, line 139: We already have a global flag called timeout, can't we use that one?

stefanprodan (approved):
LGTM

Thanks @somtochiama 🏅

PR #3924, opened 2023-05-26
Fix break lines in create helmrelease and source

hgranillo:
Hello flux maintainers!

This PR fixes a few missing break lines in the create helmrelease and create source help documentation that could lead to some unintended executions if blindly copy-pasted 

For example, I've added `--export` to the `create source helm` OCI example included in the `flux create source helm --help` command:

Discovered this while I was pasting examples in my terminal with the `--export` flag to check if I was using the newest API versions for my flux objects

```
❯   flux create source helm podinfo \
    --url=oci://ghcr.io/stefanprodan/charts/podinfo
    --username=username \
    --password=password --export
✚ generating HelmRepository source
► applying HelmRepository source
✔ source updated
◎ waiting for HelmRepository source reconciliation
^C
```

### DIFF
--- a/cmd/flux/create_helmrelease.go
+++ b/cmd/flux/create_helmrelease.go
@@ -83,9 +83,9 @@ var createHelmReleaseCmd = &cobra.Command{
 
   # Create a HelmRelease with a custom release name
   flux create hr podinfo \
-    --release-name=podinfo-dev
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @hgranillo

Issue #3931, opened 2023-05-30
Upgrade to rc4 with flux bootstrap outputs `log.SetLogger(...) was never called`


plockaby:
### Describe the bug

When I run `flux bootstrap` to upgrade to rc4 I'm getting an error from `controller-runtime`. The upgrade seems to be successful, regardless.

My upgrade command:
```
GITHUB_TOKEN= flux bootstrap github \
    --owner=XXXX \
    --repository=fluxcd \
    --path=./clusters/XXXX \
    --private \
    --read-write-key \
    --token-auth \
    --components=source-controller,kustomize-controller,helm-controller,notification-controller \
    --components-extra=image-reflector-controller,image-automation-controller \
    --interval=5m0s \
    --context=XXXX
```

```
► connecting to github.com
► cloning branch "main" from Git repository "https://github.com/XXXX/fluxcd.git"
✔ cloned repository
► generating component manifests
# Warning: 'patchesJson6902' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
✔ generated component manifests
✔ committed sync manifests to "main" ("XXXX")
► pushing component manifests to "https://github.com/XXXX/fluxcd.git"
✔ reconciled components
► determining if source secret "flux-system/flux-system" exists
► generating source secret
► applying source secret "flux-system/flux-system"
✔ reconciled source secret
► generating sync manifests
✔ generated sync manifests
✔ sync manifests are up to date
► applying sync manifests
✔ reconciled sync configuration
◎ waiting for Kustomization "flux-system/flux-system" to be reconciled
✔ Kustomization reconciled successfully
[controller-runtime] log.SetLogger(...) was never called, logs will not be displayed:
goroutine 1 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.20.4/x64/src/runtime/debug/stack.go:24 +0x65
sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/log/log.go:59 +0xbd
sigs.k8s.io/controller-runtime/pkg/log.(*delegatingLogSink).WithName(0xc0003e5a40, {0x2474abf, 0x14})
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/log/deleg.go:147 +0x4c
[...]

PR #3932, opened 2023-05-30
Set `controller-runtime` logger to a null logger

aryan9600:
This is required because `controller-runtime` expects its consumers to set a logger through `log.SetLogger` within 30 seconds of the program's initalization. If not set, the entire debug stack is printed as an error. Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/ed8be90/pkg/log/log.go#L59 Since we have our own logging and don't care about `controller-runtime`'s logger, we configure it's logger to do nothing.

Fixes: #3931

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -24,12 +24,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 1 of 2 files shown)


cmd/flux/main.go
```diff
@@ -34,6 +35,7 @@ import (
 	runclient "github.com/fluxcd/pkg/runtime/client"
 
 	"github.com/fluxcd/flux2/v2/pkg/manifestgen/install"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
```
hiddeco, line 38: Should be in external dependency group above.

hiddeco (approved):

PR #3976, opened 2023-06-14
Use equivalent and shorter way to generate shell completions

darklore:
Replace homebrew script with equivalent and [more shorter method](https://rubydoc.brew.sh/Formula#generate_completions_from_executable-instance_method) to install shell completion script.


hiddeco (approved):
Nice one, thanks! :bow:

Issue #4014, opened 2023-06-26
Update Flux components to Go 1.20 in `go.mod` and `oss-fuzz`


stefanprodan:
To align the Go version with Kubernetes and controller-runtime, we need to bump Go to 1.20 in all APIs `go.mod`, all controllers `go.mod` and all packages `go.mod`. We also have to install Go 1.20 in all oss-fuzz Dockerfiles as the base image comes with 1.19.

TODO:
- [x] pkg
- [x] source-watcher
- [x] source-controller 
- [x] kustomize-controller 
- [x] notification-controller 
- [x] helm-controller
- [x] image-reflector-controller 
- [x] image-automation-controller
- [x] flux2
- [x] terraform-provider-flux

PR #4015, opened 2023-06-26
Align `go.mod` version with Kubernetes (Go 1.20)

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4014


makkes (commented):
What about https://github.com/fluxcd/flux2/blob/main/.github/aur/flux-go/PKGBUILD.template#L17 and https://github.com/fluxcd/flux2/blob/main/.github/aur/flux-scm/PKGBUILD.template#L15?

makkes (approved):

PR #4028, opened 2023-07-03
Annotate errors from go-git-providers

### LINKED ISSUE #3623: Improve messaging for errors coming from go-git-providers
When go-git-providers isn't able to fetch repository details while bootstrapping it's hard to understand where the error has originated. Here is an error caused by the repository not being reachable because of missing proxy settings:

```
flux bootstrap github --owner=bar --repository=baz --branch=master --hostname=foo --token-auth --private=true --verbose --log-level=debug
connecting to foo
failed to get Git repository "https://foo/bar/baz": Get "https://foo/api/v3/bar/baz": Forbidden
```

We should prefix errors coming from go-git-providers with `provider error: %w` so it's easier to debug the cause.

makkes:
closes #3623


stefanprodan (approved):
LGTM

Thanks @makkes

PR #4051, opened 2023-07-07
action: use `$RUNNER_TOOL_CACHE`, support MacOS and Windows, validate checksum

hiddeco:
Plus the verification of the SHA256 of the archive, as advertised in the checksum file published together with the release.

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -1,0 +1,29 @@
+name: test-gh-action
+
+on:
+  pull_request:
+    paths:
+      - 'action/**'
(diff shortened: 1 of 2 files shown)


.github/workflows/action.yaml
```diff
@@ -0,0 +1,29 @@
+name: action
```
stefanprodan, line 1:
```suggestion
name: test-gh-action
```

Maybe?

stefanprodan (approved):
LGTM

Thanks @hiddeco 🥇

Issue #4060, opened 2023-07-10
Flux Action Token required for Github Enterprise users


ebini:
### Describe the bug

There was an update in the actions.yaml

https://github.com/fluxcd/flux2/commit/6f94ec728f7faedf2fbe7865b227967a468f1bbd

We are using Github Enterprise. With our Enterprise Runner Token we cannot access the Standard Github.
So we need the token, otherwise we cannot connect and now get  401 Unauthorized.

### Steps to reproduce

use actions pipeline on Github Enterprises
```
      - name: Setup Flux CLI
        uses: fluxcd/flux2/action@main
        with:
          token: ${{ secrets.MY_PERSONAL_TOKEN }}
```
### Expected behavior

Connect to Github with my personal Token

still working with
```
      - name: Setup Flux CLI
        uses: fluxcd/flux2/action@v2.0.0
        with:
          token: ${{ secrets.MY_PERSONAL_TOKEN }}
```

### OS / Distro

ubuntu (enterprise) runner

### Flux version

changed in main branch

### Flux check

N/A

### Git provider

GitHub Enterprise

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4061, opened 2023-07-10
action: re-allow configuration of non-default token

hiddeco:
To allow usage of action on GitHub Enterprise instances.

Fixes: #4060

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -16,9 +16,8 @@ inputs:
     description: "Alternative location for the Flux binary, defaults to path relative to $RUNNER_TOOL_CACHE."
     required: false
   token:
-    description: "GitHub Token used to authentication against the API (generally only needed to prevent quota limit errors)"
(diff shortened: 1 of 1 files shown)


action/action.yml
```diff
@@ -16,18 +16,23 @@ inputs:
     description: "Alternative location for the Flux binary, defaults to path relative to $RUNNER_TOOL_CACHE."
     required: false
   token:
-    description: "GitHub Token used to authentication against the API (generally only needed to prevent quota limit errors)"
+    description: "GitHub Token used to authentication against the API. Defaults to the token from the GitHub context for the workflow."
```
stefanprodan, line 19:
```suggestion
    description: "Token used to authentication against the GitHub.com API. Defaults to the token from the GitHub context of the workflow."
```
Maybe?

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #4076, opened 2023-07-16
image name has ":"

StillP:
If the name of image has ":", that will not get correct version. Now I try fix it.

### DIFF
--- a/cmd/flux/version.go
+++ b/cmd/flux/version.go
@@ -124,7 +124,8 @@ func splitImageStr(image string) (string, string, error) {
 		return "", "", fmt.Errorf("missing image tag in image %s", image)
 	}
 
-	name, tag := imageArr[0], imageArr[1]
+	// name, tag := imageArr[0], imageArr[1]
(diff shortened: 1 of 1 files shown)


makkes:
Colons are not allowed in OCI image names so I'm not sure what this PR is solving really.

hiddeco:
The problem here seems to be that it relies on the container image specification, which can contain e.g. `fictional.registry.example:10443/imagename:tag`. See: https://kubernetes.io/docs/concepts/containers/images/#image-names.

makkes:
> The problem here seems to be that it relies on the container image from the Deployment and/or Pod spec, which can contain e.g. `fictional.registry.example:10443/imagename:tag`.

Oh, I see. That makes sense.

stefanprodan:
We could use the image parse function instead https://github.com/fluxcd/pkg/blob/5b608cb95e2129ad9cf19ed6e26f87003518a70e/oci/client/tag.go#L29C10-L29C10

stefanprodan:
Supersedes by #4076

PR #4114, opened 2023-07-31
[RFC] Passwordless authentication for Git repositories

aryan9600:
This RFC proposes adding support for passwordless authentication for Git repositories on Git SaaS platforms and cloud providers using methods like OIDC, OAuth2 and IAM.

### DIFF
--- a/rfcs/0006-git-repo-passwordless-auth/README.md
+++ b/rfcs/0006-git-repo-passwordless-auth/README.md
@@ -1,0 +1,198 @@
+# RFC-0006 Passwordless authentication for Git repositories
+
+**Status:** provisional
+
+**Creation date:** 2023-31-07
+
+## Summary
+
+Flux should provide a mechanism to authenticate against Git repositories without
+the use of passwords. This RFC proposes the use of alternative authentication
+methods like OIDC, OAuth2 and IAM to access Git repositories hosted on specific
+Git SaaS platforms and cloud providers.
+
+## Motivation
+
+At the moment, Flux supports HTTP basic and bearer authentication. Users are
+required to create a Secret containing the username and the password/bearer
+token, which is then referred to in the GitRepository using `.spec.secretRef`.
+
+While this works fine, it has a couple of drawbacks:
+* Scalability: Each new GitRepository potentially warrants another credentials
+pair, which doesn't scale well in big organizations with hundreds of
+repositories with different owners, increasing the risk of mismanagement and
+leaks.
+* Identity: A username is associated with an actual human. But often, the
+repository belongs to a team of 2 or more people. This leads to a problem where
+teams have to decide whose credentials should Flux use for authentication.
+
+These problems exist not due to flaws in Flux, but because of the inherent
+nature of password based authentication.
+
+With support for OIDC, OAuth2 and IAM based authentication, we can eliminate
+these problems:
+* Scalability: Since OIDC is fully handled by the cloud provider, it eliminates
+any user involvement in managing credentials. For OAuth2 and IAM, users do need
+to provide certain information like the ID of the resource, private key, etc.
+but these are still a better alternative to passwords since the same resource
+can be reused by multiple teams with different members.
+* Identity: Since all the above authentication methods are associated with a
+virtual resource independent of a user, it solves the problem of a single person
+being tied to automation that several people are involved in.
+
+### Goals
+
+* Integrate with major cloud providers' OIDC and IAM offerings to provide a
+seamless way of Git repository authentication.
+* Integrate with major Git SaaS providers to support their app based OAuth2
+mechanism.
+
+### Non-Goals
+* Replace the existing basic and bearer authentication API.
+
+## Proposal
+
+A new string field `.spec.provider` shall be added to the GitRepository API.
+The field will be an enum with the following variants:
+* `azure`
+* `github`
+* `gcp`
+
+> AWS CodeCommit is not supported as it does not support authentication via IAM
+Roles without the use of https://github.com/aws/git-remote-codecommit.
+
+By default, it will be blank, which indicates that the user wants to
+authenticate via HTTP basic/bearer auth or SSH.
+
+### Azure
+
+Git repositories hosted on Azure Devops can be accessed by Flux using OIDC if
+the cluster running Flux is hosted on AKS with [managed identity](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?view=azure-devops)
+enabled. The managed identity associated with the cluster must have sufficient
+permissions to be able to access Azure Devops resources. This enables Flux to
+access the Git repository without the need for any credentials.
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: azure-devops
+spec:
+  interval: 1m
+  url: https://dev.azure.com/<org>/<project>/_git/<repository>
+  ref:
+    branch: master
+  # notice the lack of secretRef
+  provider: azure
+```
+
+### GCP
+
+Git repositories hosted on Google Cloud Source Repositories can be accessed by
+Flux via a [GCP Service Account](https://cloud.google.com/iam/docs/service-account-overview).
+The Service Account must have sufficient permissions to be able to access Google
+Cloud Source Repositories and its credentials should be specified in the secret
+referred to in `.spec.secretRef`.
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: gcp-repo
+spec:
+  interval: 1m
+  url: https://source.developers.google.com/p/<project>/r/<repository>
+  ref:
+    branch: master
+  provider: gcp
+  secretRef:
+    name: gcp-sa
+---
+kind: Secret
+metadata:
+  name: gcp-sa
+stringData:
+  gcpServiceAccount: |
+    {
+      "type": "service_account",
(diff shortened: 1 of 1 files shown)


makkes (commented):
I like the general approach here. What I missed while reading the RFC is the handling of token lifetime. How is an expired token refreshed?

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,198 @@
...
+### Non-Goals
+* Replace the existing basic and bearer authentication API.
+
+## Proposal
+
+A new string field `.spec.provider` shall be added to the GitRepository API.
+The field will be an enum with the following variants:
+* `azure`
+* `github`
+* `gcp`
```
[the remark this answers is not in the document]
stefanprodan: Which Git SaaS uses OIDC? GCP and AWS they don’t even support workload identity, GH uses the property GH App API.

aryan9600:
>I like the general approach here. What I missed while reading the RFC is the handling of token lifetime. How is an expired token refreshed?

@makkes see the "Caching" section at the bottom. do you feel like that section does not contain enough info about handling of expiring tokens? what more would you like to see in the section?

rfcs/0006-git-repo-passwordless-auth/README.md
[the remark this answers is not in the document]
aryan9600: @makkes  the goal here is to not support OIDC providers. the goal is to provide platform/cloud provider native auth mechanisms as an alternative to regular passwords.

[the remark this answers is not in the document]
aryan9600: if you feel the RFC was not clear about this, kindly recommend suggestions that will make it clearer.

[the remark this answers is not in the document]
hiddeco: > GH uses the property GH App API

GitHub Apps is pretty much just an OAuth 2.0 (i.e. [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628)) mechanism.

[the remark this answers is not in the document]
makkes: One example: Gitea [can act as an OAuth2 provider](https://docs.gitea.com/next/development/oauth2-provider). During bootstrap the Flux CLI can redirect the user to the authentication page (pretty much like [kubelogin](https://github.com/int128/kubelogin) does. That would allow Flux to support a much broader set of Git hosting implementations.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+### GitHub
+
+If `.spec.provider` is set to `github`, Flux controllers will get the app
+details from the specified Secret and use it to [generate an app installation
+token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app).
+This token is then used as the password and [`x-access-token` as the username](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/choosing-permissions-for-a-github-app#choosing-permissions-for-git-access)
+to perform HTTP basic authentication.
+
+### Caching
```
makkes, line 189:
```suggestion
### Caching and Token Refresh
```

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+To avoid calling the upstream API for a token during every reconciliation, Flux
+controllers shall cache the token after fetching it. Since GitHub and GCP tokens
+self-expire, the cache shall automatically evict the token after it has expired,
+triggering a fetch of a fresh token.
```
makkes, line 194: Why not make use of a refresh token here? Do the SaaS providers mentioned throughout this RFC support that?

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,198 @@
...
+If `.spec.provider` is set to `github`, Flux controllers will get the app
+details from the specified Secret and use it to [generate an app installation
+token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app).
+This token is then used as the password and [`x-access-token` as the username](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/choosing-permissions-for-a-github-app#choosing-permissions-for-git-access)
+to perform HTTP basic authentication.
+
+### Caching
+
+To avoid calling the upstream API for a token during every reconciliation, Flux
+controllers shall cache the token after fetching it. Since GitHub and GCP tokens
```
pjbgf, line 192: The varying auth mechanisms may provide a way to tell their token's TTL (e.g. the `JWT` may have an `expiry` attribute) which could be used to inform the caching practices. Although, a `max` cache period should be defined/enforced by the controller.

rfcs/0006-git-repo-passwordless-auth/README.md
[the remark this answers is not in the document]
aryan9600: @makkes i'm sorry if you misunderstood but this RFC is not about the CLI at all. its about users not having to put passwords in the cluster and having to manage them. adding support for OAuth2 as a part of the bootstrap process lies completely outside the scope of this RFC. out of the three providers this RFC proposes, only one (GitHub) is actually supported by bootstrap.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,198 @@
...
+  # notice the lack of secretRef
+  provider: azure
+```
+
+### GCP
+
+Git repositories hosted on Google Cloud Source Repositories can be accessed by
+Flux via a [GCP Service Account](https://cloud.google.com/iam/docs/service-account-overview).
+The Service Account must have sufficient permissions to be able to access Google
+Cloud Source Repositories and its credentials should be specified in the secret
```
pjbgf, line 94: Doesn't Workload Identity support this? Specially for cloud environments it would be great to lean on contextual auth (even if bound to the current SA) and avoid the increased attack surface (of handling cloud secrets).

rfcs/0006-git-repo-passwordless-auth/README.md
aryan9600: None of the mentioned providers support the `refresh_token` attribute.
* GitHub: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app
* Azure: https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
* GCP: A [`TokenSource`](https://pkg.go.dev/golang.org/x/oauth2@v0.10.0#TokenSource) automatically handles the refreshing of an expired token, so we can just cache the `TokenSource` object and call `.Token()` on it to always get a valid token. This is what also Argo does: https://github.com/argoproj/argo-cd/blob/72754d772610636af95c030bc0f7608633a80896/util/git/creds.go#L539-L541

rfcs/0006-git-repo-passwordless-auth/README.md
aryan9600: yes, the cache eviction will be based on the `expiry` attribute, except for GCP, because we will be caching the [`TokenSource`](https://pkg.go.dev/golang.org/x/oauth2@v0.10.0#TokenSource) object directly. this is because the token returned by a `TokenSource` will expire but the `TokenSource` object itself won't and it automatically handles refreshing expired tokens.

rfcs/0006-git-repo-passwordless-auth/README.md
aryan9600: nope it doesn't (or at least i couldn't find the relevant docs).

rfcs/0006-git-repo-passwordless-auth/README.md
aryan9600: will update the RFC to mention this workaround.

Poltergeisen:
Is this still being planned? I'm very interested in being able to use managed identities for git instead of a manual step right now of adding a secret to my clusters.

schdief:
would also be interested in using Azure DevOps managed identities/service principals

bavneetsingh16:
Would any of the maintainers sponsor this RFC? I would like to start making progress on this.

souleb:
> Would any of the maintainers sponsor this RFC? I would like to start making progress on this.

Thanks for stepping up @bavneetsingh16. I volunteer to sponsor this RFC.

rfcs/0006-git-repo-passwordless-auth/README.md
zonorti: Google cloud source repo supports using credentials-helper: https://github.com/twistedpair/google-cloud-sdk/blob/master/google-cloud-sdk/lib/googlecloudsdk/api_lib/source/git.py#L300
And passing the token as password works.
```
Cloning into 'xxxx'...
Username for 'https://source.developers.google.com': oauth
Password for 'https://oauth@source.developers.google.com': 
```
So using workload identity or even workload federation should be an option

stefanprodan:
Superseded by #4806

ageisen2000:
> Superseded by #4806

Is there another PR in flight for this or will it be implemented later then?

stefanprodan:
> Is there another PR in flight for this or will it be implemented later then?

First the RFC has to be agreed on and merged, then we can talk about implementations. I've added the RFC to the roadmap but it all depends on the availability of those involved.

Issue #3892, opened 2023-05-17
`flux create source git` with `--password` fails to create secret


DeyvsonL:
### Describe the bug

When using flux cli, I expect to be able to use Azure DevOps personal access token (PAT). But the command "flux create source git" delete the Kubernetes secret when the username is not specified. If we modify the secret manually, removing the username, everything still working.

### Steps to reproduce

1. install flux in a Kubernetes cluster
2. run command line "flux create source git password='${my-PAT}' --url='${my-git-repo}' name"
3. flux gitreposity created but authentication fails
4. secret with the same as the gitrepository

### Expected behavior

A kubernetes secret shoul be created without username but with password field.

### OS / Distro

Ubuntu 20.04

### Flux version

v0.34.0

### Flux check

► checking prerequisites
✗ flux 0.34.0 <2.0.0-rc.3 (new version is available, please upgrade)
✔ Kubernetes 1.23.7 >=1.20.6-0
✔ helm-controller: deployment ready
✔ image-automation-controller: deployment ready
✔ image-reflector-controller: deployment ready
✔ kustomize-controller: deployment ready
✔ notification-controller: deployment ready
✔ source-controller: deployment ready
✔ alerts.notification.toolkit.fluxcd.io/v1beta1
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1beta2
✔ helmcharts.source.toolkit.fluxcd.io/v1beta2
✔ helmreleases.helm.toolkit.fluxcd.io/v2beta1
✔ helmrepositories.source.toolkit.fluxcd.io/v1beta2
✔ imagepolicies.image.toolkit.fluxcd.io/v1beta1
✔ imagerepositories.image.toolkit.fluxcd.io/v1beta1
✔ imageupdateautomations.image.toolkit.fluxcd.io/v1beta1
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1beta2
✔ providers.notification.toolkit.fluxcd.io/v1beta1
✔ receivers.notification.toolkit.fluxcd.io/v1beta1
✔ all checks passed

### Git provider

Azure DevOps

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kingdonb:
The current version of the CLI is Flux v2 `2.0.0-rc.4` – the 0.34 release is a very old one.

[...]

PR #4119, opened 2023-08-01
removing username being mandatory with password in source secret with…

talife:
proposition to fix fluxcd/flux2#3892

test added with a golden file to create a secret with only a password key.


hiddeco:
This should not work, as we validate a HTTP/S URL to contain a username in the underlying Git library: https://github.com/fluxcd/pkg/blob/main/git/options.go#L62-L65. Given you highlight in the issue that manually constructing the Secret works as expected, I am wondering if you are trying to create the Secret for an SSH URL without having `ssh://` specified.

Issue #4125, opened 2023-08-02
Prepare Flux 2.1.0


stefanprodan:
Flux 2.1.0 comes with new features, bug fixes and performance improvements. 

:date: **Target dates:** 21-24 August 2023

⚠️ Breaking changes:

This release comes with breaking changes to the Flux monitoring stack (Prom+Grafana). We are leveraging `kube-state-metrics` CRD exporter to report Flux metrics, the Grafana dashboards shipped with Flux will target the new metrics. This will allow users to extend the Flux metrics with custom metadata such as Git branch, commit, Helm chart version, etc.

Notable changes:
- Add `.spec.proxySecretRef.name` to the `GitRepository` API  (source-controller)
- Add `.spec.certSecretRef`  to the `HelmRepository` API (source-controller)
- Add support for self-signed certs for OCI HelmRepositories (source-controller)
- Perf improvements when parsing large Helm index files (source-controller)
- Improve health checking by detecting stalled Kubernetes rollouts (kustomize-controller)
- Perf improvements when reconciling a Flux Kustomization with thousands of resources (kustomize-controller) 
- Better load distribution when reconciling Flux objects in parallel (all controllers)
- Revamp Flux Prometheus monitoring using kube-state-metrics
- Revamp the installation and bootstrap configuration documentation
- Publish docs for each Flux minor version on a dedicated subdomain 

TODOs:
- https://github.com/fluxcd/flux2/issues/4120
- https://github.com/fluxcd/flux2/pull/4116
- https://github.com/fluxcd/source-controller/pull/1178
- https://github.com/fluxcd/source-controller/pull/1097
- https://github.com/fluxcd/source-controller/pull/1176
- https://github.com/fluxcd/source-controller/pull/1160
- https://github.com/fluxcd/source-controller/pull/1109
- https://github.com/fluxcd/kustomize-controller/pull/933
- https://github.com/fluxcd/kustomize-controller/pull/932
- https://github.com/fluxcd/website/pull/1573
- https://github.com/fluxcd/flux2/issues/4111

Release checklist:
- [ ] source-controller v1.1.0
- [ ] kustomize-controller v1.1.0
- [ ] notification-controller v1.1.0
- [ ] helm-controller: v0.36.0
- [ ] image-reflector-controller v0.30.0
- [ ] image-automation-controller v0.36.0
- [ ] flux2 v2.1.0
- [ ] terraform-provider-flux v1.1.0

Documentation:
[...]

PR #4126, opened 2023-08-02
Set Kubernetes min version to 1.25

stefanprodan:
Drop support for Kubernetes 1.24 (EOL 28 Jul 2023)

Part of: #4125


makkes (approved):
Do

PR #4138, opened 2023-08-07
adding printing diff functionality in flux diff artifact

### LINKED ISSUE #3839: Allow diffing of two OCI artifacts
To help people to gather an idea about the difference between two artifacts, we could introduce a `flux diff artifact <URL1> <URL2>` command which provides a (Git like) diff of the differences within the compressed `tar+gzip` layer of the Artifact image.

To build this functionality, we need to decide on the diff library that should be used. Theoretically, we could piggy back on the [diff utility library](https://github.com/go-git/go-git/blob/master/utils/diff/diff.go) from go-git in combination with [`DiffPrettyText`](https://github.com/sergi/go-diff/blob/facec63e78161d6d31a9c552a679e2287e925949/diffmatchpatch/diff.go#L1147). However, there are discussions at present to [replace this](https://github.com/go-git/go-git/discussions/747) due to certain issues.

An additional thing to take into account is that full decompression of `<URL1>` and/or `<URL2>` to disk may not be required, but that we  rather iterate over the contents using [` tar#NewReader`](https://pkg.go.dev/archive/tar#NewReader) in combination with reading the file contents from the current [`tar#Header`](https://pkg.go.dev/archive/tar#Header) using `io.Copy(out, reader)`.

sbdtu5498:
Fixes #3839
cc: @hiddeco @makkes @somtochiama 
WIP. Please take a look. For the time being I have added a verbose flag for the functionallity to keep it backward compatible in terms of testing, but it can be removed as other diff commands also simply print the diff and thus it makes sense. I will be adding the test cases soon. Let me know what should be the case.

### DIFF
--- a/cmd/flux/diff_artifact.go
+++ b/cmd/flux/diff_artifact.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Flux authors
+Copyright 2023 The Flux authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,12 +17,18 @@ limitations under the License.
 package main
 
 import (
+	"archive/tar"
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 
 	oci "github.com/fluxcd/pkg/oci/client"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
+	"github.com/google/go-containerregistry/pkg/crane"
+
+	"github.com/sergi/go-diff/diffmatchpatch"
 	"github.com/spf13/cobra"
 
 	"github.com/fluxcd/flux2/v2/internal/flags"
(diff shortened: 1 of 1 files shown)


cmd/flux/diff_artifact.go
```diff
@@ -103,10 +111,112 @@ func diffArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to read tar header: %w", err)
+		}
+
+		// Read the file contents and write them to the buffer
+		if _, err := io.Copy(&contentsBuffer, tarReader); err != nil {
+			return nil, fmt.Errorf("failed to read tar entry contents: %w", err)
+		}
+	}
```
hiddeco, line 181: Instead of this looking at just file contents, I think it would be much better if path data would be taken into account (which also can be done for local directories, as they are expected to have the same structure as a decompressed tar).

This way, not only differences in file content would be observed. But also things being moved (or swapped) from file A -> B, files changing from path, etc.

Issue #4137, opened 2023-08-05
Support Kubernetes style TLS Secrets


aryan9600:
Currently, we have two APIs (OCIRepository and HelmRepository) that provide a way to configure TLS authentication. Both APIs expect the specified Secret to contain three keys, namely `caFile`, `certFile` and `keyFile`. This allows for users to specify their TLS data, but it doesn't align with the API of a [Kubernetes Secret of type TLS](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets).
A TLS secret needs to contain the keys `tls.crt` for the certificate and `tls.key` for the private key. While it doesn't specify an exact key for the CA file, the common convention is to use `ca.crt` as used by [cert-manager](https://cert-manager.io/docs/concepts/certificate/) and [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/main/docs/examples/auth/client-certs/README.md).
To make adoption and usage easier, we should adopt these standards in the Flux project as well. This change would require the following things:
* OCIRepository: Add support for `tls.crt`, `tls.key` and `ca.crt` keys in the Secret specified in `.spec.certSecretRef` and deprecate the currently supported keys, in a way that is similar to the [deprecation of `.spec.secretRef` for TLS data in HelmRepository](https://github.com/fluxcd/source-controller/pull/1160). Support for the existing keys would be completely dropped when OCIRepository receives it's next API version bump.
* HelmRepository: Drop support for the currently supported keys in the Secret specified in `.spec.certSecretRef` and replace it with the `tls.crt`, `tls.key ` and `ca.crt`.

hiddeco:
Instead of deleting `flux create secret helm` right away, I would like it to be a more gradual deprecation in which we first log a deprecation warning. Which is a tad less aggressive towards users.

stefanprodan:
As a first step we can add the new flags to all commands and deprecate the old flags. The controllers and the CLI should work with both formats in the next release. I agree that we should update the docs to use the new CLI args and Secret format.

PR #4147, opened 2023-08-09
Adopt Kubernetes style TLS Secrets and add relevant flags

aryan9600:
* `flux create secret helm`: Deprecate the `--ca-file`, `--cert-file` and `key-file` flags and recommend users to use `flux create secret tls` to generate TLS Secrets.
* `flux create secret tls`: Add flags `--tls-crt-file`, `--tls-key-file` and `--ca-crt-file` and deprecate the current TLS flags. Use the `tls.key`,  `tls.crt` and `ca.crt` keys in the generated Secret if the command is invoked using the new flags.
* `flux create secret git`: Add `--ca-crt-file` as an alternative to `--ca-file` that uses `ca.crt` in the generated Secret. `--ca-crt-file` takes precedence over `--ca-file`.

Part of: #4137


pkg/manifestgen/sourcesecret/options.go
```diff
@@ -54,12 +60,18 @@ type Options struct {
 	Keypair             *ssh.KeyPair
 	Username            string
 	Password            string
-	CAFile              []byte
-	CertFile            []byte
-	KeyFile             []byte
+	CACrt               []byte
+	TlsCrt              []byte
+	TlsKey              []byte
```
hiddeco, line 65:
```suggestion
	TLSCrt              []byte
	TLSKey              []byte
```

In addition: any specific reason to use "crt" instead of "cert"? Which is prettier to the eyes, and not much longer.

pkg/manifestgen/sourcesecret/options.go
```diff
@@ -54,12 +60,18 @@ type Options struct {
...
 	TargetPath          string
 	ManifestFile        string
 	BearerToken         string
+
+	// Depreacted: These fields are used to store TLS data that
+	// specified by the deprecated TLS flags.
+	CAFile   []byte
+	CertFile []byte
+	KeyFile  []byte
```
hiddeco, line 74:
```suggestion
	// Deprecated: Replaced by CACrt, but kept for backwards compatibility 
	// with deprecated TLS flags.
	CAFile   []byte
	// Deprecated: Replaced by TLSCrt, but kept for backwards compatibility
	// with deprecated TLS flags. 
	CertFile []byte
	// Deprecated: Replaced by TLSKey, but kept for backwards compatibility
	// with deprecated TLS flags.
	KeyFile  []byte
```

To ensure highlighting in IDEs per field, including information about what should be used instead.

hiddeco (approved):

pkg/manifestgen/sourcesecret/sourcesecret.go
```diff
@@ -156,13 +156,19 @@ func buildSecret(keypair *ssh.KeyPair, hostKey, caFile, certFile, keyFile, docke
 		secret.StringData[BearerTokenKey] = options.BearerToken
 	}
 
-	if len(caFile) != 0 {
-		secret.StringData[CAFileSecretKey] = string(caFile)
+	if len(options.CACrt) != 0 {
+		secret.StringData[CACrtSecretKey] = string(options.CACrt)
+	} else if len(options.CAFile) != 0 {
+		secret.StringData[CAFileSecretKey] = string(options.CAFile)
```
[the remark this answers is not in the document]
aryan9600: its intentional, the aim is to deprecate the current TLS flags, not modify its behavior and then remove them altogether.

PR #4175, opened 2023-08-21
build: update securejoin dependency

hiddeco:


stefanprodan (approved):

mestadler:
Does the current release of flux contain this merge or is that still pending for a next release?

hiddeco:
It is included in the current release.

PR #4177, opened 2023-08-21
Set min value for the `--ssh-rsa-bits` flag

stefanprodan:
Don't generate RSA keys with a bit size less than 1024.


makkes (commented):
We should add this constraint to the help text, too.

hiddeco (approved):

hiddeco:
We should add a test case for `512`, and change `123` to one above `1024` which can not be divided by 8.

makkes (approved):

PR #4205, opened 2023-08-30
use flux filename instead of gotk

remyleone:


stefanprodan:
Flux v2 is GA, this change would require a v3 release.

stefanprodan:
Modifying the bootstrap file names would break everyone's use of Flux.

Issue #4218, opened 2023-09-06
Flux bootstrap for a project hosted on GitLab on-prem


lakshman-au:
Having an AKS cluster in air-gapped environment and gitlab hosted on-prem, I am trying to bootstrap with the below command

flux bootstrap gitlab --hostname=$GITLAB_URL --owner=$GITLAB_GROUP --repository=$GITLAB_REPO --branch=flux_test --path=manifests/staging --registry=[azurecr.io/$ACR_NAME/fluxcd](http://azurecr.io/$ACR_NAME/fluxcd) --image-pull-secret=acr-auth --interval 10m --token-auth --reconcile

I am getting error , resource already exists. How can I bootstrap with existing repository

stefanprodan:
What's the reason for using `--reconcile`?

lakshman-au:
the repository already exists,

stefanprodan:
Run it without `--reconcile`

lakshman-au:
Let me try and update

stefanprodan:
What version of Flux are you using? Is it 2.1.0?

lakshman-au:
yes the version is 2.1.0

stefanprodan:
@somtochiama could you please have a look?

somtochiama:
@lakshman-au  I have deployed a self-hosted version of GitLab but I have been unable to recreate this error. I tried to have the same exact command including pushing flux to an acr registry. 

Can you give more details about your GitLab repository? Does the name of the repository/group) differ from the slug(the path you see in the URL in the browser?) or are there any special characters in some of the values you have omitted? Any extra info will be helpful

Also what Gitlab version are you running, I tested on `16.3.2-ee`

lakshman-au:
Gitlab Enterprise version: [15.11.9-ee]
I  just added few options ( author email,author-name,commit-message)

flux bootstrap gitlab [--author-email=lakshmanakumar.duraipandi@credit-suisse.com](mailto:--author-email=lakshmanakumar.duraipandi@credit-suisse.com) --author-name="Lakshman Duraipandi" --hostname=$GITLAB_URL \
  --insecure-skip-tls-verify --owner=$GITLAB_GROUP --repository=$GITLAB_REPO --branch=develop \
  --path=clusters/my-cluster/  --registry=[cordevuc5acrdevelopb5f3.azurecr.io/fluxcd](http://cordevuc5acrdevelopb5f3.azurecr.io/fluxcd) \                                                                                                      
[...]

PR #4226, opened 2023-09-08
Update description of kubeconfig specific flag

somtochiama:
The `--insecure-skip-tls-verify`, `--certificate-authority`, `client-key` and `client-certificate` are kubeconfig flags that are used when connecting to the Kubernetes API Server. The generic description sometimes makes users assume that this is for something else e.g [see comment](https://github.com/fluxcd/flux2/issues/4218#issuecomment-1711741628)

> Though I have the option --insecure-skip-tls-verify , still getting certifcate error.

This pull request updates the description to be clear about which server it is used for.
Before
```
--insecure-skip-tls-verify               If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
--certificate-authority string           Path to a cert file for the certificate authority
--client-certificate string              Path to a client certificate file for TLS
--client-key string                      Path to a client key file for TLS
```

After
```
--insecure-skip-tls-verify       If true, the Kubernetes API server's certificate will not be checked for validity. This will make your HTTPS connections insecure
--certificate-authority string   Path to a cert file for the certificate authority to authenticate Kubernetes API
--client-certificate string      Path to a client certificate file for TLS authentication to Kubernetes API
--client-key string              Path to a client key file for TLS authentication to Kubernetes API
```

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -151,6 +151,11 @@ func init() {
 	apiServer := ""
 	kubeconfigArgs.APIServer = &apiServer
 	rootCmd.PersistentFlags().StringVar(kubeconfigArgs.APIServer, "server", *kubeconfigArgs.APIServer, "The address and port of the Kubernetes API server")
+	// Update the description for kubeconfig TLS flags so that user's don't mistake it for a Flux specific flag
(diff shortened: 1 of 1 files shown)


stefanprodan:
I propose we do the same some for:
```
      --certificate-authority string   Path to a cert file for the certificate authority
      --client-certificate string      Path to a client certificate file for TLS
      --client-key string              Path to a client key file for TLS
```

stefanprodan (approved):

Issue #4026, opened 2023-06-30
Move away from AUR templates in favor of goreleaser feature


hiddeco:
Fairly simple: we probably should move away from our templates in https://github.com/fluxcd/flux2/tree/main/.github/aur in favor of https://goreleaser.com/customization/aur/.

hiddeco:
Reading more, this only seems to be able to replace our `-bin` package (see note `.aurs.name`) while it appears it wouldn't allow us to continue to maintain the other versions.

PR #4228, opened 2023-09-11
Improve AUR package templates

### LINKED ISSUE #4224: Improve AUR package templates
Improvement recommendations from @FabioLolix on AUR: https://aur.archlinux.org/packages/flux-bin#comment-928509

> Hello,
> 
> * you can remove `armv6h` since that is has been removed from the distro
> * please remove `${pkgname}-${pkgver}.tar.gz::` from source_*=(), renaming all the files to the same name is detrimental if one need to use the command updpkgsums
> * improve pkgver/_srcver instead of duplicate it by using for example:
> ```
> _srcver=2.0.0-rc.5
> pkgver=${_srcver//-/.}
> ```
> or
> ```
> _srcver=2.0.0-rc.5
> pkgver=${_srcver//-/}
> ```

sonbui00:
Fix #4224
- remove armv6h 
- unique source name 
- improve pkgver/_srcver
- fix source name in .SRCINFO


.github/aur/flux-bin/PKGBUILD.template
```diff
@@ -1,35 +1,29 @@
 # Maintainer: Aurel Canciu <aurelcanciu@gmail.com>
 # Maintainer: Hidde Beydals <hello@hidde.co>
 
-pkgname=flux-bin
-pkgver=${PKGVER}
-pkgrel=${PKGREL}
 _srcname=flux
 _srcver=${VERSION}
+pkgname=flux-bin
+pkgver=${_srcver//-/}
```
hiddeco, line 7: As this is a template making use of `envsubst` for variable substitution, I am not quite sure this will actually work as intended in the produced `PKGBUILD` end result.

hiddeco:
Still Arch distroless at present, @relu @makkes any chance one of you can confirm this works as expected?

relu:
I'll have a look at it a bit later today.

.github/aur/flux-bin/PKGBUILD.template
makkes: As to Hidde's comment: I tested it locally and it works.

makkes: Let's revert this line and use `PKGVER` instead.

makkes (commented):
Except for the `PKGVER` usage this lgtm.

.github/aur/flux-bin/PKGBUILD.template
sonbui00: I've reverted

relu (changes requested):
Thank you, @sonbui00! One more improvement that I suggest be made here is replacing the SHA256 hashing from the script and templates and replacing that with `updpkgsums`.

Remove [this](https://github.com/fluxcd/flux2/blob/main/.github/aur/flux-bin/publish.sh#L40-L42) and the respective envsubst references and add `updpkgsums` after envsubst. You can then also remove the `sha256sums_*` lines from  PKGBUILD.

Also, we can drop the `.SRCINFO.template` and the respective `envsubst` command to generate `.SRCINFO` in favor of running `makepkg --printsrcinfo > .SRCINFO`.

sonbui00:
> Thank you, @sonbui00! One more improvement that I suggest be made here is replacing the SHA256 hashing from the script and templates and replacing that with `updpkgsums`.
> 
> Remove [this](https://github.com/fluxcd/flux2/blob/main/.github/aur/flux-bin/publish.sh#L40-L42) and the respective envsubst references and add `updpkgsums` after envsubst. You can then also remove the `sha256sums_*` lines from PKGBUILD.
> 
> Also, we can drop the `.SRCINFO.template` and the respective `envsubst` command to generate `.SRCINFO` in favor of running `makepkg --printsrcinfo > .SRCINFO`.

@relu There are some issues/tradeoff here:
- We run this on ubuntu, so we need to install updpkgsums and makepkg. These packages can install via github action, but it also slow down process
- I think updpkgsums need to download release packages that maybe not available at that time (need to check gorelease flow) and the download also make process slow down

relu (approved):
Fair points, thanks for looking into it. I think we can leave as is for now considering the limitations.

sonbui00:
@relu Maybe this https://github.com/fluxcd/flux2/issues/4026 
https://goreleaser.com/customization/aur/

hiddeco (approved):

PR #4278, opened 2023-09-22
enable gitea support

malarinv:


makkes:
This is a duplicate of #4270. I suggest for you two to collaborate on the PR there.

PR #4284, opened 2023-09-27
Make `flux pull` work for OCI artifacts produced by other tools

errordeveloper:
Allow `flux pull artifact` to retrieve any kind of OCI artifact without an index, e.g. a Helm chart for debugging.

This is a follow-up to fluxcd/pkg#653

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,12 +21,12 @@ require (
 	github.com/fluxcd/pkg/git v0.14.0
 	github.com/fluxcd/pkg/git/gogit v0.14.0
 	github.com/fluxcd/pkg/kustomize v1.3.4
-	github.com/fluxcd/pkg/oci v0.31.0
+	github.com/fluxcd/pkg/oci v0.32.0
 	github.com/fluxcd/pkg/runtime v0.42.0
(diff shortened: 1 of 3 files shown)


errordeveloper:
@stefanprodan @hiddeco this change works for my own test case, I do wonder if we need a tag in pkg repo and if there is anything else to take care of here?

stefanprodan:
What will `flux list artifacts` do? Will it panic or error or list an empty table when the annotations are missing?

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/fluxcd/pkg/git v0.14.0
 	github.com/fluxcd/pkg/git/gogit v0.14.0
 	github.com/fluxcd/pkg/kustomize v1.3.4
-	github.com/fluxcd/pkg/oci v0.31.0
+	github.com/fluxcd/pkg/oci v0.31.2-0.20230927081137-4f69e7842ce2
```
stefanprodan, line 24: Please bump the version to `oci/v0.32.0` and run `make tidy`.

stefanprodan (approved):
LGTM

PR #4296, opened 2023-10-04
fix: only wait for changeset if the result is not empty

Skarlso:
When using Apply with a set of files that end up with an empty first change set, wait will timeout waiting for nothing.

Only wait if the changeSet isn't empty.

### DIFF
--- a/internal/utils/apply.go
+++ b/internal/utils/apply.go
@@ -76,8 +76,10 @@ func Apply(ctx context.Context, rcg genericclioptions.RESTClientGetter, opts *ru
 		changeSet.Append(cs.Entries)
 	}
 
-	if err := waitForSet(rcg, opts, changeSet); err != nil {
-		return "", err
+	if len(changeSet.Entries) > 0 {
+		if err := waitForSet(rcg, opts, changeSet); err != nil {
+			return "", err
+		}
 	}
 
 	if len(stageTwo) > 0 {


Skarlso:
Err, not sure what's wrong with the DCO job? I have this as commit message:

```
fix: only wait for changeset if the result is not empty

Signed-off-by: Gergely Brautigam <182850+Skarlso@users.noreply.github.com>
```

Is that not correct?

internal/utils/apply.go
```diff
@@ -76,8 +76,10 @@ func Apply(ctx context.Context, rcg genericclioptions.RESTClientGetter, opts *ru
 		changeSet.Append(cs.Entries)
 	}
 
-	if err := waitForSet(rcg, opts, changeSet); err != nil {
-		return "", err
+	if len(changeSet.Entries) > 0 {
```
makkes, line 79: I'm not sure if this shouldn't be rather be made part of `ResourceManager#WaitForSet`.

Skarlso: Hmmm I think if you want to wait you would want to wait for _something_ and not wait for whatever even if there is nothing there. This makes more sense here readability wise.

Ultimately I don't care where this check lives. 😊

makkes: If I call `ResourceManager#WaitForSet` with an empty set and it doesn't immediately return (because there's nothing to wait for) this is a bug in that method IMHO.

Skarlso: Why would you call Wait if there is nothing to wait for? :)

Skarlso: That's like, I don't expect any visitors, but I open the door just in case. :D

stefanprodan:
This looks Ok to me, but where did you encounter an empty changeset using the CLI?

Skarlso:
@stefanprodan I was monkeying around. :D Also, we re-used a tiny part of the flux code in OCM, like this Apply for a bootstrap command for OCM. ( We gave credit. :).

And then I applied some manifests which didn't have type 1 objects in it and it hang.

For flux, it would it mean during install or bootstrapping there is a corrupted manifest in Github I guess.

stefanprodan:
A corrupted manifest will not pass validation which happens way before you get a changeset.

Skarlso:
Alright, so outside of me monkeying, if you think it's unlikely to occur, I'm fine with ignoring it. :)

stefanprodan:
Your PR is Ok, thank you! I was really intrigued if you actually hit this issue with bootstrap, when I wrote that code I was convinced there is no possible way to get an empty changeset as the bootstrap logic would fail way before.

Skarlso:
@stefanprodan Hah, neat. :) WDYT about what Max is saying? I'm okay to move it into the wait instead if you also think it's better suited there. :)

stefanprodan:
When the changeset could be empty, then we have this check in place, here is kustomize-controller:
https://github.com/fluxcd/kustomize-controller/blob/768968d061f2da9bd933d260aabcb2c8a24131ee/internal/controller/kustomization_controller.go#L844-L847

internal/utils/apply.go
makkes: A function that behaves indeterministically if called with certain input is a recipe for disaster. Be liberal in what you accept. But I really don't mind in this specific case so just leaving this here as a general remark.

stefanprodan: I agree that we should look into the logic here and see where it hangs: https://github.com/fluxcd/pkg/blob/main/ssa/manager_wait.go But also the check in this PR is good, no need to instantiate a whole resource manager if there is nothing to wait for.

Skarlso: I can do both if that's okay with you two?!

stefanprodan (approved):
LGTM

Thanks @Skarlso

hiddeco (approved):

hiddeco:
@Skarlso any chance you can rebase this quickly to maintain your signature? :smile:

Skarlso:
done

PR #4298, opened 2023-10-04
Add support for HelmRepo OCI and NC v1beta3 static objects

darkowlzz:
Depends on https://github.com/fluxcd/notification-controller/pull/540 and https://github.com/fluxcd/source-controller/pull/1243 that make HelmRepository OCI, Alert and Provider static objects.

This change makes the CLI aware of static objects. Static objects don't have status to report readiness. The CLI shows them as ready always.

https://github.com/fluxcd/source-controller/pull/1243 makes HelmRepository OCI static. CLI shows it as ready:
```console
$ flux -n default get sources helm example 
NAME    REVISION        SUSPENDED       READY   MESSAGE                  
example                 False           True    Helm repository is Ready

$ flux -n default get alerts default-alerts 
NAME            SUSPENDED       READY   MESSAGE        
default-alerts  False           True    Alert is Ready

$ flux -n default get alert-providers slack-test 
NAME            READY   MESSAGE           
slack-test      True    Provider is Ready
```

The static objects also don't support reconciliation. The CLI prints that the objects don't support reconciliation.

```console
$ flux -n default reconcile source helm example 
✔ reconciliation not supported by the object
```

`alert` and `alert-provider` subcommands for `reconcile` have been removed as they are completely static objects in Alert and Provider v1beta3 API.

Update all the references of NC v1beta2 API to v1beta3 and updates the tests accordingly.

flux2 GCP e2e test run with these changes https://github.com/fluxcd/flux2/actions/runs/7142695935/job/19452536537.

### DIFF
--- a/tests/integration/notification_test.go
+++ b/tests/integration/notification_test.go
@@ -27,7 +27,6 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
 	notiv1 "github.com/fluxcd/notification-controller/api/v1"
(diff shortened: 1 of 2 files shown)


cmd/flux/reconcile.go
```diff
@@ -60,6 +60,7 @@ type reconcilable interface {
 	GetAnnotations() map[string]string
 	SetAnnotations(map[string]string)
 
+	hasReconciler() bool                 // does the object's current configuration have a reconciler to reconcile it?
```
aryan9600, line 63: this is just a nit but a more apt name would be `isStatic`, since the static objects do have a reconciler to help with migration

Issue #3867, opened 2023-05-10
flux bootstrap gitlab ignores --owner when creating the repo


abderrahim:
### Describe the bug

Following the installation guide for gitlab to create my first flux repo, the repo is created in the wrong location (under my username) rather than the group specified with `--owner`.

### Steps to reproduce

I ran the following command: 
`flux bootstrap gitlab --hostname gitlab.codethink.co.uk --owner software-factory/playground/abderrahim --repository flux-test1 --token-auth --personal`

I got the following output
```
► connecting to https://gitlab.codethink.co.uk
✔ repository "https://gitlab.codethink.co.uk/software-factory/playground/abderrahim/flux-test1" created
► cloning branch "main" from Git repository "https://gitlab.codethink.co.uk/software-factory/playground/abderrahim/flux-test1.git"
⚠️  clone failure: unable to clone: repository not found: git repository: 'https://gitlab.codethink.co.uk/software-factory/playground/abderrahim/flux-test1.git'
⚠️  clone failure: unable to clone 'https://gitlab.codethink.co.uk/software-factory/playground/abderrahim/flux-test1.git': repository already exists
✗ failed to clone repository: unable to clone 'https://gitlab.codethink.co.uk/software-factory/playground/abderrahim/flux-test1.git': repository already exists
```

### Expected behavior

It should work, creating the repository and setting it up.

Instead, the repository was created under my username (rather than the group I specified), and subsequently cloning it failed. 

This might be due to the fact I was using a personal token rather than a project/group token, but it should either figure it out or explicitly throw an error.

### OS / Distro

Debian testing

### Flux version

2.0.0-rc.1

### Flux check

► checking prerequisites
✔ Kubernetes 1.26.3 >=1.20.6-0
► checking controllers
✗ no controllers found in the 'flux-system' namespace with the label selector 'app.kubernetes.io/part-of=flux'
► checking crds
✗ no crds found with the label selector 'app.kubernetes.io/part-of=flux'
✗ check failed

### Git provider

gitlab entreprise

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
Can you please using `flux bootstrap git --url`? You'll need to set the deploy key by hand but it should unblock you until we figure out a fix.

abderrahim:
[...]

PR #4324, opened 2023-10-12
Update go-git-providers to v0.19.1

somtochiama:
Closes: https://github.com/fluxcd/flux2/issues/3867


makkes (approved):

makkes:
This is much better than before. One thing we could still improve on would be that the message `✗ failed to create new Git repository "https://github.com/soule/fleet-infraxx": incorrect owner 'soule' passed in` isn't very actionable because the user **did** want to create the repo under that user's account so strictly speaking it's not an "incorrect owner", is it?

stefanprodan (changes requested):
Please squash the two commits.

somtochiama:
>  because the user did want to create the repo under that user's account so strictly speaking it's not an "incorrect owner", is it?

More like the wrong string is passed into the `--owner` flag. We can check for the error and return a better error.
@makkes  Is something like `please pass correct username when using --personal flag`  better?

stefanprodan (approved):
LGTM

Thanks @somtochiama 🏅

stefanprodan:
Maybe `the specified owner doesn't match the identity associated with the given token` is more accurate?

stefanprodan:
@somtochiama please make the changes upstream, we'll do another patch release of ggp then you can update this PR.

stefanprodan (approved):

PR #4332, opened 2023-10-15
Add CLI flags for OCIRepository signature verification

matheuscscp:


cmd/flux/create_source_oci.go
```diff
@@ -80,6 +81,7 @@ func init() {
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.secretRef, "secret-ref", "", "the name of the Kubernetes image pull secret (type 'kubernetes.io/dockerconfigjson')")
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.serviceAccount, "service-account", "", "the name of the Kubernetes service account that refers to an image pull secret")
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.certSecretRef, "cert-ref", "", "the name of a secret to use for TLS certificates")
+	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.verifySecretRef, "verify-ref", "", "the name of a secret to use for signture verification")
```
stefanprodan, line 84: Let’s rename this to`verify-secret-ref`

cmd/flux/create_source_oci.go
```diff
@@ -156,6 +158,15 @@ func createSourceOCIRepositoryCmdRun(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if secretName := sourceOCIRepositoryArgs.verifySecretRef; secretName != "" {
+		repository.Spec.Verify = &sourcev1.OCIRepositoryVerification{
+			Provider: "cosign",
```
stefanprodan, line 163: We’ll support other signers, so we can’t hardcode cosign here. Let’s add a flag for `verify-provider` that must be specified if `verify-secret-ref` is provided.

matheuscscp: fixed, thanks!

cmd/flux/create_source_oci.go
```diff
@@ -80,6 +82,8 @@ func init() {
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.secretRef, "secret-ref", "", "the name of the Kubernetes image pull secret (type 'kubernetes.io/dockerconfigjson')")
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.serviceAccount, "service-account", "", "the name of the Kubernetes service account that refers to an image pull secret")
 	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.certSecretRef, "cert-ref", "", "the name of a secret to use for TLS certificates")
+	createSourceOCIRepositoryCmd.Flags().StringVar(&sourceOCIRepositoryArgs.verifyProvider, "verify-provider", "", "the provider to use for signture verification")
```
stefanprodan, line 85: It’s signature (same type below), and this should be a typed flag like OCI provider with validation of the accepted values.

matheuscscp: fixed, thanks!!

cmd/flux/create_source_oci.go
```diff
@@ -156,6 +160,20 @@ func createSourceOCIRepositoryCmdRun(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if secretName := sourceOCIRepositoryArgs.verifySecretRef; secretName != "" {
```
stefanprodan, line 163: If only the verify provider is specified then it's keyless, so we need to handle that as well. Docs here: https://fluxcd.io/flux/components/source/ocirepositories/#keyless-verification

matheuscscp: fixed, thanks!!!

stefanprodan (commented):
Looks Ok now. Can you please add tests, for when we error out due to provider not set and for export when secret ref is specified.

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #4345, opened 2023-10-19
Prevent `flux install` from overriding bootrapped cluster

somtochiama:
This pull request adds a function to get information about the flux installation on a cluster and uses it to check if the cluster has already been bootstrapped to prevent an override when using `flux install`.


cmd/flux/install.go
```diff
@@ -119,6 +119,20 @@ func installCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	kubeClient, err := utils.KubeClient(kubeconfigArgs, kubeclientOptions)
```
stefanprodan, line 122: This happens way to early and it breaks `--export` which shouldn't require a cluster being available. Please move this right before `utils.Apply`.

cmd/flux/cluster_info.go
```diff
@@ -0,0 +1,79 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// bootstrapLabels are labels put on a resource by kustomize-controller. These labels on the CRD indicates
+// that flux has been bootstrapped.
+var bootstrapLabels = []string{"kustomize.toolkit.fluxcd.io/name", "kustomize.toolkit.fluxcd.io/namespace"}
```
stefanprodan, line 32: Please use the group name from the API package instead of hardcoding `kustomize.toolkit.fluxcd.io`.

cmd/flux/cluster_info.go
```diff
@@ -0,0 +1,79 @@
...
+
+// fluxClusterInfo contains information about an existing flux installation on a cluster.
+type fluxClusterInfo struct {
+	// crdBootstrapped indicates if flux was installed using the `flux bootstrap` command.
+	bootstrapped bool
+	// managedBy contains the manager of the installation
+	managedBy string
+	// version contains the flux version of the installation.
+	version string
+}
```
stefanprodan, line 42:
```suggestion
// fluxClusterInfo contains information about an existing Flux installation on a cluster.
type fluxClusterInfo struct {
	// bootstrapped indicates that Flux was installed using the `flux bootstrap` command.
	bootstrapped bool
	// managedBy is the name of the manager who installed Flux.
	managedBy string
	// version is the Flux version number in semver format.
	version string
}
```

cmd/flux/cluster_info.go
```diff
@@ -0,0 +1,79 @@
...
+
+// gatherPreflightInfo checks the GitRepository CRD for labels passed into the function. It returns the version of flux
+// (from the app.kubernetes.io/version labels), a boolean to indicate whether the CRD was found
+// with the specified labels and an error.
```
stefanprodan, line 46:
```suggestion
// getFluxClusterInfo checks the GitRepository CRD for labels passed into the function. It returns the version of Flux
// (from the app.kubernetes.io/version label), a boolean to indicate whether the CRD was found
// with the specified labels and an error if the CRD query failed.
```

cmd/flux/cluster_info_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	. "github.com/onsi/gomega"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/fluxcd/pkg/ssa"
+)
+
+func Test_gatherPreflightInfo(t *testing.T) {
```
stefanprodan, line 32:
```suggestion
func Test_ getFluxClusterInfo(t *testing.T) {
```

cmd/flux/testdata/cluster_info/gitrepositories.yaml
```diff
@@ -0,0 +1,1196 @@
...
+      name: Age
+      type: date
+    - jsonPath: .status.conditions[?(@.type=="Ready")].status
+      name: Ready
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="Ready")].message
+      name: Status
+      type: string
+    name: v1
+    schema:
```
stefanprodan, line 33: Please trim down this file, we only need a single version and one spec field.

cmd/flux/cluster_info.go
```diff
@@ -0,0 +1,84 @@
...
+var bootstrapLabels = []string{
+	fmt.Sprintf("%s/name", kustomizev1.GroupVersion.Group),
+	fmt.Sprintf("%s/namespace", kustomizev1.GroupVersion.Group),
+}
+
+// fluxClusterInfo contains information about an existing flux installation on a cluster.
+type fluxClusterInfo struct {
+	// bootstrapped indicates that Flux was installed using the `flux bootstrap` command.
+	bootstrapped bool
+	// managedBy is the name of the manager who installed Flux.
```
makkes, line 43: Can you be a little more specific here on the specifics of a manager? I would suppose a potential manager would be Helm.

cmd/flux/cluster_info.go
```diff
@@ -0,0 +1,84 @@
...
+	managedBy string
+	// version is the Flux version number in semver format.
+	version string
+}
+
+// getFluxClusterInfo checks the GitRepository CRD for labels passed into the function. It returns the version of Flux
+// (from the app.kubernetes.io/version label), a boolean to indicate whether the CRD was found
+// with the required labels and an error if the CRD query failed.
```
makkes, line 51:
```suggestion
// getFluxClusterInfo returns information on the Flux installation running on the cluster. If the information cannot be retrieved, the boolean return value will be false. If an error occurred, the returned error will be non-nil.
// This function retrieves the GitRepository CRD from the cluster and checks it for a set of labels used to determine the Flux version and how Flux was installed.
```
Not sure about line length here, GitHub's UI doesn't help a lot. You might have to re-wrap it.

cmd/flux/install.go
```diff
@@ -183,6 +183,20 @@ func installCmdRun(cmd *cobra.Command, args []string) error {
...
+	}
+
+	info, installed, err := getFluxClusterInfo(ctx, kubeClient)
+	if err != nil {
+		return fmt.Errorf("error checking for exisiting flux installation: %w", err)
+	}
+
+	if installed && info.bootstrapped {
+		return fmt.Errorf("this cluster has already been bootstrapped with Flux %s! Please use 'flux bootstrap' to upgrade", info.version)
+	}
```
makkes, line 198: I wonder if any false positives could arise from this new behaviour in certain edge cases. Maybe a `--force` flag could help override the check in these cases.

makkes (commented):
This is great! We should incorporate the output of the `getFluxClusterInfo` into the `flux version` command, too.

cmd/flux/install.go
stefanprodan: I don't think we should allow people to force apply if the CRD is managed by Flux. The should uninstall and start fresh, forcing will only do harm as kustomize-controller will undo it.

cmd/flux/cluster_info.go
somtochiama: Yes. In the `getFluxClusterInfo` we are getting the manager on the `app.kubernetes.io/managed-by` label () but there's no constraint on what can be in the field. 
I could change it to say `the tool being used to manage the installation of Flux.`

cmd/flux/install.go
```diff
@@ -183,6 +183,20 @@ func installCmdRun(cmd *cobra.Command, args []string) error {
...
 	logger.Actionf("installing components in %s namespace", *kubeconfigArgs.Namespace)
 
+	kubeClient, err := utils.KubeClient(kubeconfigArgs, kubeclientOptions)
+	if err != nil {
+		return err
+	}
+
+	info, installed, err := getFluxClusterInfo(ctx, kubeClient)
+	if err != nil {
+		return fmt.Errorf("error checking for exisiting flux installation: %w", err)
```
stefanprodan, line 193:
```suggestion
		return fmt.Errorf("cluster info unavailable: %w", err)
```

makkes (approved):

stefanprodan (approved):
LGTM

Thanks @somtochiama 🥇

Issue #4342, opened 2023-10-19
Add prompt when overriding a Helm-managed Flux installation  


stefanprodan:
When running `flux bootstrap` or `flux install` we should detect if Flux is already installed on the cluster with Helm. 

We can look for the `app.kubernetes.io/managed-by` label on the `gitrepositories.source.toolkit.fluxcd.io` CRD. If the CRD exists and has this label, and if the value is set to `Helm` or any other manager besides `Flux`, we should prompt like so:

```
Flux <VERSION> has been installed on this cluster with <MANAGER>!
Are you sure you want to proceed with overriding the <MANAGER> installation? Y/N
```

The Flux version should be extracted from the `app.kubernetes.io/version` label value.

Both `flux bootstrap` and `flux install` commands should have a `--force` flag, when set, this will bypass the above prompt.

PR #4355, opened 2023-10-24
Confirm before overriding installation by another manager

somtochiama:
This pull request prompts a user to confirm before overriding an existing Flux installation that's not managed by flux when running `install` or `bootstrap`.

The pull request also adds a better uninstallation prompt when flux is managed by some other tool.

Fixes: https://github.com/fluxcd/flux2/issues/4342

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -17,11 +17,15 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/elliptic"
 	"fmt"
 	"strings"
 
+	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluxcd/flux2/v2/internal/flags"
 	"github.com/fluxcd/flux2/v2/internal/utils"
@@ -71,6 +75,8 @@ type bootstrapFlags struct {
 	gpgKeyRingPath string
 	gpgPassphrase  string
 	gpgKeyID       string
+
+	force bool
 
 	commitMessageAppendix string
 }
@@ -129,6 +135,7 @@ func init() {
 
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapArgs.commitMessageAppendix, "commit-message-appendix", "", "string to add to the commit messages, e.g. '[ci skip]'")
 
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapArgs.force, "force", false, "override existing Flux installation if it's managed by a diffrent tool such as Helm")
 	bootstrapCmd.PersistentFlags().MarkHidden("manifests")
 
 	rootCmd.AddCommand(bootstrapCmd)
@@ -187,4 +194,28 @@ func mapTeamSlice(s []string, defaultPermission string) map[string]string {
 	}
 
 	return m
+}
+
+// confirmBootstrap gets a confirmation for running bootstrap over an existing Flux installation.
+// It returns a nil error if Flux is not installed or the user confirms overriding an existing installation
+func confirmBootstrap(ctx context.Context, kubeClient client.Client) error {
+	installed := true
+	info, err := getFluxClusterInfo(ctx, kubeClient)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("cluster info unavailable: %w", err)
+		}
+		installed = false
+	}
+
+	if installed {
+		err = confirmFluxInstallOverride(info)
+		if err != nil {
+			if err == promptui.ErrAbort {
+				return fmt.Errorf("bootstrap cancelled")
+			}
+			return err
+		}
+	}
+	return nil
 }
--- a/cmd/flux/bootstrap_git.go
+++ b/cmd/flux/bootstrap_git.go
@@ -146,6 +146,13 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !bootstrapArgs.force {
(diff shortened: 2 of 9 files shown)


cmd/flux/bootstrap.go
```diff
@@ -129,6 +131,7 @@ func init() {
 
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapArgs.commitMessageAppendix, "commit-message-appendix", "", "string to add to the commit messages, e.g. '[ci skip]'")
 
+	bootstrapCmd.PersistentFlags().BoolVarP(&bootstrapArgs.force, "force", "f", false, "overwrite existing Flux installation on the cluster")
```
stefanprodan, line 134: Please remove the short arg to avoid confusion of `-f` file.

cmd/flux/cluster_info.go
```diff
@@ -83,5 +93,30 @@ func getFluxClusterInfo(ctx context.Context, c client.Client) (fluxClusterInfo,
...
+	var overridePrompt = `Flux %[1]s has been installed on this cluster with %[2]s!
+Are you sure you want to proceed with overriding the %[2]s installation? Y/N: `
+
+	// no need to display prompt if flux is not installed or installation is
+	// managed by Flux
+	if !info.installed || info.managedBy == "" || info.managedBy == "flux" {
+		return nil
+	}
+
+	fmt.Fprintf(os.Stdout, fmt.Sprintf(overridePrompt, info.version, info.managedBy))
```
stefanprodan, line 112: Please use promptui, example here: https://github.com/fluxcd/flux2/blob/adc04651cf1c81b8d6214273bddb77cf31b5d366/cmd/flux/uninstall.go#L63-L69

cmd/flux/install.go
```diff
@@ -201,8 +202,11 @@ func installCmdRun(cmd *cobra.Command, args []string) error {
 	}
 
 	if !installArgs.force {
-		err := confirmFluxInstallOverride(os.Stdin, info)
+		err := confirmFluxInstallOverride(info)
 		if err != nil {
+			if err == promptui.ErrAbort {
+				return fmt.Errorf("bootstrap cancelled")
```
stefanprodan, line 208:
```suggestion
				return fmt.Errorf("installation cancelled")
```

cmd/flux/cluster_info.go
```diff
@@ -97,26 +90,21 @@ func getFluxClusterInfo(ctx context.Context, c client.Client) (fluxClusterInfo,
...
 
-	fmt.Fprintf(os.Stdout, fmt.Sprintf(overridePrompt, info.version, info.managedBy))
-	ans, err := bufio.NewReader(reader).ReadString('\n')
-	if err != nil {
-		return fmt.Errorf("could not read from stdin: %w", err)
-	}
-	ans = strings.TrimRight(ans, "\r\n")
-	if ans == "Y" {
-		return nil
+	display := fmt.Sprintf("Flux %[1]s has been installed on this cluster with %s", info.version, info.managedBy)
```
stefanprodan, line 102:
```suggestion
	display := fmt.Sprintf("Flux %s has been installed on this cluster with %s!", info.version, info.managedBy)
```

cmd/flux/bootstrap_git.go
```diff
@@ -152,7 +152,7 @@ func bootstrapGitCmdRun(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("cluster info unavailable: %w", err)
 		}
 
-		err = confirmFluxInstallOverride(os.Stdin, info)
+		err = confirmFluxInstallOverride(info)
 		if err != nil {
 			return err
```
stefanprodan, line 157: Please add `if err == promptui.ErrAbort`

cmd/flux/bootstrap.go
```diff
@@ -129,6 +131,7 @@ func init() {
 
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapArgs.commitMessageAppendix, "commit-message-appendix", "", "string to add to the commit messages, e.g. '[ci skip]'")
 
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapArgs.force, "force", false, "overwrite existing Flux installation on the cluster")
```
stefanprodan, line 134:
```suggestion
	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapArgs.force, "force", false, "override existing Flux installation if it's managed by a diffrent tool such as Helm")
```

stefanprodan (approved):
LGTM

Thanks @somtochiama 🥇

cmd/flux/uninstall.go
```diff
@@ -59,24 +60,35 @@ func init() {
...
+		return err
+	}
+
 	if !uninstallArgs.dryRun && !uninstallArgs.silent {
+		info, err := getFluxClusterInfo(ctx, kubeClient)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return fmt.Errorf("cluster info unavailable: %w", err)
+			}
+		}
```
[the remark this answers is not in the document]
somtochiama: >. Where we know that flux is not installed at all if there's an error and it's not found error.

Should we consider a case where the CRDs might not be present (maybe due to a previous incomplete uninstall) but the flux-system namespace is still present? Returning early will not delete the namespace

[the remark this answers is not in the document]
stefanprodan: We should always uninstall, this saved so many users from Helm uninstall and TF delete partial failures where some objects were stuck. I'm also not for extending this PR to uninstall, let's write an issue for improving the uninstall procedure if we must, and discuss there the specs.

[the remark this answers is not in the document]
somtochiama: I think we can keep the uninstall prompt as it doesn't change any existing uninstall behaviour and only gives the user some more information. I will open a new issue for any other changes around the uninstall procedure.

PR #4380, opened 2023-11-03
Tweak permissions on created files

hiddeco:


stefanprodan (approved):

hiddeco:
Should be paired with released versions of https://github.com/fluxcd/pkg/pull/681. Want to ensure CI passes everywhere with the changes involved.

Issue #4403, opened 2023-11-14
Running `flux install anything-else` should return an error


VinGarcia:
### Describe the bug

I recently typed a wrong command when trying to update my local version of flux, I wanted to type:

```
brew install fluxcd/tap/flux
```

but I actually typed:

```
flux install fluxcd/tap/flux
```

And instead of getting an error it worked, which caused the flux-system kustomization to fail, luckly this was on a development cluster, not prod.

Running `flux help install`shows this:

```bash
$ flux help install
The install command deploys Flux in the specified namespace.
If a previous version is installed, then an in-place upgrade will be performed.

Usage:
  flux install [flags]
```

So since it doesn't expect any positional arguments I would have expected it to return an error, which would have saved me some trouble communicating to the guy that manages this (which is not me) about this issue.

### Steps to reproduce

Having any cluster with flux running on it run: `flux install anything`, it works instead of returning an error.

### Expected behavior

It should fail with an error informing that this command expects no positional arguments.

### OS / Distro

macOS 12.5

### Flux version

flux: v2.1.2

### Flux check

$ flux check
► checking prerequisites
✗ Kubernetes version v1.23.16 does not match >=1.25.0-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.36.2
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.29.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.24.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v1.1.1
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v1.1.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v1.1.2
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta2
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1
✔ helmcharts.source.toolkit.fluxcd.io/v1beta2
✔ helmreleases.helm.toolkit.fluxcd.io/v2beta1
✔ helmrepositories.source.toolkit.fluxcd.io/v1beta2
✔ imagepolicies.image.toolkit.fluxcd.io/v1beta1
[...]

PR #4404, opened 2023-11-15
Fix flux install command so it returns an error when unexpected arguments are passed

VinGarcia:
This PR was implemented to address the issue reported by myself on #4403.

I made sure that it is working on the cases where there are only flag arguments, or no arguments.

I wanted to add a test for a successful run, but when I tried I actually got an "install failed" error. So I figured that maybe there was a reason why no successful tests were present.

I can add more tests, but I would need some guidance.

### DIFF
--- a/cmd/flux/install.go
+++ b/cmd/flux/install.go
@@ -36,6 +36,7 @@ import (
 
 var installCmd = &cobra.Command{
 	Use:   "install",
+	Args:  cobra.NoArgs,
 	Short: "Install or upgrade Flux",
 	Long: `The install command deploys Flux in the specified namespace.
 If a previous version is installed, then an in-place upgrade will be performed.`,
--- a/cmd/flux/install_test.go
+++ b/cmd/flux/install_test.go
@@ -37,6 +37,11 @@ func TestInstall(t *testing.T) {
 			args:   "install --namespace='@#[]'",
 			assert: assertError("namespace must be a valid DNS label: \"@#[]\""),
 		},
+		{
+			name:   "invalid sub-command",
+			args:   "install unexpectedPosArg --namespace=example",
+			assert: assertError(`unknown command "unexpectedPosArg" for "flux install"`),
+		},
 	}
 
 	for _, tt := range tests {


cmd/flux/install.go
```diff
@@ -52,7 +53,7 @@ If a previous version is installed, then an in-place upgrade will be performed.`
   flux install --toleration-keys=node.kubernetes.io/dedicated-to-flux
 
   # Dry-run install
-  flux install --export | kubectl apply --dry-run=client -f- 
+  flux install --export | kubectl apply --dry-run=client -f-
```
VinGarcia, line 56:
```suggestion
  flux install --export | kubectl apply --dry-run=client -f- 
```

makkes (changes requested):
There's a much simpler solution here by using the `Args` field of `Command`:

```go
var installCmd = &cobra.Command{
	Args: cobra.NoArgs,
```

VinGarcia:
@makkes just implemented your suggestion, thanks =]

stefanprodan:
@VinGarcia can you please squash all commits into one and rebase.

cmd/flux/install_test.go
```diff
@@ -37,6 +37,11 @@ func TestInstall(t *testing.T) {
 			args:   "install --namespace='@#[]'",
 			assert: assertError("namespace must be a valid DNS label: \"@#[]\""),
 		},
+		{
+			name:   "invalid namespace",
```
makkes, line 41:
```suggestion
			name:   "invalid sub-command",
```

VinGarcia:
@stefanprodan and @makkes just squashed the commits, it should be ready now.

makkes (approved):

makkes:
One thing that bothers me a little here is that this PR only fixes this for the `install` command but all the others stay the same. Running `flux check foobar` still succeeds while `flux install foobar` fails. I personally would like to strive for consistency here. @stefanprodan wdyt?

VinGarcia:
@makkes I could add these changes on either this PR on a separate one. I would just need the list of commands on which to do it.

VinGarcia:
@makkes and @stefanprodan how should we proceed?

makkes:
Let's add a proper `Args` setting to each command.

VinGarcia:
@makkes I started going over all instances of `cobra.Command` and checking `flux [command] --help` for each of them, the ones with no args I will add `Args: cobra.NoArgs`, but some of them like `cmd/flux/bootstrap_bitbucket_server.go` do not have unit tests, so I have three options:

1. Not add unit tests at all
2. Only add tests to the files that already have unit tests
3. Write unit tests for all files even the ones that don't have `_test.go` files already.

The last option might be hard for me, and I don't know how much important you guys think 2. is.

makkes:
There is a certain chance that adding the wrong PositionalArgs to a command would result in the command not being usable, anymore, e.g. when adding `cobra.NoArgs` to the `get ks` command. I suggest it is desirable writing a unit test that verifies this doesn't happen.

I agree with Stefan that adding PositionalArgs to each command should be made in a subsequent PR and we keep this one focussed on what's said in the title.

makkes (approved):

makkes:
Thanks @VinGarcia! 🤜🏻🤛🏻

PR #4528, opened 2024-01-05
[RFC] Custom Health Checks for Kustomization using Common Expression Language(CEL)

souleb:
This RFC proposes to add support for custom heath checks during the health check phase of a `kustomization` reconciliation using `CEL` expressions.

### DIFF
--- a/rfcs/0000-custom-health-checks/README.md
+++ b/rfcs/0000-custom-health-checks/README.md
@@ -1,0 +1,330 @@
+# RFC-0000 Custom Health Checks for Kustomization using Common Expression Language(CEL)
+
+**Status:** provisional
+
+**Creation date:** 2024-01-05
+
+**Last update:** 2024-01-05
+
+## Summary
+
+This RFC proposes to support customization of the status readers in `Kustomizations`
+during the `healthCheck` phase for custom resources. The user will be able to declare
+the needed `conditions` in order to compute a custom resource status.
+In order to provide flexibility, we propose to use `CEL` expressions to declare 
+the expected conditions and their status.
+This will introduce a new field `customHealthChecks` in the `Kustomization` CRD
+which will be a list of `CustomHealthCheck` objects.
+
+## Motivation
+
+Flux uses the `Kstatus` library during the `healthCheck` phase to compute owned 
+resources status. This works just fine for all standard resources and custom resources
+that comply with `Kstatus` interfaces.
+
+In the current Kustomization implementation, we have addressed such a problem for
+kubernetes Jobs. We have implemented a `customJobStatusReader` that computes the
+status of a Job based on a defined set of conditions. This is a good solution for
+Jobs, but it is not generic and thus not applicable to other custom resources.
+
+Another use case is relying on non-standard `conditions` to compute the status of
+a custom resource. For example, we might want to compute the status of a custom
+resource based on a condtion other then `Ready`. This is the case for `Resources`
+that do intermediate patching like `Certificate` where you should look at the `Issued`
+condition to know if the certificate has been issued or not before looking at the
+`Ready` condition.
+
+In order to provide a generic solution for custom resources, that would not imply
+writing a custom status reader for each new custom resource, we need to provide a
+way for the user to express the `conditions` that need to be met in order to compute
+the status of a given custom resource. And we need to do this in a way that is
+flexible enough to cover all possible use cases, without having to change `Flux`
+source code for each new use case.
+
+### Goals
+
+- provide a generic solution for user to customize the health check of custom resources
+- support non-standard resources in `kustomize-controller`
+
+### Non-Goals
+
+- We do not plan to support custom `healthChecks` for core resources.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


souleb:
cc @tmmorin

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,330 @@
...
+// The CEL expressions must evaluate to a boolean value. The expressions are used
+// to determine the status of the custom resource.
+type CustomHealthCheckExprs struct {
+	// apiVersion of the custom health check.
+	// +required
+	APIVersion string `json:"apiVersion"`
+	// Kind of the custom health check.
+	// +required
+	Kind string `json:"kind"`
+	// InProgress is the CEL expression that verifies that the status
```
tmmorin, line 203: I'm thinking that it would be nice to here have a `Name` field, to allow definitng a healthCheck applicable to a specific resource, with a finer grained granularity than the CRD.

The kind of scenario where this would be relevant would be situations where a Kustomization would produce multiple resources of the same kind but the right healthcheck to do would be different among those resources.

An optional `Name` field would be interpreted as: the healthCheck to use for a resource of Kind Y and Name X, would be the item in CustomHealthCheckExprs that specifies Kind Y and Name X if any, or if there is none an item for Kind Y and Name unspecified.

tmmorin:
Thanks for starting this RFC, which looks very promising. :+1:

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,330 @@
...
+source code for each new use case.
+
+### Goals
+
+- provide a generic solution for user to customize the health check of custom resources
+- support non-standard resources in `kustomize-controller`
+
+### Non-Goals
+
+- We do not plan to support custom `healthChecks` for core resources.
```
tmmorin, line 51: I'm curious why this restriction ?

It seems perhaps arbitrary to not allow custom heathChecks for core resources. For instance for a Deployment, depending on the situation we might sometimes want ready to mean "all pods are ready", or "at least one pod is ready", or "at least minAvailable pods are ready".

rfcs/0000-custom-health-checks/README.md
tmmorin: I will gladly admit not having any exact use case in mind.

The idea expressed [here](https://github.com/fluxcd/flux2/pull/4528/files#r1463040255) might be one, though, but whether or not it's relevant is certainly debatable.  The intersection of "I want this Deployment to be healhcheck'd in a specific way" and "my Kustomization produces multiple Deploments" is perhaps empty.

From a user standpoint, it simply feels like "if this is going to be extensible, why not make it fully flexible".  But from an implementation and code maintenance standpoint, I would understand that this idea might  be perceived as over-engineering.

I just wanted to share this to see how relevant and implementable the idea might be.

rfcs/0000-custom-health-checks/README.md
stefanprodan: The reason for this is that kstatus covers all the bultin types, excluding them on a per kind bases is not possible, so we would need to drop them all and make users write hundreds of lines of CEL.

stefanprodan: The scope of this RFC is to allow defining health checks for custom resources. Once this is implemented, we could consider expanding CEL to native resources in a followup RFC, if it deems to be feasible and if Flux users will ask for it. So far, no one asked for more than what kstatus does for builtin kinds.

souleb: To expand on Stefan answer, the default poller uses a slice of readers:

```go
  statusReaders = append(statusReaders,
    deploymentStatusReader,
    statefulSetStatusReader,
    replicaSetStatusReader,
    defaultStatusReader,
  )
```

There is a `Support()` method that will be used to find a reader that support a given objet `Kind`. The `defaultStatusReader` is the one that is used for all custom resources. It is able to compute the status if the CR is compliant with `Kstatus`. The actual proposal is to expand this list. So in order to support `native kinds` we would need to remove the other Readers.

Another thing to consider is the `SSA`. Objects are applied in stages, and we perform a normalization of all `native kinds` mainly working around various upstream Kubernetes API bugs. This implies that we expect a certain behaviour. Also normalization could invalidate a `CEL` expression as we pass the whole object to the evaluator.

rfcs/0000-custom-health-checks/README.md
tmmorin: sorry for the slow feedback

yes @souleb there would be workarounds

and this is overall a very useful evolution to what we have today which works pretty well for many cases already, so the wise path is of course, to keep it simple

rfcs/0000-custom-health-checks/README.md
tmmorin: +1 to keeping the thing simple

my ideas of refinements may prove relevant later... or not ;)

stefanprodan:
Closing in favour of: #5151

PR #4557, opened 2024-01-23
`flux stats` should error on unrecognised args

twinguy:
flux cli does not properly detect unexpected argument on `flux stats` command.


stefanprodan (approved):
LGTM

PR #4558, opened 2024-01-23
`flux check` should error on unrecognised args

twinguy:
flux cli does not properly detect unexpected argument on `flux check` command.

### DIFF
--- a/cmd/flux/check.go
+++ b/cmd/flux/check.go
@@ -40,6 +40,7 @@ import (
 
 var checkCmd = &cobra.Command{
 	Use:   "check",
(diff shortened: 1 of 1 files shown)


stefanprodan:
@twinguy can you please rebase

stefanprodan:
@twinguy please rebase instead of merge, you need to do it with the git CLI and force push so it's only one commit.

twinguy:
@stefanprodan can you please confirm this rebase is correct?  If not, I can just close this PR and open a new one since the changes are minor.  Thanks!!

stefanprodan (approved):
LGTM

Issue #4621, opened 2024-02-21
Generate image pull secret at bootstrap


stefanprodan:
To simplify the [air-gapped bootstrap](https://fluxcd.io/flux/installation/configuration/air-gapped/) we could generate the image pull secret in the same way we do for the Git secret.

Example:
```sh
flux bootstrap git \
  --registry=registry.internal/fluxcd \
  --registry-creds="username:password" \
  --image-pull-secret="regcred"
```

When both the `--registry-creds` and `--image-pull-secret` are specified, we'll create or update the secret using the specified credentials. If `--registry-creds` is specified without `--image-pull-secret`, we'll error out and ask the user to supply the secret name.

The `flux bootstrap --registry-creds` input format is the same as for `flux push artifact --creds`, we can reuse the parsing logic and validation from `--creds`.

PR #4706, opened 2024-04-05
Add `--registry-creds` flag to bootstrap and install commands

stefanprodan:
Add an optional flag called `--registry-creds` to the `flux bootstrap` and `flux install` commands for generating an image pull secret for container images stored in private registries.

TODO (after merge):
- Update the air-gapped installation guide https://fluxcd.io/flux/installation/configuration/air-gapped/
- Expose the new option in Flux Terraform provider https://github.com/fluxcd/terraform-provider-flux/blob/8733e7566bd505616fc9562988f7196607ade593/internal/provider/resource_bootstrap_git.go#L90

Closes: #4621

### DIFF
--- a/pkg/bootstrap/bootstrap_plain_git.go
+++ b/pkg/bootstrap/bootstrap_plain_git.go
@@ -207,6 +207,14 @@ func (b *PlainGitBootstrapper) ReconcileComponents(ctx context.Context, manifest
 		b.logger.Successf("installed components")
 	}
 
+	// Reconcile image pull secret if needed
+	if options.ImagePullSecret != "" && options.RegistryCredential != "" {
+		if err := reconcileImagePullSecret(ctx, b.kube, options); err != nil {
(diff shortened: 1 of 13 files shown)


pkg/bootstrap/bootstrap_plain_git.go
```diff
@@ -207,6 +207,14 @@ func (b *PlainGitBootstrapper) ReconcileComponents(ctx context.Context, manifest
 		b.logger.Successf("installed components")
 	}
 
+	// Reconcile image pull secret if needed
+	if options.ImagePullSecret != "" && options.RegistryCredential != "" {
```
souleb, line 211: shouldn't this step be done before applying the manifest?

stefanprodan: No, because there is no namespace to deploy to.

souleb (approved):
LGTM!

Issue #4694, opened 2024-03-26
Option to error on missing postBuild substitution variable


sjdrc:
### Describe the bug

I'm currently trying to convert some helmfiles into flux manifests.

We currently use [environment values](https://helmfile.readthedocs.io/en/latest/#environment-values) to allow us to configure an entire application stack, and allow users to install everything according to our configuration but with setting their own required inputs. As helmfile uses gotemplates for generating the final manifests we can do logic such as requiring environment values be set (with `{{ required "You must provide a value for someValue!" .Values.someValue }}`).

It looks like flux [post build variable substitution](https://fluxcd.io/flux/components/kustomize/kustomizations/#post-build-variable-substitution) will do the trick for us, and the bash string replacement functions are useful, but I'm quite surprised that by default an unset variable will just evaluate to an empty string. I want to set up flux manifests that require these values to be set and returns an error if you do not.

The kubernetes cluster-api project have [an open issue](https://github.com/kubernetes-sigs/cluster-api/issues/8325) discussing switching to an alternative implementation of envsubst, but there is not feature parity between the two projects.

Would it be possible to add an option to flux to error if any of the variables are not set? (Or evaluate to an empty string)

### Steps to reproduce

N/A

### Expected behavior

Able to enforce users to provide values for postBuild.substitute

### OS / Distro

N/A

### Flux version

N/A

### Flux check

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
> Would you consider switching to use this package instead?

Let's see if @tamalsaha is keen on maintaining it going forward.

tamalsaha:
Yes, we maintain that project and use it internally.

stefanprodan:
@sjdrc we could switch to `gomodules/envsubst` and add a [feature flag](https://github.com/fluxcd/kustomize-controller/blob/main/internal/features/features.go) to kustomize-controller that when set, would break the reconciliation when vars are missing.  We also need to look into what other changes are in `gomodules/envsubst` and make sure they don't change the behaviour. Given that kustomize-controller is GA, any behaviour changes to substitutions must be opt-in.

sjdrc:
That would be excellent and would fulfill my use case perfectly!

stefanprodan:
@sjdrc can you help implementing this?

sjdrc:
@stefanprodan 
I'm not too familiar with Go, but I'm happy to give it a try over the long weekend!

[...]

PR #4709, opened 2024-04-08
Add `--strict-substitute` flag to `flux build ks` and `flux diff ks` 

stefanprodan:
Allow setting strict mode for post build substitutions in `flux build ks` and `flux diff ks` commands.

Refs:
- https://github.com/fluxcd/pkg/pull/759
- https://github.com/fluxcd/flux2/issues/4694


souleb (approved):
LGTM!

PR #4717, opened 2024-04-11
Set `GOMAXPROCS` and `GOMEMLIMIT` to all Flux controllers

hawwwdi:
if applied, set `GOMAXPROCS` and `GOMEMLIMIT` for all controllers based on container resources which leads to reduce throttling and better performance.

ref: https://github.com/fluxcd/source-controller/pull/1443
ref: https://github.com/fluxcd/flux-benchmark/pull/5


stefanprodan:
@hawwwdi have you've seen any improvements to CPU throttling on your cluster?

stefanprodan (approved):
LGTM

Thanks @hawwwdi

Issue #4692, opened 2024-03-25
Add a subcommand to create a notation secret


souleb:
Update the Flux CLI subcommand `flux create secret` to generate a generate a `secret` containing `notation` configuration that can be used to verify a source.

This is a follow-up to https://github.com/fluxcd/source-controller/pull/1075.

JasonTheDeveloper:
Happy to pick this up, if you haven't ready started on it, @souleb

souleb:
hello @JasonTheDeveloper what is your plan for this issue?

JasonTheDeveloper:
Hey @souleb, I've made some progress on this. Just have to write a unit tests to cover the changes introduced before creating the PR. You can see what I've done in [feat/4692](https://github.com/JasonTheDeveloper/flux2/tree/feat/4692).

Basically, in short, I've introduced a new `flux create secret notation` command that requires two arguments, a `--trust-policy-file` and a `--ca-cert-file`.  `--ca-cert-file` is a slice and can accept multiple file paths. I've also added support for directories so that the user can pass in a folder and left flux walk the directory. 

I've added validations to check the trust policy is a valid notation policy as well as checking the certs passed end in `.pem` or `.crt`.

Because Notation can handle multiple certs in one secret, I had to update `manifestgen/sourcesecret.options` to handle multiple certs and their file names.

I didn't have time today to add those unit tests but I aim to do that tomorrow before creating the PR.

souleb:
Awesome. Thank you!

souleb:
I have updated our docs, but it depends on this feature. See: https://github.com/fluxcd/website/pull/1911. It would be nice if you could review it too.

stefanprodan:
Implemented in #4735

PR #4735, opened 2024-04-21
feat(secret): add create notation secret handler

JasonTheDeveloper:
This PR is in relation to #4692. 

The purpose of this PR is to aid in the creation of notation configuration secret used to validate artifacts signed by notation. The `create secret notation` will only add root CA certificates ending in `.crt` and `.pem` and valid `trust-policy.json`. You can pass in multiple root CA certificates as well as a folder containing multiple certs.


cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+	Use:   "notation [name]",
+	Short: "Create or update a Kubernetes secret for verifications of artifacts signed by Notation",
+	Long:  withPreviewNote(`The create secret notation command generates a Kubernetes secret with root ca certificates and trust policy.`),
+	Example: ` # Create a Notation configuration secret on disk and encrypt it with Mozilla SOPS
+  flux create secret notation config \
+    --namespace=my-namespace \
+    --trust-policy-file=./my-trust-policy.json \
+    --ca-cert-file=my-./my-cert.crt \
+	--ca-cert-file=my-./my-cert.pem \
+    --export > repo-auth.yaml
```
stefanprodan, line 45:
```suggestion
  flux create secret notation my-notation-cert \
    --namespace=my-namespace \
    --trust-policy-file=./my-trust-policy.json \
    --ca-cert-file=./my-cert.crt \
    --export > my-notation-cert.yaml
```

If Flux doesn't need the pem, then let's leave it out.

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+
+  sops --encrypt --encrypted-regex '^(data|stringData)$' \
+    --in-place repo-auth.yaml`,
```
stefanprodan, line 48:
```suggestion
  sops --encrypt --encrypted-regex '^(data|stringData)$' \
    --in-place my-notation-cert.yaml`,
```

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+	createSecretCmd.AddCommand(createSecretNotationCmd)
+}
+
+func createSecretNotationCmdRun(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("name is required")
+	}
+
+	if secretNotationArgs.caCrtFile == nil || len(secretNotationArgs.caCrtFile) == 0 {
+		return fmt.Errorf("--ca-cert-file cannot be empty")
```
souleb, line 73:
```suggestion
		return fmt.Errorf("--ca-cert-file is required")
```

In order to be consistent with other errors

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if secretNotationArgs.trustPolicyFile == "" {
+		return fmt.Errorf("--trust-policy-file cannot be empty")
```
souleb, line 77:
```suggestion
		return fmt.Errorf("--trust-policy-file is required")
```

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	var doc trustpolicy.Document
+
+	if err := json.Unmarshal(policy, &doc); err != nil {
+		return fmt.Errorf("error occurred while parsing %s: %w", secretNotationArgs.trustPolicyFile, err)
+	}
+
+	if err := doc.Validate(); err != nil {
+		return fmt.Errorf("trust policy validation failed: %w", err)
```
souleb, line 99:
```suggestion
		return fmt.Errorf("invalid trust policy: %w", err)
```

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+
+	policy, err := os.ReadFile(secretNotationArgs.trustPolicyFile)
+	if err != nil {
+		return fmt.Errorf("unable to read trust policy file: %w", err)
+	}
+
+	var doc trustpolicy.Document
+
+	if err := json.Unmarshal(policy, &doc); err != nil {
+		return fmt.Errorf("error occurred while parsing %s: %w", secretNotationArgs.trustPolicyFile, err)
```
souleb, line 95:
```suggestion
		return fmt.Errorf("failed to unmarshal trust policy %s: %w", secretNotationArgs.trustPolicyFile, err)
```

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if err := doc.Validate(); err != nil {
+		return fmt.Errorf("trust policy validation failed: %w", err)
+	}
+
+	caCerts := []sourcesecret.VerificationCrt{}
+	for _, caCrtFile := range secretNotationArgs.caCrtFile {
+		isDir, err := isDirectory(caCrtFile)
```
souleb, line 104: I think we should accept a list of files. Otherwise we would have to name it `caCrtPath`.

cmd/flux/create_secret_notation.go
JasonTheDeveloper: On this I'm in two minds about it. In [source-controller](https://github.com/fluxcd/source-controller/blob/014365d0e13ecee62174f67bcaf4724fa65f8cc2/internal/oci/notation/notation.go#L201-L217) we check for verification level `skip` and ensure `TrustStores` and `TrustedIdentities` are empty.  In situations where `TrustStores` and `TrustedIdentities` are populated and `VerificationLevel` is set to `skip`, `doc.Validate()` will return an error.

We do check the `trust-policy.json` is a valid json which might be enough but at the same time it might be nice to also ensure the trust policy we're supplying is actually valid. Perhaps adding a flag to fail if the validation returns an error otherwise log it?

What do you think, @souleb?

cmd/flux/create_secret_notation.go
JasonTheDeveloper: I've made it so that `--ca-cert-file` can take in a list of files (e.g. `--ca-cert-file=test-ca-1.crt,test-ca-2.crt` or `--ca-cert-file=test-ca-1.crt --ca-cert-file=test-ca-2.crt`) or a path to the parent directory so that if you have multiple certs, you can just pass the directory.

Perhaps it might be better to rename `--ca-cert-file` to `--ca-cert-path` and use `caCrtPath` instead of `caCrtFile`. Thoughts?

cmd/flux/create_secret_notation.go
JasonTheDeveloper: Resolved in [9160c2a](https://github.com/fluxcd/flux2/pull/4735/commits/9160c2aac0be213fe3d1e774301c58343d14b8a5).

cmd/flux/create_secret_notation.go
JasonTheDeveloper: Resolved in [9160c2a](https://github.com/fluxcd/flux2/pull/4735/commits/9160c2aac0be213fe3d1e774301c58343d14b8a5).

cmd/flux/create_secret_notation.go
stefanprodan: We don't have this dir scanning feature in any command in the Flux CLI. I propose we remove it and only allow pointing to files.

cmd/flux/create_secret_notation.go
souleb: I think we should return the error as is. Diverging from `notation` in any way could be a pain in the future. The user should have the same experience that it would get with `notation`.

Also This is just a helper. The users that want to switch back to `skip` with existing `TrustStores` and `TrustedIdentities` can do it manually.

JasonTheDeveloper: Works for me! Resolved in [9160c2a](https://github.com/fluxcd/flux2/pull/4735/commits/9160c2aac0be213fe3d1e774301c58343d14b8a5).

cmd/flux/create_secret_notation.go
JasonTheDeveloper: Works for me. Removed in [8397e0d](https://github.com/fluxcd/flux2/pull/4735/commits/8397e0d4eff530ac52479903ef6c8338cf7ca37a).

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if err := doc.Validate(); err != nil {
+		return fmt.Errorf("invalid trust policy: %w", err)
+	}
+
+	caCerts := []sourcesecret.VerificationCrt{}
+	for _, caCrtFile := range secretNotationArgs.caCrtFile {
+		fileName := filepath.Base(caCrtFile)
+		if !strings.HasSuffix(fileName, ".crt") && !strings.HasSuffix(fileName, ".pem") {
```
souleb, line 104: we could do something like

```golang
	var res error
	for _, caCrtFile := range secretNotationArgs.caCrtFile {
		fileName := filepath.Base(caCrtFile)
		if !strings.HasSuffix(fileName, ".crt") && !strings.HasSuffix(fileName, ".pem") {
			res = errors.Join(res, fmt.Errorf("%s must end with either .crt or .pem", fileName))
			continue
		}
		caBundle, err := os.ReadFile(caCrtFile)
		if err != nil {
			res = errors.Join(res, fmt.Errorf("unable to read TLS CA file: %w", err))
			continue
		}
		caCerts = append(caCerts, sourcesecret.VerificationCrt{Name: fileName, CACrt: caBundle})
	}

	if res != nil {
		return res
	}
```

This way, we return all the invalid file extensions at once.

JasonTheDeveloper: Love it! Excellent idea, @souleb! Updated in [163da9f](https://github.com/fluxcd/flux2/pull/4735/commits/163da9fc9871b24e3a5218b7400a8ede5ab945a5).

souleb:
can you please squash your commits?

cmd/flux/create_secret_notation.go
```diff
@@ -0,0 +1,159 @@
...
+	if err := json.Unmarshal(policy, &doc); err != nil {
+		return fmt.Errorf("failed to unmarshal trust policy %s: %w", secretNotationArgs.trustPolicyFile, err)
+	}
+
+	if err := doc.Validate(); err != nil {
+		return fmt.Errorf("invalid trust policy: %w", err)
+	}
+
+	caCerts := []sourcesecret.VerificationCrt{}
+	var res error
```
souleb, line 103: nit:
```suggestion
	var (
		caCerts []sourcesecret.VerificationCrt
		fileErr error
	)
```

JasonTheDeveloper:
@stefanprodan @souleb all commits have been squashed.

souleb (approved):
LGTM! thanks @JasonTheDeveloper

stefanprodan (approved):

PR #4746, opened 2024-04-25
Specifying go version in setup-go github action.

swade1987:
Ref: https://x.com/stefanprodan/status/1780966643201388991


stefanprodan (approved):
LGTM

Thanks @swade1987

PR #4749, opened 2024-04-27
[RFC] Flux Bootstrap for OCI-compliant Container Registries

stefanprodan:
Proposal for a Git-less bootstrap using OCI-compliant Container Registries as the desired state storage.

Spec preview: https://github.com/fluxcd/flux2/blob/rfc-flux-bootstrap-oci/rfcs/000X-flux-bootstrap-oci/README.md

Example:

```shell
flux bootstrap oci \
--url=oci://ghcr.io/stefanprodan/flux-manifests:production \
--username=stefanprodan \
--password=$GITHUB_TOKEN \
--kustomization=flux-manifests/kustomization.yaml \
--cluster-url=oci://ghcr.io/stefanprodan/fleet-manifests:production \
--cluster-path=clusters/production
```

### DIFF
--- a/rfcs/000X-flux-bootstrap-oci/README.md
+++ b/rfcs/000X-flux-bootstrap-oci/README.md
@@ -54,6 +54,21 @@ flux bootstrap oci \
 --cluster-path=<path/inside/oci/artifact>
 ```
 
+The bootstrap shares the following flags with the `flux install` command:
+
+```text
+--cluster-domain string      internal cluster domain (default "cluster.local")
+--components strings         list of components, accepts comma-separated values (default [source-controller,kustomize-controller,helm-controller,notification-controller])
+--components-extra strings   list of components in addition to those supplied or defaulted, accepts values such as 'image-reflector-controller,image-automation-controller'
+--image-pull-secret string   Kubernetes secret name used for pulling the toolkit images from a private registry
+--log-level logLevel         log level, available options are: (debug, info, error) (default info)
+--network-policy             deny ingress access to the toolkit controllers from other namespaces using network policies (default true)
+--registry string            container registry where the toolkit images are published (default "ghcr.io/fluxcd")
+--toleration-keys strings    list of toleration keys used to schedule the components pods onto nodes with matching taints
+--version string             toolkit version, when specified the manifests are downloaded from https://github.com/fluxcd/flux2/releases
+--watch-all-namespaces       watch for custom resources in all namespaces, if set to false it will only watch the namespace where the toolkit is installed (default true)
+```
+
 The Terraform/OpenTofu counterpart is the `flux_bootstrap_oci` provider that exposes
 the same configuration options as the CLI.
 
--- a/rfcs/000X-flux-bootstrap-oci/README.md
+++ b/rfcs/000X-flux-bootstrap-oci/README.md
@@ -1,0 +1,319 @@
+# [RFC] Flux Bootstrap for OCI-compliant Container Registries
+
+**Status:** provisional
+
+**Creation date:** 2024-04-27
+
+**Last update:** 2024-04-27
+
+## Summary
+
+Flux should allow a Git-less bootstrap procedure where the cluster desired state is stored in OCI artifacts.
+
+On the client-side, the Flux CLI should offer a command for packaging its own Kubernetes manifests into
+an OCI artifact and pushing the artifact to a container registry.
+
+On the server-side, the Flux controllers should be configured to self-update from the registry
+and reconcile the cluster state from OCI artifacts stored in the same or a different registry.
+
+## Motivation
+
+Given that OCI registries are evolving into a generic artifact storage solution,
+we should allow Flux users who don't want to run a Git server as part of their
+production infrastructure to bootstrap and manage their Kubernetes clusters using OCI artifacts.
+
+To decouple the clusters reconciliation from the Git repositories, Flux allows packaging and publishing
+the Kubernetes manifests stored in Git to an OCI registry by running the `flux push artifact`
+command in CI pipelines.
+
+### Goals
+
+- Add support to the Flux CLI for bootstrapping with a container registry as the source of truth.
+- Make it easy for users to switch from Git repositories to OCI repositories.
+
+### Non-Goals
+
+- Automate the migration of Flux manifests from a Git bootstrap repository to OCI.
+
+## Proposal
+
+Implement the `flux bootstrap oci` command with the following specifications:
+
+```shell
+flux bootstrap oci \
+--url=<registry-url>/<flux-manifests>:<tag> \
+--username=<registry-username> \
+--password=<registry-password> \
+--kustomization=<local/path/to/kustomization.yaml> \
+--cluster-url=<registry-url>/<fleet-manifests>:<tag> \
+--cluster-path=<path/inside/oci/artifact>
+```
+
+The Terraform/OpenTofu counterpart is the `flux_bootstrap_oci` provider that exposes
+the same configuration options as the CLI.
+
+The bootstrap operations are split into two phases:
+
+- Install and self-update configuration for the Flux components.
+- Cluster state reconciliation configuration.
+
+### Install and self-update configuration
+
+The command performs the following steps based on the `url`, `username`,
+`password` and `kustomization` arguments:
+
+1. Logs in to the OCI registry using the provided credentials.
+2. Generates an OCI artifact from the Flux components manifests and the `kustomization.yaml` file.
+3. Applies the Flux components manifests along with their customisations to the cluster.
+4. Pushes the OCI artifact to the container registry using the specified tag.
+5. Generates an image pull secret, an OCIRepository that points to the OCI artifact and
+   a Flux Kustomization object that reconciles the OCI artifact contents.
+6. Applies the image pull secret, OCIRepository and Flux Kustomization to the cluster.
+
+Artifacts pushed to the registry:
+- `<registry-url>/<flux-manifests>:<checksum>` (immutable artifact)
+- `<registry-url>/<flux-manifests>:<tag>` (tag pointing to the immutable artifact)
+
+Objects created by the command in the `flux-system` namespace:
+- `flux-components` Secret
+- `flux-components` OCIRepository
+- `flux-components` Kustomization
+
+### Cluster state reconciliation configuration
+
+After the OCIRepository and Flux Kustomization called `flux` become ready, the command
+continues with the following steps:
+
+1. Logs in to the OCI registry where the cluster artifacts are stored using the provided credentials.
+2. If the cluster OCI artifact is not found, an empty artifact is created
+   and pushed to the registry using the provided tag.
+3. Generates an image pull secret, an OCIRepository and a Flux Kustomization object
+   that reconciles the cluster OCI artifact contents.
+4. Applies the image pull secret, OCIRepository and Flux Kustomization to the cluster.
+
+Objects created by the command in the `flux-system` namespace:
+- `flux-system` Secret
+- `flux-system` OCIRepository
+- `flux-system` Kustomization
+
+If the cluster registry is the same as the Flux components registry, the command could reuse the
+`flux-components` image pull secret.
+
+### Registry authentication
+
+The `flux bootstrap oci` command supports the following authentication methods:
+
+- Basic authentication with `--username` and `--password`. The credentials are stored in a Kubernetes Secret.
+- OIDC authentication with `--provider=<aws|azure|gcp>`. No credentials are stored in the cluster, source-controller
+  will use Kubernetes Workload Identity to authenticate to the registry.
+
+To avoid passing the credentials as CLI flags, the password can be read from the standard input, e.g.:
+`echo <password> | flux bootstrap oci` or using an environment variable `OCI_PASSWORD`.
+
+If the registry is self-hosted and uses a self-signed TLS certificate,
+the root CA certificate can be provided with the `--ca-file` flag.
+
+If the registry is exposed on HTTP and not HTTPS, the `--allow-insecure-http`
+flag can be used to force non-TLS connections.
+
+### Signing and verification
+
+The `flux bootstrap oci` command supports the following signing and verification methods:
+
+- Cosign
+- Notation
+
+TODO: Add more details about the signing and verification methods, flags and options.
+
+### User Stories
+
+#### Story 1
+
+> As a platform operator I want to bootstrap a Kubernetes cluster with Flux
+> using OCI artifacts stored in a container registry.
+
+The following example demonstrates how to bootstrap a Flux instance using GitHub Container Registry
+as the OCI registry for Flux components and the cluster state.
+
+```shell
+flux bootstrap oci \
+--url=ghcr.io/stefanprodan/flux-manifests:production \
+--username=<ghcr-username> \
+--password=<ghcr-token> \
+--kustomization=flux-manifests/kustomization.yaml \
+--cluster-url=ghcr.io/stefanprodan/fleet-manifests:production \
+--cluster-username=<ghcr-username> \
+--cluster-password=<ghcr-token> \
+--cluster-path=clusters/production
+```
+
+Generated OCI artifacts:
+
+- `ghcr.io/stefanprodan/flux-manifests:88b028f`
+- `ghcr.io/stefanprodan/flux-manifests:production`
+- `ghcr.io/stefanprodan/fleet-manifests:6f7a258`
+- `ghcr.io/stefanprodan/fleet-manifests:production`
+
+Objects created in the `flux-system` namespace:
+
+Flux components reconciliation:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: flux-components
+  namespace: flux-system
+spec:
+  interval: 1m
+  url: oci://ghcr.io/stefanprodan/flux-manifests
+  ref:
+    tag: production
+  secretRef:
+    name: flux-components
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: flux-components
+  namespace: flux-system
+spec:
+    interval: 1h
+    retryInterval: 5m
+    sourceRef:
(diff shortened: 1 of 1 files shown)


rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+### Install and self-update configuration
+
+The command performs the following steps based on the `url`, `username`,
+`password` and `kustomization` arguments:
+
+1. Logs in to the OCI registry using the provided credentials.
+2. Generates an OCI artifact from the Flux components manifests and the `kustomization.yaml` file.
+3. Applies the Flux components manifests along with their customisations to the cluster.
+4. Pushes the OCI artifact to the container registry using the specified tag.
+5. Generates an image pull secret, an OCIRepository that points to the OCI artifact and
```
errordeveloper, line 69:
```suggestion
5. Optionally, generates an image pull secret, an OCIRepository that points to the OCI artifact and
```

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+- Install and self-update configuration for the Flux components.
+- Cluster state reconciliation configuration.
+
+### Install and self-update configuration
+
+The command performs the following steps based on the `url`, `username`,
+`password` and `kustomization` arguments:
+
+1. Logs in to the OCI registry using the provided credentials.
+2. Generates an OCI artifact from the Flux components manifests and the `kustomization.yaml` file.
```
errordeveloper, line 66: Might be worse mentioning here regarding workload identity mods that may be needed, e.g. on EKS role ARN needs to be set as an annotation, I forgot if that was necessary in GKE also.

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+- `<registry-url>/<flux-manifests>:<tag>` (tag pointing to the immutable artifact)
+
+Objects created by the command in the `flux-system` namespace:
+- `flux-components` Secret
+- `flux-components` OCIRepository
+- `flux-components` Kustomization
+
+### Cluster state reconciliation configuration
+
+After the OCIRepository and Flux Kustomization called `flux` become ready, the command
```
errordeveloper, line 84: Does this have to be strictly sequential and synchronous?

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+continues with the following steps:
+
+1. Logs in to the OCI registry where the cluster artifacts are stored using the provided credentials.
+2. If the cluster OCI artifact is not found, an empty artifact is created
+   and pushed to the registry using the provided tag.
+3. Generates an image pull secret, an OCIRepository and a Flux Kustomization object
```
errordeveloper, line 90:
```suggestion
3. Optionally, generates an image pull secret, an OCIRepository and a Flux Kustomization object
```

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Summary
+
+Flux should allow a Git-less bootstrap procedure where the cluster desired state is stored in OCI artifacts.
+
+On the client-side, the Flux CLI should offer a command for packaging its own Kubernetes manifests into
+an OCI artifact and pushing the artifact to a container registry.
+
+On the server-side, the Flux controllers should be configured to self-update from the registry
+and reconcile the cluster state from OCI artifacts stored in the same or a different registry.
```
errordeveloper, line 17: I would imagine that users may wish to choose between pulling from upstream OCI artifact that is published as part of the Flux release or having a full copy of it. If they choose to use a copy, another command may be needed to keep their copy up to date. Does it make sense?

errordeveloper (commented):
Thanks for drafting this, Stefan! Here are a few thoughts/suggestions from me :)

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: You would always have a copy in your registry that includes customisations, same as with Git, bootstrap means vendoring the Flux manifests that in 99% of cases would need some fine tuning.

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: We have all of those documented here https://fluxcd.io/flux/installation/configuration/workload-identity/, people will need to read the docs and adapt their kustomization.yaml.

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: It will always generate the manifests then it will use Flux SSA dry-run to determine if there is drift, and only then apply the changes. Maybe instead of `apply` should be `reconcile`.

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: Ah, that's true that in most cases right now people will end up with a copy. How do they bring the copy up to date, e.g. some component has new pod spec fields that have to be set or there is an RBAC change? Do they have to read changelog and implement such changes? If in OCI world this could be avoided by means of referencing an upstream artifact and local changes stored as a patch/kustonization, it might be very nice actually, I guess it wasn't feasible with Git as an upstream.

stefanprodan: For both the Git and OCI bootstrap, the Flux update is fully automated. See the Story 3

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: This suggestion and other suggestion in wording of steps are meant to highlight that in OCI case IAM auth eliminates the need for secrets.

stefanprodan: Ah I get it now, optionally is for the image pull secret!

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: But perhaps there could be bootstrap argument to specify provider-specific attributes that would be handled accordingly based on provider flag?

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+Download the latest CLI version and update Flux directly in the registry, without rerunning bootstrap:
+
+```shell
+# pull the latest manifests from the registry
+flux pull artifact oci://ghcr.io/stefanprodan/flux-manifests:production \
+--output=./flux-manifests
+
+# update the Flux components manifests
+flux install --export > ./flux-manifests/flux-system/gotk-components.yaml
```
errordeveloper, line 267: These are two alternative methods right? It's not very clear from the text at the moment.

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: There is `--provider=<aws|azure|gcp>` but this is for the CLI to use the role of the machine where it's running. The IAM of the bastion host may be different from the one that you want to use for Flux source-controller. So we'll need yet another flag with the identity name. This is how Flux AIO works: https://timoni.sh/flux-aio/#__tabbed_1_3

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: If you don't have access to the cluster (what the user story is about), this is the only way. If you have API access, then like with Git bootstrap, you can just rerun it to update. OCI bootstrap behaves the same as Git bootstrap.

errordeveloper: I just thought these two commands will write the same kind of output, except that install lets you select a subset of controllers... maybe I am missing something else.

errordeveloper: Is it the case right now that one has to rerun bootstrap on major/minor releases while patch releases are taken care of by in-cluster image version bumps?

errordeveloper:
> ... bootstrap means vendoring the Flux manifests that in 99% of cases would need some fine tuning.

Can this fine tuning be done with an in-cluster Kustomization, or has that been proven somehow challenging?

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: Every time we release Flux, users get an PR opened to update the manifests in Git. For OCI you would need some kind of semver range or some other manual gate e.g. a GitHub workflow dispatch to approve minor bumps and let only patch versions be automatically push to the registry.

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: > ... bootstrap means vendoring the Flux manifests that in 99% of cases would need some fine tuning.

Just want to clarify, can this fine tuning be done with an in-cluster Kustomization, or has that been proven somehow challenging?

stefanprodan: > can this fine tuning be done with an in-cluster Kustomization

Yes but up to some point. Some customisations require more than just patches e.g. [persistent storage for artifacts](https://fluxcd.io/flux/installation/configuration/vertical-scaling/#persistent-storage-for-flux-internal-artifacts).

rfcs/000X-flux-bootstrap-oci/README.md
stefanprodan: > I just thought these two commands will write the same kind of output, except that install lets you select a subset of controllers... maybe I am missing something else.

All the `flux bootstrap` commands have the same args as `flux install`, so you can pick controllers, etc with bootstrap too. If you bootstrap with `flux bootstrap oci --components --log-level=source-controller,kustomize-controller`, to update your would run `flux install --log-level=source-controller,kustomize-controller --export` in CI.

errordeveloper: I mean 'pull artifact' vs 'install --export'.

stefanprodan: You need to pull to preserve the existing kustomization.yaml and any other extra resources you may have added at bootstrap. The `install -export` command only generates the components YAML.

errordeveloper:
Ah, of course, this is a rebase ;)
Maybe text needs to make that clearer.

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: I can see now, I think difficulty start with controller selection actually, doing it with kustomize makes kind of poor UX.

Maybe embedding Timoni could solve it? :)

stefanprodan: Controller selection is already supported via flags common to all bootstrap sub commands.

stefanprodan: These are the common flags:

```
Flags:
      --cluster-domain string      internal cluster domain (default "cluster.local")
      --components strings         list of components, accepts comma-separated values (default [source-controller,kustomize-controller,helm-controller,notification-controller])
      --components-extra strings   list of components in addition to those supplied or defaulted, accepts values such as 'image-reflector-controller,image-automation-controller'
      --force                      override existing Flux installation if it's managed by a diffrent tool such as Helm
      --image-pull-secret string   Kubernetes secret name used for pulling the toolkit images from a private registry
      --log-level logLevel         log level, available options are: (debug, info, error) (default info)
      --network-policy             deny ingress access to the toolkit controllers from other namespaces using network policies (default true)
      --registry string            container registry where the toolkit images are published (default "ghcr.io/fluxcd")
      --toleration-keys strings    list of toleration keys used to schedule the components pods onto nodes with matching taints
  -v, --version string             toolkit version, when specified the manifests are downloaded from https://github.com/fluxcd/flux2/releases
      --watch-all-namespaces       watch for custom resources in all namespaces, if set to false it will only watch the namespace where the toolkit is installed (default true)
```

errordeveloper: > Controller selection is already supported via flags common to all bootstrap sub commands.

Yes, I know, what I mean is that it would be kind of less natural specify controllers with kustomize, one would probably need to select bases as there is no meaningful parameters and if or switch statements (at least last time I checked), the CLI offers are more meaningful option.
So because you need to select controllers, you start with CLI that gives you a single file that you kustomize a little, but it's all much complicated then you might have wished and defeats the purpose of putting kustomize in at this stage.

errordeveloper: CUE on the other hand could do better in all of this and you could potentially remove the need for CLI and make custom configs easier to introduce. Does what I said earlier make more sense now?

stefanprodan: The way Flux can be customized at bootstrap is all via Kustomize patches and CLI options, this must be 100% compatible with the oci sub-command. I’m not considering using CUE or anything really that would diverge from the current Git bootstrap procedure. Users should be able to migrate from Git to OCI by simply reusing their current flux-system overlay including patches, image overrides, configGenerator, volumes, etc.

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: Ah, of course, this is a rebase ;)
Maybe the text needs to make that clearer.

rfcs/000X-flux-bootstrap-oci/README.md
errordeveloper: That is for sure, I'm just thinking maybe self-managed flux in the future could use CUE for this, possibly even without exposing CUE to the user. CLI could still work the same way on the surface also. I do recall we once spoke of an installer operator too, you could use CUE there and in the CLI. Just an idea :)

rfcs/000X-flux-bootstrap-oci/README.md
userbradley: This is probably not the right place to bring this up but the workload identity for GKE is partially incorrect, you dont need to annotate service accounts any more with a GCP SA for GKE workload identity. You grant the Kubernetes service account access to what ever resources it needs via a member statement like below

```yaml
principal://iam.googleapis.com/projects/PROJECT_NUMBER/locations/global/workloadIdentityPools/PROJECT_ID.svc.id.goog/subject/ns/flux-system/sa/source-controller
```

I'm going to have a think on how I can update the docs on this one, but thought I'd raise it before I forget

rfcs/000X-flux-bootstrap-oci/README.md
```diff
@@ -0,0 +1,396 @@
...
+
+### Install and self-update configuration
+
+The command performs the following steps based on the `url`, `username`,
+`password` and `kustomization` arguments:
+
+1. Logs in to the OCI registry using the provided credentials.
+2. Generates an OCI artifact from the Flux components manifests and the `kustomization.yaml` file.
+3. Applies the Flux components manifests along with their customisations to the cluster.
+4. Pushes the OCI artifact to the container registry using the specified tag.
```
nagyv, line 88: I wonder if it would make sense to have two registry auth. A read-only for the image pull secrets, and read-write for pushing the artifacts to the registry. Storing a read-write secret in the cluster for image pull secrets does not seem like a good idea.

sestegra: It is even possible to consider more pull secrets with specific permissions (least privileges principle):
- one with a read-only permission to be used by the Deployments of the Flux controllers to pull images
- one with a read-only permission (different namespace than images) to be used by the OCIRepository resource to pull OCI artifacts
- one with write permissions to be used by the bootstrap command line

Then, it is possible to consider different OCI registry: one for images and another for Flux artifacts, because the latter could contain sensible infrastructure information.

stefanprodan: > I wonder if it would make sense to have two registry auth. A read-only for the image pull secrets, and read-write for pushing the artifacts to the registry.

Yes this is something the command could support. Currently our OCI implementation supports reading the Docker config file from the host OS, so we could use that for write operations and the flags for in-cluster secret.

@sestegra the pull secret for the container images is already supported, it's one of the common flags to all bootstrap commands.

mo-saeed:
@stefanprodan  is there any estimation when this can be released ?

rfcs/000X-flux-bootstrap-oci/README.md
thejosephstevens: We've recently adopted Flux for our multi-cloud architecture. In order to support that we're actually overriding env vars, volumes, and volume bindings directly in order to set up OIDC-based auth to each cloud on our workloads. If we can extend those parameters on the source controller, or if we can pass in a pod template that would allow us to easily inject OIDC auth for accessing the OCI backend (I suspect we could do this with the Helm install method, but support in the CLI would be nice, as the bootstrap command is great).

stefanprodan:
Closing this in favour of [Flux Operator](https://github.com/controlplaneio-fluxcd/flux-operator) and its [OCI sync implementation](https://fluxcd.control-plane.io/operator/flux-sync/#sync-from-a-container-registry).  The main difference from what I've prosed in this RFC and the approach I took with Flux Operator, is that the operator does not need write access to the container registry. The provisioning of the container registry repository, as well as the pushing and signing of artifacts are driven from CI. Flux permissions are limited to listing tags and pulling artifacts.

If you want to migrate existing clusters from `flux bootstrap` to Flux Operator and OCI Artifacts, please see the migration guide [here](https://fluxcd.control-plane.io/operator/flux-bootstrap-migration/#migration-from-git-to-oci-artifacts).

stefanprodan:
>  Does this mean that the operator will be the official way of installing flux? Coz based on the flux2 docs now, the bootstrap process is the official way to install flux.

The CLI is the official way of installing Flux, the bootstrap & install commands are GA and will be supported for as long as Flux v2 will be maintained.

Flux Operator offers an alternative to `flux bootstrap` and `flux install`, same as the community maintained Helm chart.  I'll open a PR to add it as an alternative deploy method.

Issue #4726, opened 2024-04-16
Azure DevOps: SSH-RSA is deprecated 💥 


uidmehdi:
### Describe the bug

I'm seeing errors stating that source reconciliation is failing with the following error:

`failed to checkout and determine revision: unable to list remote for 'ssh://git@ssh.dev.azure.com/v3/<organization_name>/<project_name>/<repo_name>': unknown error: remote: Command git-upload-pack: You’re using ssh-rsa that is about to be deprecated and your request has been blocked intentionally. Any SSH session using SSH-RSA is subject to brown out (failure during random time periods). Please use rsa-sha2-256 or rsa-sha2-512 instead. For more details see https://aka.ms/ado-ssh-rsa-deprecation.`

### Steps to reproduce

1. Install Flux 
2. Create source `flux create source git flux-system --url=ssh://git@ssh.dev.azure.com/v3/<organization_name>/<project_name>/<repo_name> --branch=main --ssh-key-algorithm=rsa --ssh-rsa-bits=4096 --interval=1m`
3. Add the provided deploy key in Azure DevOps SSH Public Keys
4. Wait for the brown out periods of Azure DevOps repos

### Expected behavior

Source reconciliation should succeed

### OS / Distro

Kubernetes

### Flux version

v2.2.3

### Flux check

► checking prerequisites
✔ Kubernetes 1.27.9 >=1.26.0-0
► checking version in cluster
✔ distribution: flux-v2.2.3
✔ bootstrapped: true
► checking controllers
✔ helm-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/helm-controller:v0.37.4
✔ image-automation-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/image-automation-controller:v0.37.1
✔ image-reflector-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/image-reflector-controller:v0.31.2
✔ kustomize-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/kustomize-controller:v1.2.2
✔ notification-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/notification-controller:v1.2.4
✔ source-controller: deployment ready
► digprvtstgwecr.azurecr.io/fluxcd/source-controller:v1.2.4
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta3
✔ buckets.source.toolkit.fluxcd.io/v1beta2
[...]

PR #4754, opened 2024-05-01
Add `--ssh-hostkey-algos` flag to bootstrap command

stefanprodan:
Allow configuring the list of host key algorithms to use for SSH connections initialised by the CLI during bootstrap.

This allows using `flux bootstrap git` for Azure DevOps repos over SSH.

Refs:
- https://github.com/fluxcd/flux2/issues/4726
- https://github.com/fluxcd/terraform-provider-flux/issues/678

### DIFF
--- a/cmd/flux/bootstrap.go
+++ b/cmd/flux/bootstrap.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


souleb (approved):
LGTM!

PR #4755, opened 2024-05-03
feat: expose HelmRelease IgnoreMissingValuesFiles feature

isometry:
Signed-off-by: Robin Breathe <robin@isometry.net>

### DIFF
--- a/cmd/flux/create_helmrelease.go
+++ b/cmd/flux/create_helmrelease.go
@@ -110,20 +110,21 @@ var createHelmReleaseCmd = &cobra.Command{
 }
 
 type helmReleaseFlags struct {
-	name                string
-	source              flags.HelmChartSource
-	dependsOn           []string
-	chart               string
-	chartVersion        string
-	targetNamespace     string
-	createNamespace     bool
-	valuesFiles         []string
-	valuesFrom          []string
-	saName              string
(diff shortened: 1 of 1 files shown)


cmd/flux/create_helmrelease.go
```diff
@@ -239,6 +241,9 @@ func createHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		for _, v := range helmReleaseArgs.valuesFiles {
 			data, err := os.ReadFile(v)
 			if err != nil {
+				if helmReleaseArgs.ignoreMissingValueFiles && os.IsNotExist(err) {
+					continue
+				}
```
stefanprodan, line 246: This is not what the flag should, we should map it to HelmRelease spec instead of skipping over value files passed to the CLI.

stefanprodan (changes requested):

cmd/flux/create_helmrelease.go
isometry: I think PR is probably unnecessary then. It only made sense to me if you were passing `--values-files` and `--ignore-missing-values-files` together (which we're not ourselves planning to do). Let's drop it.

isometry:
Redundant change :-)

Issue #4712, opened 2024-04-09
Prepare for Flux v2.3.0


stefanprodan:
Flux 2.3.0 promotes the Helm APIs to GA and comes with new features, bug fixes and performance improvements. 

:date: **Target dates:** 13-14 May

🚀  **Roadmap milestone:** https://fluxcd.io/roadmap/#v23-q2-2024

### API promotion

- [x] **HelmRelease** `helm.toolkit.fluxcd.io/v2` (v2beta2 deprecated)
- [x] **HelmRepository** `source.toolkit.fluxcd.io/v1` (v1beta2 deprecated)
- [x] **HelmChart** `source.toolkit.fluxcd.io/v1` (v1beta2 deprecated)
- [x] **ImageUpdateAutomation** `image.toolkit.fluxcd.io/v1beta2` (v1beta1 deprecated)

### Dependencies updates

- [x] Kubernetes 1.30
- [x] Go 1.22 
- [x] Kustomize 5.4
- [x] Helm 3.14
- [x] Cosign 2.2

### Release checklist

- [x] source-controller v1.3.0
- [x] kustomize-controller v1.3.0
- [x] notification-controller v1.3.0
- [x] helm-controller: v1.0.1
- [x] image-reflector-controller v0.32.0
- [x] image-automation-controller v0.38.0
- [x] flux-benchmark v2.3.0
- [x] flux2 v2.3.0
- [ ] terraform-provider-flux v1.3.0

###  Documentation

- [x] Update docs to promoted APIs
- [x] Publish Flux release change log to GitHub & Slack
- [x] [Publish v2.3 website](https://fluxcd.io/flux/releases/procedure/#distribution-minor-release-website)

### Conformance testing

| Distribution | Versions |
|:--------|:--------|
| Kubernetes KinD | 1.28.9, 1.29.4, 1.30.0 |
| OpenShift | 4.15.0-okd |
| K3s | 1.28.7,  1.29.2 | 
| Google Cloud | 1.28.7-gke.1026000 |

### TODO

- [x] https://github.com/fluxcd/source-controller/pull/1428
- [x] https://github.com/fluxcd/source-controller/issues/1391
- [x] https://github.com/fluxcd/source-controller/pull/1075
- [x] https://github.com/fluxcd/source-controller/pull/1412
- [x] https://github.com/fluxcd/kustomize-controller/pull/1129
- [x] https://github.com/fluxcd/kustomize-controller/pull/1130
[...]

PR #4759, opened 2024-05-06
Update Helm Source APIs to v1 (GA) 

stefanprodan:
Part of: #4712


cmd/flux/delete_source_helm.go
```diff
@@ -25,7 +25,7 @@ import (
 var deleteSourceHelmCmd = &cobra.Command{
 	Use:   "helm [name]",
 	Short: "Delete a HelmRepository source",
-	Long:  withPreviewNote("The delete source helm command deletes the given HelmRepository from the cluster."),
+	Long:  "The delete source helm command deletes the given HelmRepository from the cluster.",
```
souleb, line 28: the import need updates too.

PR #4766, opened 2024-05-07
Add support for HelmRelease v2 in `flux reconcile` and `flux create`

### LINKED ISSUE #4761: Add support for HelmReleases v2 with `.spec.chartRef`
In HelmRelease v2 API we've added support for `.spec.chartRef` that allows referencing chart artifacts from OCIRepository and HelmChart objects.

The following CLI commands should be extended to support the new field:

- `flux create hr --chart-ref` e.g. `flux create hr --chart-ref=OCIRepository/default/podinfo`
- `flux reconcile hr --with-source` should trigger an OCIRepository or HelmChart reconciliation if `.spec.chartRef` is specified

souleb:
fixes #4761

### DIFF
--- a/cmd/flux/reconcile_helmrelease.go
+++ b/cmd/flux/reconcile_helmrelease.go
@@ -24,6 +24,7 @@ import (
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	sourcev1b2 "github.com/fluxcd/source-controller/api/v1beta2"
 )
 
 var reconcileHrCmd = &cobra.Command{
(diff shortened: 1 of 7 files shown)


cmd/flux/reconcile_helmrelease.go
```diff
@@ -74,13 +74,25 @@ func (obj helmReleaseAdapter) getSource() (reconcileSource, types.NamespacedName
...
+		ns string
+	)
+	switch {
+	case obj.Spec.Chart != &helmv2.HelmChart{}:
+		ns = obj.Spec.Chart.Spec.SourceRef.Namespace
+		name = fmt.Sprintf("%s-%s", obj.Namespace, obj.Name)
+	case obj.Spec.ChartRef != nil:
+		ns = obj.Spec.ChartRef.Namespace
+		name = obj.Spec.ChartRef.Name
+	}
```
stefanprodan, line 88: What if the kind of chartRef is `OCIRepository`?

stefanprodan (approved):
LGTM

Thanks @souleb 🥇

PR #4780, opened 2024-05-10
Update helm-controller to v1.0.1

stefanprodan:
Update helm-controller to v1.0.1 and adapt code.

Part of: https://github.com/fluxcd/flux2/issues/4712


souleb (approved):
LGTM!

PR #4781, opened 2024-05-10
Set Kubernetes 1.28 as min required version

stefanprodan:
Changes:
- Update the Kubernetes version check and drop EOL versions
- Update kubectl to 1.30.0 in the flux-cli Dockerfile

Part of: #4712

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3.19 as builder
 RUN apk add --no-cache ca-certificates curl
 
 ARG ARCH=linux/amd64
-ARG KUBECTL_VER=1.28.6
+ARG KUBECTL_VER=1.30.0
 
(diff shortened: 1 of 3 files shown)


souleb (approved):
LGTM!

stefanprodan:
We only support the last 3 minor versions of Kubernetes. Also 1.27 active support ended 2 weeks ago, see https://endoflife.date/kubernetes

cwrau:
> We only support the last 3 minor versions of Kubernetes.

I understand, but I would've expected a major release for such a change, I guess?

> Also 1.27 active support ended 2 weeks ago, see [endoflife.date/kubernetes](https://endoflife.date/kubernetes)

Huh, interesting, I just looked at https://kubernetes.io/releases/#release-v1-27 and it shows it to be not EOL for another ~1.5 months 😅

PR #4806, opened 2024-05-21
Rfc for git passwordless auth

dipti-pai:
Based on top of this PR - https://github.com/fluxcd/flux2/pull/4114 
Created separate PR as I don't have the required permissions to the main repo.


rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+> AWS CodeCommit is not supported as it does not support authentication via IAM
+Roles without the use of https://github.com/aws/git-remote-codecommit.
+
+By default, it will be blank, which indicates that the user wants to
+authenticate via HTTP basic/bearer auth or SSH.
+
+### Azure
+
+Git repositories hosted on Azure Devops can be accessed by Flux using OIDC if
```
souleb, line 69: is it possible to add that when you deploy an AKS cluster, a system-assigned managed identity is automatically created, and it's managed by the Azure platform.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,188 @@
...
+---
+kind: Secret
+metadata:
+  name: gcp-sa
+stringData:
+  githubAppID: <app-id>
+  githubInstallationID: <installation-id>
+  githubPrivateKey: |
+    <PEM-private-key>
+  githubApiURl: <github-enterprise-api-url> #optional, required only for GitHub Enterprise users
```
souleb, line 161: I am not sure that we actually need that. This is not needed to get a `token` for the gihubApp. It is need for an enterprise client in `google-go-github` lib.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,188 @@
...
+If `.spec.provider` is set to `azure`, Flux controllers will reach out to
+[Azure IMDS (Instance Metadata Service)](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-go)
+to get an access token. This [access token will be then used as a bearer token](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?view=azure-devops#q-can-i-use-a-service-principal-to-do-git-operations-like-clone-a-repo)
+to perform HTTP bearer authentication.
+
+### GCP
+
+If `.spec.provider` is set to `gcp`, Flux controllers will get the Service
+Account credentials from the specified Secret and use
+[`google.CredentialsFromJSON`](https://pkg.go.dev/golang.org/x/oauth2/google#CredentialsFromJSON)
```
souleb, line 177: We need to verify this. The current code state:

```go
// SERVICE_ACCOUNT_TOKEN_URL is the default GCP metadata server endpoint to
// fetch the access token for a GCP service account.
const SERVICE_ACCOUNT_TOKEN_URL = "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
```

souleb:
@nagyv can this work with gitlab?

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,188 @@
...
+mechanism.
+
+### Non-Goals
+* Replace the existing basic and bearer authentication API.
+
+## Proposal
+
+A new string field `.spec.provider` shall be added to the GitRepository API.
+The field will be an enum with the following variants:
+* `azure`
```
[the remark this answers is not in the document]
dipti-pai: Added details for using `generic` as the default value mimicking oci spec.

rfcs/0006-git-repo-passwordless-auth/README.md
dipti-pai: Resolving as per this example - https://github.com/bradleyfalzon/ghinstallation?tab=readme-ov-file#github-enterprise-example

nagyv:
@souleb Kind of. Reading through the proposal, it's not really passwordless as you extract the password every time from the OIDC response, and the password is tied either to a service account or an app (in case of GitHub).

At GitLab, we support [service accounts](https://docs.gitlab.com/ee/user/profile/service_accounts.html) that are fully independent of every user account, can not log in through the UI, but are very much user-like. For example, they can own deploy keys and can be made members of projects.

Flux in its current state can already support service accounts. I'll try to extend the proposal to describe the "with GitLab" solution.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,214 @@
...
+and the `client_email` as the username to perform HTTP basic authentication.
+
+### GitHub
+
+If `.spec.provider` is set to `github`, Flux controllers will get the app
+details from the specified Secret and use it to [generate an app installation
+token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app).
+This token is then used as the password and [`x-access-token` as the username](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/choosing-permissions-for-a-github-app#choosing-permissions-for-git-access)
+to perform HTTP basic authentication.
+
```
nagyv, line 214:
```suggestion

### GitLab

This RFC does not have a proposal for GitLab as GitLab offers [Service accounts](https://docs.gitlab.com/ee/user/profile/service_accounts.html) that already work with Flux. Service accounts are fully user-independent accounts that do not allow accessing the GitLab UI, but behave as user accounts in every other aspect.

The recommendation is to use Flux with a service account, as service accounts can be members of projects and groups and can own deploy keys. If Flux is used with deploy tokens, it's worth noting that deploy tokens are not tied to users.

GitLab Service accounts fully address the identity drawback mentioned in the motivation, and partially address the scalability problem. With respect to scalability, the service account needs to be added as a member of the respective Flux projects or their parent groups, and the service account needs to own and rotate its deploy keys. Disabling the service account invalidates its deploy keys as well. Removing a service account from a project, does not invalidate its deploy keys. Deploy keys owned by a service account can be enabled/disabled separately by project.

```

souleb: Thanks for this addition. From what I read, users have to create a `service account` and `access token` with the Gitlab API. They also have to handle the rotation of the tokens when an `expiration` is set.  So it's quite different than the supposition here where the cloud provider or an app handle the token management.

nagyv: @souleb  I just figured out the Applications at GitLab seem to work very similarly to Applications at GitHub. The user would need to create an Application either through the UI or [the API](https://docs.gitlab.com/ee/api/applications.html), followingly Flux could use [the OAuth access token for API requests](https://docs.gitlab.com/ee/api/oauth2.html#access-gitlab-api-with-access-token). This still needs to be checked if it would work to retrieve a repo or to access the container registry.

dipti-pai: @nagyv can the gitlab application be used to generate app installation access token similar to gh - https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app ?

souleb: from what I understood, when creating an application there is an attribute for `scopes` and we would have a  response like:

```json
{
    "id":1,
    "application_id": "5832fc6e14300a0d962240a8144466eef4ee93ef0d218477e55f11cf12fc3737",
    "application_name": "MyApplication",
    "secret": "ee1dd64b6adc89cf7e2c23099301ccc2c61b441064e9324d963c46902a85ec34",
    "callback_url": "http://redirect.uri",
    "confidential": true
}
```

The scope that is needed in our case is `scopes=read_repository`. If this scope is set, we should be able to use the `application` in order to retrieve an [access token](https://docs.gitlab.com/ee/api/oauth2.html#authorization-code-flow), which can then be used as defined [here](https://docs.gitlab.com/ee/api/oauth2.html#access-git-over-https-with-access-token) to pull the git repository.

I believe the currently defined api is correct here. As for `github` we would need a secret like:

```yaml
kind: Secret
metadata:
  name: gcp-sa
stringData:
  gitlabAppID: <application_id>
  client_secret: <client_secret>
  redirectURl: <redirect_uri>
```

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,244 @@
...
+        labels:
+          azure.workload.identity/use: "true"
+      spec:
+        template:
+          metadata:
+            labels:
+              azure.workload.identity/use: "true" 
+```
+
+Example of using an Azure Devops repository with azure provider:
```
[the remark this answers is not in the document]
dipti-pai: Changed as suggested.

rfcs/0006-git-repo-passwordless-auth/README.md
dipti-pai: Thank you @nagyv and @souleb. Updated the RFC to include gitlab details as suggested.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,244 @@
...
+spec:
+  interval: 1m
+  url: https://dev.azure.com/<org>/<project>/_git/<repository>
+  ref:
+    branch: master
+  # notice the lack of secretRef
+  provider: azure
+```
+
+### GCP
```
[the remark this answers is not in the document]
dipti-pai: I saw this earlier as well. Looks like it only impacts new users and existing customers can continue to use GCSR.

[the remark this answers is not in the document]
souleb: I think they will replace it by https://cloud.google.com/secure-source-manager/docs/overview.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+Git repositories hosted on Google Cloud Source Repositories can be accessed by
+Flux via a [GCP Service Account](https://cloud.google.com/iam/docs/service-account-overview).
+
+Workload Identity Federation for GKE is [unsupported](https://cloud.google.com/iam/docs/federated-identity-supported-services)
+for Cloud Source Repositories. The user must instead create the GCP Service Account and 
+link it to the Flux source controller service account in order to enable workload identity.
```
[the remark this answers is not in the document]
souleb: I'm relying on the [following](https://cloud.google.com/source-repositories/docs/configure-access-control). It specifically list `roles` and states `members`, not `users`.

The following [doc](https://cloud.google.com/source-repositories/docs/viewing-users-and-permissions) also list  `SA` as a valid `principal`.

I'll write a small poc to validate this though.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,244 @@
...
+    <PEM-private-key>
+  githubApiURl: <github-enterprise-api-url> #optional, required only for GitHub Enterprise users
+```
+
+## Design Details
+
+### Azure
+
+If `.spec.provider` is set to `azure`, Flux controllers will use [DefaultAzureCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential) to build the workload identity credential. This credential type uses the environment variables injected by the Azure Workload Identity mutating webhook. The [access token from the credential will be then used as a bearer token](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?view=azure-devops#q-can-i-use-a-service-principal-to-do-git-operations-like-clone-a-repo)
+to perform HTTP bearer authentication.
```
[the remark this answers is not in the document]
dipti-pai: Thank you for the review :) 

The current thinking is to have the code in pkg with the goal of reusing existing oci auth code when possible. Source controller would have spec changes to pass the provider to the auth code in pkg. 

Good point about the image-automation-controller. Since it uses source api's git repository, we would need changes similar to source controller to pass the provider to the auth code for git in pkg.

rfcs/0006-git-repo-passwordless-auth/README.md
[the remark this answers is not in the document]
souleb: My POC is conclusive. This works with `cloud-source-repositories`.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,351 @@
...
+  githubAppID: <app-id>
+  githubInstallationID: <installation-id>
+  githubPrivateKey: |
+    <PEM-private-key>
+  githubApiURl: <github-enterprise-api-url> #optional, required only for GitHub Enterprise users
+```
+
+### Gitlab
+
+Git repositories hosted on Gitlab can be accessed via OAuth2 Gitlab Applications created from the [UI](https://docs.gitlab.com/ee/integration/oauth_provider.html) or using [API](https://docs.gitlab.com/ee/api/applications.html). The Gitlab Oauth2 application must be created with the required scope to access gitlab repositories. The application's application_id, secret and redirect_uri are used to request an [access token](https://docs.gitlab.com/ee/api/oauth2.html#authorization-code-flow). These parameters are configured in the secret referred to by `.spec.secretRef`.
```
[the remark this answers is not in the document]
nagyv: Wrapped

```suggestion
Git repositories hosted on Gitlab can be accessed via OAuth2 Gitlab Applications
created from the
[UI](https://docs.gitlab.com/ee/integration/oauth_provider.html) or using
[API](https://docs.gitlab.com/ee/api/applications.html). The Gitlab Oauth2
application must be created with the required scope to access gitlab
repositories. The application's `application_id`, `secret` and `redirect_uri` are used
to request an [access
token](https://docs.gitlab.com/ee/api/oauth2.html#authorization-code-flow).
These parameters are configured in the secret referred to by `.spec.secretRef`.
```

nagyv (commented):
Line wrapped and formatted the GitLab section. LGTM.

rfcs/0006-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+### Image Automation controller
+
+Image automation controller updates YAML files when new images are available and commits changes to a given Git repository. The `ImageUpdateAutomation` API defines an automation process that updates the Git repository referenced in it's `.spec.sourceRef`. If the new optional string field `.spec.provider` is specified in the `GitRepository` API, the respective provider is used to configure the authentication to check out the source. 
+
+When `azure` is specified as the provider, image automation controller service account and deployment must be patched to use workload identity (similar to source controller workflow described above) and the managed identity should have sufficient permissions to the referenced Azure Devops Repository. A federated identity must be created between the image automation controller service account and the managed identity.
+
+When `gcp` is specified as the provider, the image automation controller service account must be patched to specify the GCP service account (similar to source controller workflow described above). The GCP service account must have sufficient permissions to the Google Cloud Source repository.
```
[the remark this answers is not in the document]
dipti-pai: Thanks @darkowlzz . Updated document as suggested.

stefanprodan (approved):
LGTM

Thanks @dipti-pai 🏅

rfcs/0000-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,403 @@
+# RFC-0000 Passwordless authentication for Git repositories
```
stefanprodan, line 1:
```suggestion
# RFC-0007 Passwordless authentication for Git repositories
```

Please also rename the directory to `0007-git-repo-passwordless-auth`

rfcs/0000-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+**Status:** provisional
+
+**Creation date:** 2023-31-07
+
```
stefanprodan, line 6: Please add `**Last update:** 2024-06-12`

rfcs/0000-git-repo-passwordless-auth/README.md
```diff
@@ -0,0 +1,403 @@
+# RFC-0000 Passwordless authentication for Git repositories
+
+**Status:** provisional
```
stefanprodan, line 3:
```suggestion
**Status:** implementable
```

alvaroaleman:
For my understanding, how is the work to actually implement this tracked? Are there individual sub work items for the various provides one could follow and contribute to?

dipti-pai:
> For my understanding, how is the work to actually implement this tracked? Are there individual sub work items for the various provides one could follow and contribute to?

#4846 has the workitems for tracking the implementation of this RFC

PR #4842, opened 2024-06-13
Add `part-of` label to controllers base

stefanprodan:
Add `app.kubernetes.io/part-of: flux` label to controllers base to allow targeting all deployments by label in the install kustomization.


souleb (approved):
LGTM!

Issue #4844, opened 2024-06-13
Migrate CI to GitHub ARM64 Linux runners


stefanprodan:
GitHub now offers [ARM64 Linux](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/ ) runners so we can decommission the self-hosted Equinix ones.

Migrate workflows for:

- [x] flux2
- [x] flux-benchmark
- [x] source-controller
- [x] image-automation-controller

Decommission the Equinix machines:

- [x] equinix-arm-dc-1
- [x] equinix-arm-da-1

PR #4845, opened 2024-06-13
Run ARM64 e2e tests on GitHub runners

stefanprodan:
Part of: #4844

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -13,9 +13,8 @@ env:
 
 jobs:
   conform-kubernetes:
-    # Hosted on Equinix
-    # Docs: https://github.com/fluxcd/flux2/tree/main/.github/runners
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance.yaml
```diff
@@ -13,9 +13,8 @@ env:
 
 jobs:
   conform-kubernetes:
-    # Hosted on Equinix
-    # Docs: https://github.com/fluxcd/flux2/tree/main/.github/runners
-    runs-on: [self-hosted, Linux, ARM64, equinix]
+    runs-on:
+      group: "ARM64"
```
souleb, line 17: what is the runner setup? How many runners do we have?

souleb (approved):
LGTM!

.github/workflows/conformance.yaml
stefanprodan: This is custom group I created using GH EE, I used the default 50 max runners.

PR #4892, opened 2024-07-17
Run conformance tests for Kubernetes v1.31

stefanprodan:
Part of: #4928

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         # Keep this list up-to-date with https://endoflife.date/kubernetes
         # Build images with https://github.com/fluxcd/flux-benchmark/actions/workflows/build-kind.yaml
-        KUBERNETES_VERSION: [ 1.28.11, 1.29.6, 1.30.2 ]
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         # Keep this list up-to-date with https://endoflife.date/kubernetes
         # Build images with https://github.com/fluxcd/flux-benchmark/actions/workflows/build-kind.yaml
-        KUBERNETES_VERSION: [ 1.28.11, 1.29.6, 1.30.2 ]
+        KUBERNETES_VERSION: [ 1.28.11, 1.29.6, 1.30.2, 1.31.0 ]
```
makkes, line 22: shouldn't this be
```suggestion
        KUBERNETES_VERSION: [ 1.29.6, 1.30.2, 1.31.0 ]
```
?

makkes (approved):

.github/workflows/conformance.yaml
stefanprodan: Dropping support for a Kubernetes version should have it's own PR. But 1.28 support ends on 28 Aug 2024, we can't remove it now.

Issue #3817, opened 2023-04-19
The "owner" argument for flux bootstrap gitlab (probably others too) does not make sense if URL includes groups/subgroups


aleksandaratanasov:
The research institute I am working for has its own GitLab instance. My team has its own group and inside we have subgroups and what not. The repository I would like to use has the following URL (placeholders used due to privacy):

    https://gitlab.example.com/institute/department/research-group/cloud-computing/administration/k8s-playground.git

As you can see the structure is nested:

 * `institute` - the name of the institute
 * `department` - the name of the department with focus on specific research area(s)
 * `research-group` - the name of the team with focus on specific research area(s)
 * `cloud-computing` - a sub-group of the my team's group I have created (I am the owner) to put everything related to cloud in it
 * `administration` - a sub-group of the `cloud-computing` group for admin tasks (e.g. GitOps for managing the infrastructure)
 * `k8s-playground` - the actual project that will be used by `flux` to sync the Kubernetes (in my case an Ubuntu 22.04 Server with `microk8s`; for setting up `flux` to work with `microk8s` see [here](https://boxofcables.dev/using-flux2-with-microk8s/) )

I am quite new to GitOps and Kubernetes in general but I do believe that `flux` is rather confusing when it comes to anything that is not in the form of

    https://gitlab.example.com/USER/REPOSITORY.git

The URL I have used at the beginning needs to be split into chunks that, given the name of the argument `--owner`, is anything but helpful. It took me half an hour trying to figure out what's going on:

    flux bootstrap gitlab --hostname=https://gitlab.example.com --repository=k8s-playground --branch=main --path=./clusters/playground/ --token-auth --owner=institute/department/research-group/cloud-computing/administration

Notice the absence of trailing/leading `/`, which is an extra thing the user needs to consider not to mention the absence of `.git` suffix. I do realize that internally a lot of concatenation is going on and this is how I managed to figure out what to put where.

However in the end `--owner=institute/department/research-group/cloud-computing/administration` is not really the owner but rather a nested structure.

I would suggest to rename the owner to something like `url-path` or similar. Best would be remove that completely and just go with the full URL for the `--repository` argument since most people are familiar with that anyway. I would further suggest to remove the requirement to discard the `.git` suffix and remove `--hostname`:

    flux bootstrap gitlab --repository=https://gitlab.example.com/institute/department/research-group/cloud-computing/administration/k8s-playground.git --branch=main --path=./clusters/playground/ --token-auth

[...]

PR #4899, opened 2024-07-20
Gitlab bootstrap project path

nagyv:
Introduce `projectPath` argument for GitLab bootstrap

Deprecates the `owner`, `repository` and `private` arguments.

Not ready for merging. Needs testing on self-managed instance.

Closes: https://github.com/fluxcd/flux2/issues/3817


stefanprodan:
@nagyv I would like to include this in Flux 2.4, do you think you can test with self-managed in the following 2 weeks?

nagyv:
I'm just returning from the offline world. I'll try to test it this week.

-------- Eredeti üzenet --------
2024. 08. 22. 15:04-kor, Stefan Prodan ezt írta:

PR #4916, opened 2024-08-07
flux diff artifact: Print the differences in human readable form.

octo:
I was hoping to use `flux diff artifact` as part of a CI pipeline to show the difference between the merge request and the currently deployed artifact. The existing implementation doesn't work for us, since it only compares the checksums.

This commit changes the output produced by `flux diff artifact` to a list of changes in human readable form. The code is using the `dyff` package to produce a semantic diff of the YAML files. That means, for example, that changes in the order of map fields are ignored, while changes in the order of lists are not.

Example output:

```
$ ./bin/flux diff artifact "oci://${IMAGE}" --path=example-service/

spec.replicas  (apps/v1/Deployment/example-service-t205j6/backend-production)
  ± value change
    - 1
    + 7

✗ "oci://registry.gitlab.com/${REDACTED}/example-service-t205j6/deploy:production" and "example-service/" differ
```

The new `--brief` / `-q` flag enables users to revert to the previous behavior of only printing a has changed/has not changed line.


stefanprodan:
@octo you fork is behind, please rebase with upstream main and force push, also please fix the diff test that's currently failing.

octo:
@stefanprodan 

✅ rebased
✅ fixed unit test – I was quite confused by the fact that `go test ./...` passes but `make test` fails. I suspect that the test harness could be improved, but I haven't tracked down the problem.

cmd/flux/diff_artifact.go
```diff
@@ -103,10 +113,112 @@ func diffArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
 
-	logger.Successf("no changes detected")
-	return nil
+	if diff == "" {
+		logger.Successf("no changes detected")
+		return nil
+	}
+
+	if !diffArtifactArgs.brief {
+		fmt.Print(diff)
```
stefanprodan, line 127: This should be `cmd.Print` like in `flux diff kustomization`

octo: ✅ done

stefanprodan:
xref:
- https://github.com/fluxcd/flux2/issues/3839
- https://github.com/fluxcd/flux2/pull/4138

stefanprodan:
@octo what happens if Helm charts are inside the artifact? Are the templated YAMLs excluded or the CLI will crash?

stefanprodan:
We can't assume an artifact has Kubernetes YAMLs, Flux artifacts are used for many other things besides YAML, for example Terraform modules that work with [tofu-controller](https://github.com/flux-iac/tofu-controller).

I think we need to put the YAML diff behind a flag e.g. `--content-type=yaml`, by default the diff shouldn't assume YAMLs, the logic in this PR should be opt-in by users via the flag.

octo:
> @octo what happens if Helm charts are inside the artifact? Are the templated YAMLs excluded or the CLI will crash?

I don't know, since I haven't tried. My suspicion is that the YAML parsing will fail and that the command errors out.

In that case, we're probably better off using a non-semantic diff, something akin to using `diff -u` on the command line. I'll see if I can come up with something.

stefanprodan:
I think there is value in showing a semantic diff for YAML content, but we need to make it opt-in and exclude templated files, like those in Helm charts.

octo:
@stefanprodan I added code that uses `diff -ur` to compare the two directories, which is now the default.

The `--semantic-diff yaml` flag can be used to trigger the dyff based semantic diff.

I unified the handling of the two arguments, allowing each to be a directory, `.tgz` file, or OCI url. This allows users to diff in any direction they desire. With this change, it feld more natural to use two positional arguments. I left the `--path` flag in for backwards compatibility though.

Unfortunately, the `--ignore-paths` flag is no longer honored. I'm not sure I'll have the time to fix this before going on vacation. I'm happy to look into it, but I'll be out for three weeks.

cmd/flux/diff_artifact.go
```diff
@@ -17,66 +17,107 @@ limitations under the License.
...
 var diffArtifactCmd = &cobra.Command{
-	Use:   "artifact",
+	Use:   "artifact <from> <to>",
 	Short: "Diff Artifact",
-	Long:  withPreviewNote(`The diff artifact command computes the diff between the remote OCI artifact and a local directory or file`),
+	Long: withPreviewNote(fmt.Sprintf(
+		"The diff artifact command prints the diff between the remote OCI artifact and a local directory or file.\n\n"+
+			"You can overwrite the command used for diffing by setting the %q environment variable.", externalDiffVar)),
 	Example: `# Check if local files differ from remote
-flux diff artifact oci://ghcr.io/stefanprodan/manifests:podinfo:6.2.0 --path=./kustomize`,
```
stefanprodan, line 36: This is a breaking change, please undo it.

octo: The previous `flux diff artifact <path1> --path=<path2>` syntax is still supported for backwards compatibility. This simply makes `flux diff artifact <path1> <path2>` the _documented_ syntax. If you don't like it, I can revert back to the `--path` syntax, but compatibility should not be broken.

cmd/flux/diff_artifact.go
```diff
@@ -98,15 +139,308 @@ func diffArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
+
+// extractTo extracts the .tar.gz / .tgz archive at archivePath into the destDir directory.
+func extractTo(archivePath, destDir string) error {
+	archiveFH, err := os.Open(archivePath)
+	if err != nil {
 		return err
 	}
 
-	logger.Successf("no changes detected")
+	gzipReader, err := gzip.NewReader(archiveFH)
```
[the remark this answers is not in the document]
octo: Great pointer, thanks! ✅ changed

cmd/flux/diff_artifact.go
```diff
@@ -98,15 +139,308 @@ func diffArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
+	var buf bytes.Buffer
+
+	if err := printers.NewDyffPrinter().Print(&buf, report); err != nil {
+		return "", fmt.Errorf("formatting dyff report: %w", err)
+	}
+
+	return buf.String(), nil
+}
+
+// semanticDiffFlag implements pflag.Value for chosing a semantic diffing algorithm.
```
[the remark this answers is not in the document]
octo: ✅ fixed

cmd/flux/diff_artifact.go
```diff
@@ -98,15 +139,308 @@ func diffArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
+	}
+}
+
+func (c externalDiffCommand) Diff(ctx context.Context, fromDir, toDir string) (string, error) {
+	var args []string
+
+	args = append(args, c.flags...)
+	args = append(args, fromDir, toDir)
+
+	cmd := exec.CommandContext(ctx, c.name, args...)
```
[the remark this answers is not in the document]
octo: ✅ done – the `--differ` flag now takes one of three arguments:

1. `dyff`: Use the `dyff` package to generate a semantic YAML diff.
2. `external`: execute the command in the `FLUX_EXTERNAL_DIFF` environment variable. Using `--differ external` without setting `FLUX_EXTERNAL_DIFF` is now an error, there is no default command being executed. That means flux makes no assumptions about the system is runs on.
3. `unified`: uses [github.com/hexops/gotextdiff](https://pkg.go.dev/github.com/hexops/gotextdiff) to generate a unified diff without invoking external commands.

   I looked into unified diffs and Golang quite a bit more than I wanted to — the situation is surprisingly bad. I opted to use [github.com/hexops/gotextdiff](https://pkg.go.dev/github.com/hexops/gotextdiff), which is a clone of one of the internal-only packages by the Go authors. The repo is marked as archived as of earlier this year, but overall this seems like the least bad option.

> In the past, we have taken other such code from upstream Go and adapted it for our needs

As a maintainer myself, I'd work hard to avoid that. Copying and adapting code rather than referencing it means that all the maintenance overhead is on the Flux maintainers, which can snowball quickly.

octo:
@stefanprodan @darkowlzz Thank you so much for your reviews! I have addressed all review comments and rebased my branch on top of `main`.

go.mod
```diff
@@ -6,6 +6,7 @@ go 1.22.4
 replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
 
 require (
+	bitbucket.org/creachadair/stringset v0.0.14
```
stefanprodan, line 9: This project is not Apache 2.0 licensed, so we can't use it.

go.mod
```diff
@@ -157,6 +159,7 @@ require (
 	github.com/hashicorp/go-version v1.7.0 // indirect
 	github.com/hashicorp/golang-lru/arc/v2 v2.0.5 // indirect
 	github.com/hashicorp/golang-lru/v2 v2.0.5 // indirect
+	github.com/hexops/gotextdiff v1.0.3
```
stefanprodan, line 162: This package is no longer maintained, so we can't use it.

PR #4922, opened 2024-08-09
Stop spinner on cancel of flux diff kustomization

bkreitch:
This PR stops spinner on cancel of `flux diff kustomization` command - otherwise cursor (hidden by spinner package) remains invisible.

`stopSpinner()` is called before `CleanDirectory()` since later happens to return an error if _kustomize.yaml_ doesn't exist and then `Cancel()` exits still leaving cursor invisible.


stefanprodan (approved):
LGTM

Thanks @bkreitch for the fix 🏅

Issue #4928, opened 2024-08-14
Update Flux components to Kubernetes 1.31


stefanprodan:
Update the Kubernetes Go packages and adapt to the breaking changes in controller-runtime v0.19 in all Flux projects and run conformance for Kubernetes 1.31.0.

 Projects:
- [x] cli-utils
- [x] pkg
- [x] source-controller 
- [x] kustomize-controller 
- [x] notification-controller 
- [x] helm-controller
- [x] image-reflector-controller 
- [x] image-automation-controller
- [x] flux2
- [x] terraform-provider-flux

PR #4934, opened 2024-08-15
Update dependencies to Kubernetes v1.31.0

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4928


stefanprodan:
The GCP failure can be ignored, the issue is with the upstream Terraform provider.

souleb (approved):

Issue #4927, opened 2024-08-14
Build Flux components with Go 1.23


stefanprodan:
Update Go to 1.23 in Dockerfiles and GHA workflows.

Projects:
- [x] cli-utils
- [x] pkg
- [x] go-git-providers
- [x] source-watcher
- [x] source-controller 
- [x] kustomize-controller 
- [x] notification-controller 
- [x] helm-controller
- [x] image-reflector-controller 
- [x] image-automation-controller
- [x] flux2
- [x] terraform-provider-flux

PR #4936, opened 2024-08-15
Build with Go 1.23

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4927

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
-FROM alpine:3.19 as builder
+FROM alpine:3.20 as builder
 
 RUN apk add --no-cache ca-certificates curl
 
 ARG ARCH=linux/amd64
-ARG KUBECTL_VER=1.30.0
+ARG KUBECTL_VER=1.31.0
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,15 +1,15 @@
-FROM alpine:3.19 as builder
+FROM alpine:3.20 as builder
 
 RUN apk add --no-cache ca-certificates curl
 
 ARG ARCH=linux/amd64
-ARG KUBECTL_VER=1.30.0
+ARG KUBECTL_VER=1.33.0
```
makkes, line 6:
```suggestion
ARG KUBECTL_VER=1.31.0
```
?

makkes (approved):

Dockerfile
stefanprodan: Good catch! Thanks

makkes:
I suppose we should also update `.github/runners/prereq.sh` to use kubectl 1.31.

stefanprodan:
The runners have been decommissioned, we no longer use Equinix. We are now using GitHub EE ARM runners.

Issue #4961, opened 2024-08-30
flux commands panic when $HOME $XDG_CONFIG_HOME unset


hillsprig:
### Describe the bug

I'm running a couple of flux commands in daemon scripts to check and do various operations towards my edge cluster.

In Flux 2.2.3 this worked fine. In Flux 2.3.0 $HOME or $XDG_CONFIG_HOME seems to be required and we get a panic otherwise.

### Steps to reproduce

% env -u HOME -u XDG_CONFIG_HOME ./flux_2.3.0_linux_amd64 check 

panic: neither $XDG_CONFIG_HOME nor $HOME are defined

goroutine 1 [running]:
github.com/notaryproject/notation-go/dir.loadUserPath()
	/home/runner/go/pkg/mod/github.com/notaryproject/notation-go@v1.1.0/dir/path.go:85 +0xc5
github.com/notaryproject/notation-go/dir.init.0()
	/home/runner/go/pkg/mod/github.com/notaryproject/notation-go@v1.1.0/dir/path.go:77 +0xf

### Expected behavior

% env -u HOME -u XDG_CONFIG_HOME ./flux_2.2.3_linux_amd64 check

► checking prerequisites
✗ flux 2.2.3 <2.3.0 (new CLI version is available, please upgrade)
✔ Kubernetes 1.30.4+k3s1 >=1.26.0-0
► checking version in cluster
✔ distribution: flux-v0.0.0-dev.0
✔ bootstrapped: false
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v1.0.1
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v1.3.0
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v1.3.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v1.3.0
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta3
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1
✔ helmcharts.source.toolkit.fluxcd.io/v1
✔ helmreleases.helm.toolkit.fluxcd.io/v2
✔ helmrepositories.source.toolkit.fluxcd.io/v1
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1
[...]

PR #4962, opened 2024-08-30
Remove notation validation to avoid panics

stefanprodan:
Fix: #4961


stefanprodan:
Thanks to @JasonTheDeveloper we'll have a fix upstream. We can close this PR once the notation-go is released.

PR #4967, opened 2024-09-03
tests: use tempdir to avoid manual gc

### LINKED ISSUE #3594: Use `t.TempDir()` where applicable instead of manually dealing with GC
> There is more here that could be done, as using `./testdata/` for something temporary seems wrong.
>
> I think we should instead use `t.TempDir()` as a target for the file that is been written. Which would prevent having to think about garbage collecting it later on.
>
> This is however beyond the time I have available now, as I was merely trying to fix things I ran into while working on #3587.
>
> _Originally posted by @hiddeco in https://github.com/fluxcd/flux2/pull/3592#discussion_r1107067652_         

This is one example, but there may be more.

mxtw:
I was not able to find any other places where this was applicable, other places that use temporary files are not within the context of a testing.T, so I was not sure how to deal with this.
I have also removed a redundant call to testEnv.CreateObjectFile, let me know if you would rather have this in a separate commit!

Fixes #3594

### DIFF
--- a/cmd/flux/build_kustomization_test.go
+++ b/cmd/flux/build_kustomization_test.go
@@ -22,6 +22,7 @@ package main
 import (
 	"bytes"
 	"os"
+	"path/filepath"
 	"testing"
 	"text/template"
 )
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @mxtw

PR #5060, opened 2024-11-04
Use replicated-actions in conformance tests 

jdewinne:
Official repo is `replicated-actions`

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -107,7 +107,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITPROVIDER_BOT_TOKEN }}
       - name: Create cluster
         id: create-cluster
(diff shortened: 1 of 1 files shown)


AmberAlston:
@stefanprodan FYI this PR updates the reference from the older Replicated Compatibility Matrix repo name to the current one to make sure nothing breaks in the future for your OpenShift and k3s testing. 

I noticed that Flux hasn't created any new test clusters since Sept 30, but it looks like that also lines up with your last release tag of Sept 30. As always, when you start up testing again, let us know if you have any issues or feedback.

stefanprodan:
Thanks @jdewinne and @AmberAlston, the conformance tests will run after this gets merged, we cannot run them on PRs from external contributors due to security constraints.

stefanprodan (approved):
LGTM

PR #5071, opened 2024-11-12
fix: skip remote Kustomizations on recursive diff

milas:
When recursively diffing Kustomization objects on a cluster, each Kustomization is built and then a server-side dry-run apply is performed. This isn't practical to perform if the Kustomization references a remote cluster via `kubeConfig.secretRef`.

Currently, because it's not skipped, it's treated as though it's being applied to the context cluster, not the remote cluster, which results in an incorrect diff.

Instead, write out special message / warnings indicating that it has been skipped:
```
► Kustomization/my-ns/my-kst skipped: diff not supported for remote clusters
```

See also:
 * https://github.com/fluxcd/flux2/discussions/5065

### DIFF
--- a/internal/build/diff.go
+++ b/internal/build/diff.go
@@ -146,14 +146,18 @@ func (b *Builder) diff() (string, bool, error) {
 			}
 
 			if !kustomizationsEqual(k, b.kustomization) {
-				subOutput, subCreatedOrDrifted, err := b.kustomizationDiff(k)
-				if err != nil {
-					diffErrs = append(diffErrs, err)
-				}
-				if subCreatedOrDrifted {
-					createdOrDrifted = true
-					output.WriteString(bunt.Sprint(fmt.Sprintf("📁 %s changed\n", ssautil.FmtUnstructured(obj))))
-					output.WriteString(subOutput)
+				if k.Spec.KubeConfig != nil {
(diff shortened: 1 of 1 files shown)


milas:
FYI I did _not_ make this return an error code. I think there's three reasonable options:
* Leave as-is -> it's a warning only (initial impl of this PR), no change to exit code
* Overload `createdOrDrifted` to be "created or drifted _or skipped_"
* Make it a hard error -> personally, I'd want a CLI flag to then ignore it 🫤

internal/build/diff.go
```diff
@@ -146,14 +146,18 @@ func (b *Builder) diff() (string, bool, error) {
...
-				subOutput, subCreatedOrDrifted, err := b.kustomizationDiff(k)
-				if err != nil {
-					diffErrs = append(diffErrs, err)
-				}
-				if subCreatedOrDrifted {
-					createdOrDrifted = true
-					output.WriteString(bunt.Sprint(fmt.Sprintf("📁 %s changed\n", ssautil.FmtUnstructured(obj))))
-					output.WriteString(subOutput)
+				if k.Spec.KubeConfig != nil {
+					output.WriteString(writeString(fmt.Sprintf("► %s skipped: diff not supported for remote clusters\n", ssautil.FmtUnstructured(obj)), bunt.Orange))
```
stefanprodan, line 150: Let's use the warning emoji here

stefanprodan (approved):
LGTM

Thanks @milas 🏅

PR #5073, opened 2024-11-13
docs: Mention Flux upgrade guide in release notes

mloskot:
See my brief exchange with @stefanprodan in
- https://github.com/fluxcd/flux2/discussions/5072


stefanprodan (approved):
LGTM

Thanks @mloskot

Issue #5101, opened 2024-12-09
Troubleshoot helpers for Helm release install/upgrade failures


stefanprodan:
To help users troubleshoot `HelmRelease` install and upgrade failures, we could add a new CLI command dedicated for debugging. 

### Specification

Introduce a new command called `flux debug helmrelease` that would help solve https://github.com/fluxcd/helm-controller/issues/497

### Debug values composition

The `flux debug helmrelease <name> --show-values` command could be used to print the final values by merging the `HelmRelease` inline values from `.spec.values` with the values coming from Kubernetes ConfigMaps/Secrets specified in `.spec.valuesFrom`. This command could help users debug failed releases by allowing them to export the final values from cluster into a local file. Using the final values, users can run locally `helm template --values in-cluster-values.yaml` and inspect the resulting resources. 

Note that the exported values may contain sensitive information extracted from Kubernetes Secrets, this should be highlighted in the command help docs to make users aware of the implications of running this command on untrusted machines. Give that the Flux CLI uses the local kubeConfig, only users with read access to the secrets will be able to run the debug command.

### Debug current status

The `flux debug helmrelease <name> --show-status` command can be used to print the release status along with the upgrade history, conditions and errors messages. This command could help users debug failed releases by inspecting the status sub-resource which contains rich informations about a HelmRelease compared to the `flux get helmreleaase` command which only displays the Ready condition message. 

## Implementation 

- [x] Extract the values composition logic from helm-controller to a shared Go package under `fluxcd/pkg/chartutil`
- [x] Refactor helm-controller to make use of the shared package from `fluxcd/pkg/chartutil`
- [ ] Implement `flux debug helmreleaase --show-values` using the shared package from `fluxcd/pkg/chartutil`
- [ ] Implement `flux debug helmreleaase --show-status`
- [ ] Document the command usage in the HelmRelease API docs.

PS. Refactoring the chartutil into a shared package is the first step towards `flux build helmrelease`.

PR #5106, opened 2024-12-10
Implement `flux debug helmrelease` command

stefanprodan:
This PR implements the `flux debug hr` command as specified in #5101 

```
The debug helmrelease command can be used to troubleshoot failing Helm release reconciliations.
WARNING: this command will print sensitive information if Kubernetes Secrets are referenced in the HelmRelease .spec.valuesFrom field.

Usage:
  flux debug helmrelease [name] [flags]

Aliases:
  helmrelease, hr

Examples:
  # Print the status of a Helm release
  flux debug hr podinfo --show-status

  # Export the final values of a Helm release composed from referred ConfigMaps and Secrets
  flux debug hr podinfo --show-values > values.yaml
```

Closes: #5101 
Fixes: https://github.com/fluxcd/helm-controller/issues/497


cmd/flux/debug_helmrelease.go
```diff
@@ -0,0 +1,121 @@
...
+	if debugHelmReleaseArgs.showStatus {
+		status, err := yaml.Marshal(hr.Status)
+		if err != nil {
+			return err
+		}
+		rootCmd.Print(string(status))
+	}
+
+	if debugHelmReleaseArgs.showValues {
+		// TODO(stefan): remove the mapping when helm-controller/api v1.2.0 has been released
```
matheuscscp, line 92: Why? Are we going to store the values in the status in v1.2.0?

cmd/flux/debug_helmrelease.go
```diff
@@ -0,0 +1,121 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/yaml"
+
+	"github.com/fluxcd/flux2/v2/internal/utils"
+)
+
+var debugHelmReleaseCmd = &cobra.Command{
+	Use:     "helmrelease [name]",
+	Aliases: []string{"hr"},
+	Short:   "Debug a HelmRelease resource",
```
matheuscscp, line 37: Since the command expects exactly one argument we use this cobra utility to make it return a nice error/usage message in case the argument is not provided instead of a panic:

```suggestion
	Short:   "Debug a HelmRelease resource",
	Args:    cobra.ExactArgs(1),
```

cmd/flux/debug_helmrelease.go
stefanprodan: This is a type conversion from helm api (old) to the meta api (new). See the PR that you've reviewed again https://github.com/fluxcd/helm-controller/pull/1122

cmd/flux/debug_helmrelease.go
stefanprodan: Nice! We need to add this to all commands

cmd/flux/debug_helmrelease.go
matheuscscp: got it, thanks!

cmd/flux/debug_helmrelease.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		values, err := yaml.Marshal(finalValues)
+		if err != nil {
+			return err
+		}
+		rootCmd.Print(string(values))
+	}
+
+	return nil
```
matheuscscp, line 121: LGTM, just two comments:

* It would be nice to print something if neither flags are passed? Maybe default one of them to true, e.g. the status?
* If both flags are passed the two marshaled YAMLs will get merged together. Can we print the document separation between them if that happens? The `---`.

cmd/flux/debug_helmrelease.go
stefanprodan: So before we release Flux 2.5, we can delete the for loop and use directly the Values type from HC API.

cmd/flux/debug_helmrelease.go
stefanprodan: > It would be nice to print something if neither flags are passed? 

An error is returned, see line 67. 

> Can we print the document separation 

Yeap, I'll do that, thanks!

matheuscscp: > An error is returned, see line 67.

Oh I missed that, sorry.

cmd/flux/debug_helmrelease.go
stefanprodan: I've added the ExactArgs to this command, for all others we'll do it in a separate PR.

cmd/flux/debug_helmrelease.go
```diff
@@ -0,0 +1,125 @@
...
+  flux debug hr podinfo --show-status
+
+  # Export the final values of a Helm release composed from referred ConfigMaps and Secrets
+  flux debug hr podinfo --show-values > values.yaml`,
+	RunE: debugHelmReleaseCmdRun,
+	Args: cobra.ExactArgs(1),
+}
+
+type debugHelmReleaseFlags struct {
+	name       string
```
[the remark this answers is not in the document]
stefanprodan: Fix in #5117 along with completion for the resource name.

cmd/flux/debug_helmrelease.go
```diff
@@ -0,0 +1,125 @@
...
+	if err := kubeClient.Get(ctx, hrName, hr); err != nil {
+		return err
+	}
+
+	if debugHelmReleaseArgs.showStatus {
+		status, err := yaml.Marshal(hr.Status)
+		if err != nil {
+			return err
+		}
+		rootCmd.Print(string(status))
```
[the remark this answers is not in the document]
stefanprodan: After implementing `flux debug kustomization` I realise that mixing the outputs is not the best idea. Here by chance both flags output YAML, but this is not the case in the `flux debug kustomization`. I'm for not allowing both flags to be set, similar to how it's done now in #5117

PR #4729, opened 2024-04-17
Add OpenShift to the conformance test suite

stefanprodan:
This PR adds end-to-end tests for bootstrapping Flux on OpenShift clusters (latest two minor versions corresponding to Kubernetes to 1.27 and 1.28).

Big thanks to @marccampbell and [Replicated](https://www.replicated.com/) for sponsoring the Flux project with on-demand OpenShift clusters 🤗 

PS. Our OpenShift docs are wrong and Flux fails to bootstrap. The docs need to be updated with the patches from this PR!

Closes: https://github.com/fluxcd/flux2/issues/4625
Part of: https://github.com/fluxcd/flux2/issues/4712 and #4609

PR #5137, opened 2025-01-09
Add OpenShift 4.16 & 4.17 to conformance testing

scottrigby:
OpenShift 4.16 & 4.17 are also now available to add to the existing conformance testing.
See https://github.com/fluxcd/flux2/pull/4729#issuecomment-2578138218


stefanprodan (approved):
LGTM

Thanks @scottrigby

PR #5142, opened 2025-01-12
Fix create command always using imageRepositoryType

matheuscscp:


stefanprodan (approved):
LGTM

Thanks @matheuscscp

PR #5151, opened 2025-01-17
[RFC] Custom Health Checks using CEL expressions

stefanprodan:
This RFC proposes to extend the Flux `Kustomization` API with custom health checks for custom resources using the Common Expression Language (CEL).

Based on #4528


rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,278 @@
...
+resource based on a condition other than `Ready`. This is the case for resources
+that do intermediate patching like `Certificate` where you should look at the `Issued`
+condition to know if the certificate has been issued or not before looking at the
+`Ready` condition.
+
+In order to provide a generic solution for custom resources, that would not imply
+writing a custom `kstatus` reader for each CRD, we need to provide a way for the user
+to express the conditions that need to be met in order to determine the status.
+And we need to do this in a way that is flexible enough to cover all possible use cases,
+without having to change Flux source code for each new use case.
```
souleb, line 37:
```suggestion
It should be done in a way that is flexible enough to cover all possible use cases,
without having to change Flux source code for each new one.
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,278 @@
...
+Using `.spec.healthCheckExprs`, Flux users can specify that the `Cluster`
+kind is expected to have a `Ready` condition which will force Flux into waiting
+for the ClusterAPI resources status to be populated.
+
+### Alternatives
+
+We need an expression language that is flexible enough to cover all possible use
+cases, without having to change Flux source code for each new use case.
+
+An alternative that have been considered was to use `CUE` instead of `CEL`.
```
souleb, line 169:
```suggestion
An alternative that has been considered was to use `CUE` instead of `CEL`.
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+Note that all the Kubernetes core resources are discarded from the `healthCheckExprs` list.
+
+### Custom Health Check Library
+
+To help users define custom health checks, we will provide on the [fluxcd.io](https://fluxcd.io)
+website a library of custom health checks for popular custom resources.
+
+The Flux community will be able to contribute to this library by submitting pull requests
+to the `fluxcd/website` repository.
```
souleb, line 123: maybe a link here?

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,298 @@
...
+specify the conditions that need to be met in order to determine the status of
+a custom resource. This enables Flux to query any `.status` field,
+besides the standard `Ready` condition, and evaluate it using a CEL expression.
+
+Example for `SealedSecret` which has a `Synced` condition:
+
+```yaml
+  - apiVersion: bitnami.com/v1alpha1
+    kind: SealedSecret
+    inProgress: "metadata.generation != status.observedGeneration"
```
souleb, line 144: note that checking the generation for `inProgress` is done by default by in the POC. But as we perform the evaluation with the whole object, any field can used in the expression.

stefanprodan: Not all CRDs have `status.observedGeneration` for example one of Flux Operator nor ClusterAPI, it's kinda bad practice since observedGeneration should be per condition.

stefanprodan: I've added a note about automatic evaluation of observedGeneration.

rfcs/0000-custom-health-checks/README.md
stefanprodan: Done

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,314 @@
...
+[library](#custom-health-check-library) will be super useful as some of the expressions might be complex.
+
+The `InProgress` expression is optional, when not specified the  controller will determine
+if the resource is in progress if both `Failed` and `Current` evaluate to `false`.
+Moreover, if the `InProgress` expression is not specified and the custom resource has a
+`.status.observedGeneration` field, the controller with compare it with the `.metadata.generation`
+field to determine if the resource is in progress.
+
+The `Failed` expression is optional, when not specified the controller will keep evaluating the
+`Current` expression until it returns `true`, and gave up after the timeout is reached.
```
souleb, line 232:
```suggestion
`Current` expression until it returns `true`, and give up after the timeout is reached.
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,314 @@
...
+Users should be encouraged to provide a `Failed` expression to avoid staling the reconciliation
```
souleb, line 233:
```suggestion
Users will be encouraged to provide a `Failed` expression to avoid staling the reconciliation
```

souleb (approved):
LGTM!

matheuscscp:
Will review next week.

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+
+		for statusKey, expr := range exprs {
+			// Use CEL to evaluate the expression
+			result, err := cel.ProcessExpr(expr, obj)
+			if err != nil {
+				// handle error
+			}
+			switch statusKey {
+			case status.CurrentStatus.String():
+			// If the expression evaluates to true, we return the current status
```
matheuscscp, line 297:
```suggestion
			// If the expression evaluates to true, we return the Current status
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+			case status.FailedStatus.String():
+			// If the expression evaluates to true, we return the failed status
```
matheuscscp, line 299:
```suggestion
			// If the expression evaluates to true, we return the Failed status
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+			case status.InProgressStatus.String():
+			// If the expression evaluates to true, we return the reconciling status
```
matheuscscp, line 301:
```suggestion
			// If the expression evaluates to true, we return the InProgress status
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+			}
+		}
+		
+		// If none of the expressions evaluate to true, we return the reconciling status
```
matheuscscp, line 305:
```suggestion
		// If none of the expressions evaluate to true, we return the InProgress status
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+	Failed string `json:"failed,omitempty"`
+}
+```
+
+If a CEL expression evaluation results in an error, for example looking for a field that does not exist,
+the health check will fail. Users will be encouraged to test their expressions
+in the [CEL Playground](https://playcel.undistro.io/). Here is where the community maintained
+[library](#custom-health-check-library) will be super useful as some of the expressions might be complex.
+
+The `InProgress` expression is optional, when not specified the  controller will determine
```
matheuscscp, line 225:
```suggestion
The `InProgress` expression is optional, when not specified the controller will determine
```

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+  "github.com/fluxcd/cli-utils/pkg/kstatus/status"
+  "github.com/fluxcd/pkg/runtime/cel"
+  "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func genericConditions(kind string, exprs map[string]string) func(u *unstructured.Unstructured) (*status.Result, error) {
+  return func(u *unstructured.Unstructured) (*status.Result, error) {
+		obj := u.UnstructuredContent()
+
+		for statusKey, expr := range exprs {
```
matheuscscp, line 289: This `map[string]string` could be a struct:

```go
type HealthCheckExpressions struct {
	// Current is the CEL expression that determines if the status
	// of the custom resource has reached the desired state.
	// +required
	Current string `json:"current"`
	// InProgress is the CEL expression that determines if the status
	// of the custom resource has not yet reached the desired state.
	// +optional
	InProgress string `json:"inProgress,omitempty"`
	// Failed is the CEL expression that determines if the status
	// of the custom resource has failed to reach the desired state.
	// +optional
	Failed string `json:"failed,omitempty"`
}
```

And the struct `CustomHealthCheck` could embed this struct inlining the JSON fields:

```go
type CustomHealthCheck struct {
	// APIVersion of the custom resource under evaluation.
	// +required
	APIVersion string `json:"apiVersion"`
	// Kind of the custom resource under evaluation.
	// +required
	Kind string `json:"kind"`

	HealthCheckExpressions `json:",inline"`
}
```

This is not only better in terms of typing but also avoids the random iteration order of maps in Go. Why is a predictable order better in this case? Because, despite being a configuration error to write CEL expressions that evaluate both `Current` and `Failed` to `true`, a random iteration order in the map will produce random results, sometimes one that the user will deem expected, sometimes one that does not make sense. A predictable processing order will make it easier to debug this scenario. This is useful because we are not able to parse the expressions and say if they're conflicting or not before evaluating them for an actual input.

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+}
+```
+
+If a CEL expression evaluation results in an error, for example looking for a field that does not exist,
+the health check will fail. Users will be encouraged to test their expressions
+in the [CEL Playground](https://playcel.undistro.io/). Here is where the community maintained
+[library](#custom-health-check-library) will be super useful as some of the expressions might be complex.
+
+The `InProgress` expression is optional, when not specified the  controller will determine
+if the resource is in progress if both `Failed` and `Current` evaluate to `false`.
```
matheuscscp, line 226:
```suggestion
if the resource is in progress if both `Failed` and `Current` do not evaluate to `true`.
```

`Failed` is also optional, so the suggested wording makes more sense.

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,316 @@
...
+without having to change Flux source code for each new CRD.
+
+### Goals
+
+- Provide a generic solution for users to customise the health check evaluation of custom resources.
+- Provide a space for the community to contribute custom health checks for popular custom resources.
+
+### Non-Goals
+
+- We do not plan to support custom health checks for Kubernetes core resources.
```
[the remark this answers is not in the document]
stefanprodan: Yes we'll not create an exclusion list, this is more to say that  users should let kstatus will run its own logic for core resources which are quite complex and can't be replaced with CEL.

rfcs/0000-custom-health-checks/README.md
stefanprodan: Great suggestion! We can iterate over it in the implementation, the guide lines here are more about how the public API will look for users than how we'll actually implement the kstatus extension.

rfcs/0000-custom-health-checks/README.md
```diff
@@ -0,0 +1,339 @@
...
+				return nil, err
+			}
+			if result {
+				// return InProgress status
+			}
+		} else {
+			// compare status.observedGeneration with metadata.generation
+		}
+
+		// return InProgress status
```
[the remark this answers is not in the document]
stefanprodan: > The InProgress expression does not need to exist.

We shouldn't be evaluating `Current` nor `Failed` if `InProgress` returns `true`, so makes a lot of sense to have `InProgress` since we avoid false positives. As for `status.observedGeneration` not all CRs have it, also not CRs have `observedGeneration` on their conditions, so there is no way to avoid false positives of Current/Failed if we don't evaluate `InProgress`.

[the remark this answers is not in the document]
matheuscscp: > We shouldn't be evaluating `Current` nor `Failed` if `InProgress` returns `true`

Then `InProgress` must be the first one evaluated, correct?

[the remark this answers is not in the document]
stefanprodan: The order would be:

- if `status.observedGeneration` exists and if it is diffrent to `metadata.generation` then return `InProgress` and don't bother evaluating anything.
-  if `exprs.InProgress` is defined, evaluate it and if it's `true` then return `InProgress` and don't bother evaluating anything else. If it evaluates to `false` then continue with  `exprs.Failed`.
- if `exprs.Failed` is defined, evaluate it and if it's `true`  then return `Failed` and don't bother evaluating `Current`.  If it evaluates to `false` then continue with  `exprs.Current`.
- if `exprs.Current` evaluates to `true` then return `Current`, if evaluates to `false` then return `InProgress`.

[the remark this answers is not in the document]
matheuscscp: Checking the generation in this context and bailing if it doesn't match, i.e. not checking any of the configured expressions, seems a bit weird to me. This means that if the resource type has `status.observedGeneration` then it doesn't matter what healthcheck expressions you configure for it, they will never be checked. At the same time, if we implement the logic like this, the only way you can have a resource type that has `status.observedGeneration` be checked through this observed generation is by configuring a custom healthcheck for this type, which requires you to specify at least a `Current` expression, which will never be checked! This UX does not look good to me. If this is the logic that we want, then we should decouple checking observed generation from this feature.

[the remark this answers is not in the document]
matheuscscp: By that I mean: Regardless if the type has a `CustomHealthCheck` object configured for it or not, we first check the generation. If it exists and differs, then we consider the resource `InProgress`. Else, we let a custom health check for that type kick in if it was configured.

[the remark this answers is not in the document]
stefanprodan: kstatus does this already, for any CR is looks at `status.observedGeneration`, hence my suggestion to do the same when custom checks are defined.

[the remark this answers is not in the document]
matheuscscp: Cool, we can definitely do that 👍

But the configuration for that should be: nothing.

And not this:

```yaml
healthCheckExprs:
- apiVersion: my.type.with.observed.generation.io/v1
  kind: MyType
  current: whatever
```

Which is what I implied from the current state of the RFC and this discussion.

I'll update the RFC to make it clear that this configuration is not needed for types with observed generation 👌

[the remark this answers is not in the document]
matheuscscp: In other words, a custom healthcheck should only be defined for a type if `status.observedGeneration == metadata.generation` does not imply `Current`.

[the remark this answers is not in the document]
matheuscscp: Okay I understand now. The fact is that `status.observedGeneration == metadata.generation` is most of the time inconclusive for CRs, you usually have to look at conditions to determine the health. I'll try to rephrase things to make this clearer 👍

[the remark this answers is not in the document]
matheuscscp: Is it possible that `status.observedGeneration` exists and is different from `metadata.generation`, but the CR is actually `Failed` due to an unconventional design for the conditions?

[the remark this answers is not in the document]
stefanprodan: No that would be a bug in the controller, since you can't reach a final state unless the observedGeneration reaches the actual generation.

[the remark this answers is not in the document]
matheuscscp: all fixed :+1:

matheuscscp (approved):

rfcs/0009-custom-health-checks/README.md
```diff
@@ -0,0 +1,333 @@
...
+the health check will fail. Users will be encouraged to test their expressions
+in the [CEL Playground](https://playcel.undistro.io/). Here is where the community-maintained
+[library](#custom-health-check-library) will be super useful as some of the expressions might be complex.
+
+The evaluation logic will be as follows.
+
+First, we check if the custom resource has a `.status.observedGeneration` field, if it does
+we compare it with the `.metadata.generation` field to determine if the custom resource is in
+progress. We consider it in progress if these fields differ, and don't evaluate any of the
+expressions if that's the case. However, if these fields are equal there's no immediate
```
makkes, line 231: If Flux does this all the time, why does the Cluster example above define the `InProgress` CEL as `"metadata.generation != status.observedGeneration"`. Users might be led to the conclusion that they'd always have to specify that while in fact they can just leave the `InProgress` field empty for these cases.

rfcs/0009-custom-health-checks/README.md
```diff
@@ -0,0 +1,333 @@
...
+conclusion about the health of the custom resource, so we proceed with the evaluation.
+
+For each of the `InProgress`, `Failed` and `Current` expressions, we evaluate the expressions
+that are specified (`InProgress` and `Failed` are optional) in this specific order and return
+the respective status of the first expression that evaluates to `true`. If none of the
+expressions evaluate to `true`, we consider the custom resource in progress.
+
+When the `Failed` expression is not specified the controller will keep evaluating the
+`Current` expression until it returns `true`, and will give up after the timeout is reached.
```
makkes, line 240:
```suggestion
`Current` expression until it returns `true`, and will give up after the timeout defined in the Kustomization's `spec.timeout` field is reached.
```

rfcs/0009-custom-health-checks/README.md
stefanprodan: You're write, I missed to add a condition here, that it is expected to have `.status.observedGeneration` while our code skips the check if status in empty.

matheuscscp: Stefan is talking about making the Cluster `inProgress` expression above like this:

```
has(status.observedGeneration) && status.observedGeneration != metadata.generation
```

If the `.status` is not present in the object, this expression will error out with something like "status field does not exist", in which case we will stop the whole evaluation and return this error to the kstatus `StatusPoller`. In this situation, kstatus will consider that the object has the status `Unknown`, whereas our code would be able to return `InProgress` instead.

This is a limitation of CEL: While `has(status.observedGeneration)` is something that works, calling has for a top-level symbol i.e. `has(status)` returns this compilation error:

```
failed to compile the CEL expression: ERROR: <input>:9:5: invalid argument to has() macro
 | has(status)
 | ....^
```

makkes: My point is that I don't need to add `inProgress: "metadata.generation != status.observedGeneration"` at all because that's exactly what the controller does, anyway.

matheuscscp: In the specific moment in time when the object does not have `.status`, this behavior implemented directly in the controller and the evaluation of this CEL expression will be (very slightly) different, but I agree with you, the difference is so small that I'd personally prefer not setting an `inProgress` expression like this one specifically and just rely on the controller behavior.

stefanprodan: Ok we can remove `inProgress` from the use-cases

matheuscscp: fixed @makkes, thanks!

makkes (approved):

Issue #5022, opened 2024-10-03
[RFC-0010] Implementation


matheuscscp:
This issue tracks the implementation of [RFC-0010](https://github.com/fluxcd/flux2/blob/main/rfcs/0010-multi-tenant-workload-identity/README.md) ([PR](https://github.com/fluxcd/flux2/pull/5209)).

RFC patches:
* [x] EKS Pod Identity: https://github.com/fluxcd/flux2/pull/5309
* [x] Feature gate: https://github.com/fluxcd/flux2/pull/5354, https://github.com/fluxcd/flux2/pull/5355
* [x] KubeConfig: https://github.com/fluxcd/flux2/pull/5434
* [x] KubeConfig `generic` provider: https://github.com/fluxcd/flux2/pull/5452

RBAC:
* [x] https://github.com/fluxcd/flux2/pull/5332

Libraries:
* [x] Core library: https://github.com/fluxcd/pkg/pull/906
* [x] `aws`: https://github.com/fluxcd/pkg/pull/907
* [x] `azure`: https://github.com/fluxcd/pkg/pull/909
* [x] `gcp`: https://github.com/fluxcd/pkg/pull/908
* [x] Unit Tests: https://github.com/fluxcd/pkg/pull/917
* [x] Integration Tests: https://github.com/fluxcd/pkg/pull/924
* [x] Delete old code: https://github.com/fluxcd/pkg/pull/919
* [x] Cross-cloud support: https://github.com/fluxcd/pkg/pull/925, https://github.com/fluxcd/pkg/pull/928, https://github.com/fluxcd/pkg/pull/932
* [x] Feature gate: https://github.com/fluxcd/pkg/pull/927
* [x] KubeConfig: https://github.com/fluxcd/pkg/pull/960
* [x] KubeConfig for all Azure clouds: https://github.com/fluxcd/pkg/pull/988

CLI:
* [x] OCI auth: https://github.com/fluxcd/flux2/pull/5333

Controllers:
* [x] kustomize-controller
  * [x] SOPS Decryption: https://github.com/fluxcd/kustomize-controller/pull/1426
  * [x] KubeConfig: https://github.com/fluxcd/kustomize-controller/pull/1476
  * [x] KubeConfig for all Azure clouds: https://github.com/fluxcd/kustomize-controller/pull/1488
  * [x] Add docs for token cache controller flags: https://github.com/fluxcd/website/pull/2212
[...]

PR #5209, opened 2025-02-23
[RFC-0010] Multi-Tenant Workload Identity

matheuscscp:
In this RFC we aim to add support for multi-tenant workload identity in Flux, i.e. the ability to specify at the object-level which set of cloud provider permissions must be used for interacting with the respective cloud provider on behalf of the reconciliation of the object. In this process, credentials must be obtained automatically, i.e. this feature must not involve the use of secrets. This would be useful in a number of Flux APIs that need to interact with cloud providers, including all controllers except helm-controller.

Preview: https://github.com/fluxcd/flux2/blob/rfc-multi-tenant-workload-identity/rfcs/0010-multi-tenant-workload-identity/README.md

Umbrella issue for implementation: #5022

### DIFF
--- a/rfcs/0010-multi-tenant-workload-identity/README.md
+++ b/rfcs/0010-multi-tenant-workload-identity/README.md
@@ -1,0 +1,1256 @@
+# RFC-0010 Multi-Tenant Workload Identity
+
+**Status:** implementable
+
+<!--
+Status represents the current state of the RFC.
+Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+-->
+
+**Creation date:** 2025-02-22
+
+**Last update:** 2025-04-14
+
+## Summary
+
+In this RFC we aim to add support for multi-tenant workload identity in Flux,
+i.e. the ability to specify at the object-level which set of cloud provider
+permissions must be used for interacting with the respective cloud provider
+on behalf of the reconciliation of the object. In this process, credentials
+must be obtained automatically, i.e. this feature must not involve the use
+of secrets. This would be useful in a number of Flux APIs that need to
+interact with cloud providers, spanning all the Flux controllers except
+for helm-controller.
+
+### Multi-Tenancy Model
+
+In the context of this RFC, multi-tenancy refers to the ability of a single
+Flux instance running inside a Kubernetes cluster to manage Flux objects
+belonging to all the tenants in the cluster while still ensuring that each
+tenant has access only to their own resources according to the Least Privilege
+Principle. In this scenario a tenant is often a team inside an organization,
+so the reader can consider the
+[multi-team tenancy model](https://kubernetes.io/docs/concepts/security/multi-tenancy/#multiple-teams).
+Each team has their own namespaces, which are not shared with other teams.
+
+## Motivation
+
+Flux has strong multi-tenancy features. For example, the `Kustomization` and
+`HelmRelease` APIs support the field `spec.serviceAccountName` for specifying
+the Kubernetes `ServiceAccount` to impersonate when interacting with the
+Kubernetes API on behalf of a tenant, e.g. when applying resources. This
+allows tenants to be constrained under the Kubernetes RBAC permissions
+granted to this `ServiceAccount`, and therefore have access only to the
+specific subset of resources they should be allowed to use.
+
+Besides the Kubernetes API, Flux also interacts with cloud providers, e.g.
+container registries, object storage, pub/sub services, etc. In these cases,
+Flux currently supports basically two modes of authentication:
+
+- *Secret-based multi-tenant authentication*: Objects have the field
+  `spec.secretRef` for specifying the Kubernetes `Secret` containing the
+  credentials to use when interacting with the cloud provider. This is
+  similar to the `spec.serviceAccountName` field, but for cloud providers.
+  The problem with this approach is that secrets are a security risk and
+  operational burden, as they must be managed and rotated.
+- *Workload-identity-based single-tenant authentication*: Flux offers
+  single-tenant workload identity support by configuring the `ServiceAccount`
+  of the Flux controllers to impersonate a cloud identity. This eliminates
+  the need for secrets, as the credentials are obtained automatically by
+  the cloud provider Go libraries used by the Flux controllers when they
+  are running inside the respective cloud environment. The problem with
+  this approach is that it is single-tenant, i.e. all objects are reconciled
+  using the same cloud identity, the one associated with the respective controller.
+
+For delivering the high level of security and multi-tenancy support that
+Flux aims for, it is necessary to extend the workload identity support to
+be multi-tenant. This means that each object must be able to specify which
+cloud identity must be impersonated when interacting with the cloud provider
+on behalf of the reconciliation of the object. This would allow tenants to
+be constrained under the cloud provider permissions granted to this identity,
+and therefore have access only to the specific subset of resources they are
+allowed to manage.
+
+### Goals
+
+Provide multi-tenant workload identity support in Flux, i.e. the ability to
+specify at the object-level which cloud identity must be impersonated to
+interact with the respective cloud provider on behalf of the reconciliation
+of the object, without the need for secrets.
+
+### Non-Goals
+
+It's not a goal to provide multi-tenant workload identity *federation* support.
+The (small) difference between workload identity and workload identity federation
+is that the former assumes that the workloads are running inside the cloud
+environment, while the latter assumes that the workloads are running outside
+the cloud environment. All the major cloud providers support both, as the majority
+of the underlying technology is the same, but the configuration is slightly
+different. Because the differences are small we may consider workload identity
+federation support in the future, but it's not a goal for this RFC.
+
+## Proposal
+
+For supporting multi-tenant workload identity at the object-level for the Flux APIs
+we propose associating the Flux objects with Kubernetes `ServiceAccounts`. The
+controller would need to create a token for the `ServiceAccount` associated with
+the object in the Kubernetes API, and then exchange it for a short-lived access
+token for the cloud provider. This would require the controller `ServiceAccount`
+to have RBAC permission to create tokens for any `ServiceAccounts` in the cluster.
+
+### User Stories
+
+#### Story 1
+
+> As a cluster administrator, I want to allow tenant A to pull OCI artifacts
+> from the Amazon ECR repository belonging to tenant A, but only from this
+> repository. At the same time, I want to allow tenant B to pull OCI artifacts
+> from the Amazon ECR repository belonging to tenant B, but only from this
+> repository.
+
+For example, I would like to have the following configuration:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: tenant-a-repo
+  namespace: tenant-a
+spec:
+  ...
+  provider: aws
+  serviceAccountName: tenant-a-ecr-sa
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tenant-a-ecr-sa
+  namespace: tenant-a
+  annotations:
+    eks.amazonaws.com/role-arn: arn:aws:iam::123456789123:role/tenant-a-ecr
+---
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: tenant-b-repo
+  namespace: tenant-b
+spec:
+  ...
+  provider: aws
+  serviceAccountName: tenant-b-ecr-sa
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tenant-b-ecr-sa
+  namespace: tenant-b
+  annotations:
+    eks.amazonaws.com/role-arn: arn:aws:iam::123456789123:role/tenant-b-ecr
+```
+
+#### Story 2
+
+> As a cluster administrator, I want to allow tenant A to pull and push to the Git
+> repository in Azure DevOps belonging to tenant A, but only this repository. At
+> the same time, I want to allow tenant B to pull and push to the Git repository
+> in Azure DevOps belonging to tenant B, but only this repository.
+
+For example, I would like to have the following configuration:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: tenant-a-repo
+  namespace: tenant-a
+spec:
+  ...
+  provider: azure
+  serviceAccountName: tenant-a-azure-devops-sa
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tenant-a-azure-devops-sa
+  namespace: tenant-a
+  annotations:
+    azure.workload.identity/client-id: d6e4fc00-c5b2-4a72-9f84-6a92e3f06b08 # client ID for my tenant A
+    azure.workload.identity/tenant-id: 72f988bf-86f1-41af-91ab-2d7cd011db47 # azure tenant for the cluster (optional, defaults to the env var AZURE_TENANT_ID set in the controller)
+---
+apiVersion: image.toolkit.fluxcd.io/v1beta2
+kind: ImageUpdateAutomation
+metadata:
+  name: tenant-a-image-update
+  namespace: tenant-a
+spec:
+  ...
+  sourceRef:
+    kind: GitRepository
+    name: tenant-a-repo
+---
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: GitRepository
+metadata:
+  name: tenant-b-repo
+  namespace: tenant-b
+spec:
+  ...
+  provider: azure
+  serviceAccountName: tenant-b-azure-devops-sa
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tenant-b-azure-devops-sa
+  namespace: tenant-b
+  annotations:
+    azure.workload.identity/client-id: 4a7272f9-f186-41af-9f84-6a92e32d7cd0 # client ID for my tenant B
+    azure.workload.identity/tenant-id: 72f988bf-86f1-41af-91ab-2d7cd011db47 # azure tenant for the cluster (optional, defaults to the env var AZURE_TENANT_ID set in the controller)
+---
+apiVersion: image.toolkit.fluxcd.io/v1beta2
+kind: ImageUpdateAutomation
+metadata:
+  name: tenant-b-image-update
+  namespace: tenant-b
+spec:
+  ...
+  sourceRef:
+    kind: GitRepository
+    name: tenant-b-repo
+```
+
+#### Story 3
+
+> As a cluster administrator, I want to allow tenant A to pull manifests from
+> the GCS bucket belonging to tenant A, but only from this bucket. At the same
+> time, I want to allow tenant B to pull manifests from the GCS bucket
+> belonging to tenant B, but only from this bucket.
+
+For example, I would like to have the following configuration:
+
+```yaml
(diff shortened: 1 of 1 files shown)


matheuscscp:
@stefanprodan This weekend I finished testing most of the integrations for this to work across all the Flux APIs that could benefit from multi-tenant workload identity. I tested with real EKS, AKS and GKE clusters and with real resources, e.g. ECR and ACR registries, S3 bucket and Azure Event Hub. I looked into the current state of the authentication code for all the APIs and cloud providers to evaluate the effort and feasibility, and detailed how the integration would look like for all of them in the RFC. The cache key design is pretty bullet-proof, and the security questions are very well discussed and addressed. The only limitation I found, which is not a blocker for the overall project, is the lack of support from SOPS for GCP KMS through `oauth2.TokenSource`, but I already filed this [issue](https://github.com/getsops/sops/issues/1792) for addressing and plan to open a PR there next weekend.

This RFC is effectively ready for review. I'd like to unblock the review for other maintainers:

* @pjbgf We need a thorough security review :pray: 
* @hiddeco We need a feature in SOPS, please see the link above :pray: 
* @dipti-pai The details I put in https://github.com/fluxcd/notification-controller/issues/1047 and https://github.com/fluxcd/notification-controller/issues/1008 are described here in more details
* @makkes, @kingdonb PTAL :pray: :pray: :pray:

matheuscscp:
Here is a snapshot of my tests showcasing fetching tokens for each cloud provider and respective container registries through the `ServiceAccount` `default/default` in each cluster using an in-memory HTTP proxy (btw the code doesn't need to run in-cluster, it can run in my machine :smile:):

```shell
$ go run .
2025/03/09 23:37:37 [001] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:37 [001] INFO: Accepting CONNECT to sts.us-east-1.amazonaws.com:443
==== provider: aws

duration = 59m59.053550277s
role_arn = arn:aws:sts::211125720409:assumed-role/my-role/default.default.us-east-1.eks.fluxcd.io

2025/03/09 23:37:39 [002] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:39 [002] INFO: Accepting CONNECT to sts.us-east-1.amazonaws.com:443
2025/03/09 23:37:40 [003] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:40 [003] INFO: Accepting CONNECT to api.ecr.us-east-2.amazonaws.com:443
==== provider: aws/registry

duration = 11h59m59.805655685s
username = AWS
password = eyJwYXlsb2...YwNjY2MH0=

2025/03/09 23:37:41 [004] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:41 [004] INFO: Accepting CONNECT to login.microsoftonline.com:443
==== provider: azure

duration = 23h59m58.529390929s
client_id = e4c1293e-f258-496a-ad3f-6b2210988657
tenant_id = 4bd9a3a0-4393-4c05-ab26-77f6c0bfc693

2025/03/09 23:37:42 [005] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:42 [005] INFO: Accepting CONNECT to login.microsoftonline.com:443
2025/03/09 23:37:42 [006] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:43 [006] INFO: Accepting CONNECT to fluxtests.azurecr.io:443
==== provider: azure/registry

duration = 2h59m59.747754581s
username = 00000000-0000-0000-0000-000000000000
password = eyJhbGciOi...AtXWCpgEhA

2025/03/09 23:37:44 [007] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:44 [007] INFO: Accepting CONNECT to sts.googleapis.com:443
2025/03/09 23:37:44 [008] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:44 [008] INFO: Accepting CONNECT to iamcredentials.googleapis.com:443
==== provider: gcp

duration = 59m59.701158422s
email = flux-mt-wi-poc@flux-gitops-playground.iam.gserviceaccount.com

2025/03/09 23:37:45 [009] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:45 [009] INFO: Accepting CONNECT to sts.googleapis.com:443
2025/03/09 23:37:45 [010] INFO: Running 0 CONNECT handlers
2025/03/09 23:37:45 [010] INFO: Accepting CONNECT to iamcredentials.googleapis.com:443
==== provider: gcp/registry

duration = 59m59.033109082s
username = oauth2accesstoken
password = ya29.c.c0A...1c-BejMa-h
$ 
```

FabianTek:
Hi @matheuscscp,

I'm really looking forward to this, and guess this would adress https://github.com/fluxcd/flux2/issues/5022

I haven't seen you mention flux bootstrapping. Currently, I can use Username/Password, PAT or SSH keys during bootstrapping and flux will update the repository with its two commits as well as generate the secret in the flux-system namespace also containing the authentication data for future reconciliation.

With potentially upcoming workload identity support, I could see the initial _flux-system_ GitRepository also referencing a dedicated service account with workload identity annotations. There is a difference though. Through Username/Password, PAT or SSH keys a user and flux source controller can impersonate the same identity. When using workload identity to reconcile the GitRepository it's unlikely for the user or pipeline agent to be able to impersonate the desired workload identity. 

-> Users / Pipeline Agents could still leverage e.g. a PAT for the initial commits
-> Source-Controller would be configured to use the workload identity

This is similar to e.g. [GitHub Deploy Keys](https://fluxcd.io/flux/installation/bootstrap/github/#github-deploy-keys), where you decouple the initial identity from the source-controller identity.


Is this something you see aligned with the _goals_ of this RFC or rather a _non-goal_?

matheuscscp:
My PR in SOPS got merged :tada: https://github.com/getsops/sops/issues/1792

Now waiting for a release

pjbgf (commented):
@matheuscscp good job putting this together. From a high-level it is a great feature and would further bolster Flux's least privileged model.

I added a few nits, but overall multi-tenancy is a charged term, so it is quite important to define what precisely we mean by that and what guarantees we are providing as part of this RFC.

My take is that this supports out-of-the-box the [multiple teams](https://kubernetes.io/docs/concepts/security/multi-tenancy/#multiple-teams) model. The [multiple customers](https://kubernetes.io/docs/concepts/security/multi-tenancy/#multiple-customers) model is achievable, but requires stronger security boundaries at cluster topology level - for coping with more hostile environment.

It would be nice to explicitly call out in the RFC:
1. The cache keys format, so that it is crystal clear that it will be resistant to cross-tenant takeovers. As you already pointed out, when sharing the same cache storage, the cache key will require some level of tenant or object level information to provide isolation. What in the format picked will result in a malicious tenant not being able to forge it (e.g. the fully qualified service account name plus x, y and z)?
2. What security controls are in place (or can be implemented by the user) to ensure that even if `tenant A` knows the Cloud Provider annotations for `tenant B`, it won't be able to impersonate them. 
  - This can be as simple as they enforcing rules via Admission Controllers (e.g. [Kubewarden](https://www.kubewarden.io/)), as per some examples of [Flux multi-tenancy](https://github.com/fluxcd/flux2-multi-tenancy). Or something more sophisticated as part of the new feature.
3. The limitations of this feature for multi-tenancy in hostile environments where the tenants are not trustworthy. Suggestions on how to overcome them is optional, as that can be a larger topic - e.g. stronger isolation so that tenants can't impersonate each other's identities even if they know them and can bypass controls that may be operating at a degraded state (i.e. admission controllers).

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -0,0 +1,1195 @@
...
+helm-controller does not need this feature, and a significant amount of
+dependencies for talking to cloud providers would be pulled in unnecessarily.
+
+This library should support both single-tenant and multi-tenant workload
+identity because single-tenant implementations are already supported in
+GA APIs and hence they must remain available for backwards compatibility
+with a large base of users. Furthermore, it would be easier to support
+both use cases in a single library as opposed to mingling a new library
+into the currently existing ones, so this new library should be a
+definitive unified solution for workload identity in Flux.
```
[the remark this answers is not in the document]
matheuscscp: I think "single source of truth" is more associated with data/state/configuration rather than code :thinking: I think I prefer the current wording in this case

matheuscscp:
New SOPS version 3.10.0 released with the GCP KMS `oauth2.TokenSource` authentication method and already bumped in kustomize-controller:

https://github.com/fluxcd/kustomize-controller/pull/1410

rfcs/0010-multi-tenant-workload-identity/README.md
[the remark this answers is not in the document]
matheuscscp: I changed "should be a" to "becomes the", though, that looks good :+1:

dipti-pai (approved):

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -22,6 +22,16 @@ of secrets. This would be useful in a number of Flux APIs that need to
...
 
+### Multi-tenancy
+
+In the context of this RFC, multi-tenancy refers to the ability of a single
+Flux instance running inside a Kubernetes cluster to manage Flux objects
+belonging to all the tenants in the cluster while still ensuring that each
+tenant has access only to their own resources according to the Least Privilege
+Principle. In this scenario a tenant is often a team inside an organization,
+so the reader can consider the
+[multi-team tenancy model](https://kubernetes.io/docs/concepts/security/multi-tenancy/#multiple-teams).
```
pjbgf, line 33:
```suggestion
Principle. In this scenario a tenant is often a team inside an organization,
so the reader can consider the
[multi-team tenancy model](https://kubernetes.io/docs/concepts/security/multi-tenancy/#multiple-teams). Each tenant has their own namespaces, which are not shared with other teams.
```

matheuscscp:
I'm now addressing the offline comments I got during KubeCon EU 2025.

From @stealthybox:

> An alternative to using service account tokens would be using a token whose subject string encodes a direct reference to the respective Flux resource, this way a resource would be its own identity. This is more secure than having a configuration knob to define another resource (a service account in this case) as the object identity, as it avoids another object in the same namespace abusing the same service account.

From @hiddeco:

> We should move the interfaces in `interfaces.go` to multiple files with names matching the interface names themselves.

From @stefanprodan:

> To avoid introducing `kustomizations.spec.decryption.key.serviceAccountName` for disambiguating single-tenant and multi-tenant workload identity we should instead introduce a binary flag in the controllers to switch to and enforce multi-tenant workload identity, e.g. `--require-service-account-for-provider-auth`

matheuscscp:
@stealthybox @hiddeco @stefanprodan Comments from KubeCon addressed, please feel free to do another pass/approve.

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -0,0 +1,1268 @@
...
+to use an HTTP/S proxy. Most Flux APIs already support the field
+`spec.proxySecretRef` for accessing the respective cloud resources, and some
+already use it for talking to the STS service in the single-tenant version
+of workload identity as well, e.g. `Bucket`. We propose using this field for
+specifying the proxy to use when talking to the STS service also in the
+multi-tenant version of workload identity.
+
+Finally, the `Kustomization` API uses Key Management Services (KMS) for
+decrypting SOPS-encrypted secrets. We propose adding the optional field
+`spec.decryption.key` with the required sub-field `.provider`: One of
```
stefanprodan, line 604: This does not match the Kustomization API from use-cases where we use `spec.decryption.keyProvider`. I think `keyProvider` is better than having `spec.decryption.key.provider`.

matheuscscp: Fixed, thanks!

pjbgf (approved):
LGTM

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -0,0 +1,1270 @@
...
+section). If only the `spec.provider` field is present, then the controller would
+use single-tenant workload identity, i.e. the identity of the controller itself.
+
+Alongside `spec.serviceAccountName` in the Flux APIs, we propose adding the binary
+flag `--use-service-account-for-provider-auth`. By default this flag would be set
+to `false`, meaning that the controller would use single-tenant workload identity
+for all the objects, i.e. the identity of the controller itself. When set to `true`,
+workload identity would require the `spec.serviceAccountName` field to be present
+in the object, and the controller would use the identity of the `ServiceAccount`
+specified in the object for accessing the cloud provider resources.
```
stefanprodan, line 579: This would break how multi-tenancy works now in kustomize-controller and helm-controller which have a `--default-service-account` flag for setting the SA when `spec.serviceAccountName` is not specified. For these two controllers we can't make `spec.serviceAccountName` required, maybe the best way would be to add `--default-service-account` to all SC, NC, IAC, IRC as well.

matheuscscp: But I see your point, if you enable the flag and specify `spec.decryption.keyProvider` then now you are forced to add a service account and configure RBAC for it. Maybe we should have separated SA fields for `Kustomization` after all.

matheuscscp: As discussed via DM, we're doing the following changes:

In SC, NC, IRC and IAC we will introduce the opt-in feature gate `EnableProviderServiceAccount`. This gate must be enabled for using MT WI when both `spec.provider` and `spec.serviceAccountName` are specified. In future versions this gate will become opt-out.

In KC we will have `spec.decryption.serviceAccountName` to specify a SA for being used with a cloud provider KMS service.

hiddeco (approved):
🚀

rfcs/0010-multi-tenant-workload-identity/README.md
matheuscscp: (By the way I just noticed we don't have `spec.serviceAccountName` for specifying an imagePullSecret in the `HelmRepository` API for OCI charts, we only have `spec.secretRef`)

matheuscscp: Hmm I think we don't need the feature gate. If a user specifies a service account containing image pull secrets in the `OCIRepository` and `ImageRepository` APIs, their intention is clearly to use that secret for authenticating with the container registry, the `spec.serviceAccountName` is not used for anything else at the moment.

Even better, the documentation states that `spec.provider` is specifically for *secret-less* authentication! Image pull secrets from either `spec.secretRef` or from `spec.serviceAccountName` are only used when `spec.provider` is set to `generic` (or left empty, when it defaults to `generic`).

In other words, the value of `spec.provider` fully determines what `spec.serviceAccountName` should be used for! We don't need a feature gate!

matheuscscp: I updated the RFC with all of this, thanks!

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -0,0 +1,1251 @@
...
+
+The `Kustomization` API uses Key Management Services (KMS) for decrypting
+SOPS-encrypted secrets. We propose adding the dedicated optional field
+`spec.decryption.serviceAccountName` for multi-tenant workload identity
+when intercting with the KMS service. We choose having a dedicated field
+for the `Kustomization` API because the field `spec.serviceAccountName`
+already exists and is used for a major part of the functionality which
+is authenticating with the Kubernetes API when applying resources. If
+we used the same field for both purposes users would be forced to use
+multi-tenancy for both cloud and Kubernetes API interactions.
```
matheuscscp, line 588: Putting it like this, it actually doesn't sound like a bad idea after all. Why would you want to use multi-tenancy for Kubernetes RBAC but not for cloud? Or the opposite? It's hard for me to imagine a decent use case where multi-tenancy should apply only to one of those two. That said, users currently using multi-tenancy for RBAC would suddenly be forced to use it also for cloud if we used the same field for both things, and that would be a really bad breaking change.

With all of this in mind, if there is no justifiable use case of multi-tenancy only for RBAC or only for cloud, I propose we use the existing `spec.serviceAccountName` and existing `--default-service-account`, and introduce also a feature gate to avoid the breaking change, e.g. `UseServiceAccountForDecryption`. This feature gate can start opt-in and then change to opt-out later.

But I'm not so much hell bent on this idea, if we can find arguments or use cases of multi-tenancy only for RBAC or only for cloud I'm totally fine with giving up on this

rfcs/0010-multi-tenant-workload-identity/README.md
```diff
@@ -0,0 +1,1251 @@
...
+
+In the `Bucket` API we have previously introduced the optional field `spec.sts`.
+This field has configuration for interacting with the STS service of the cloud
+provider, most specifically the `spec.sts.endpoint` field. In the specific case
+of the `Bucket` API, there's also the field `spec.sts.provider`, which allows
+using other STS configurations relavant only to the `generic` bucket provider,
+whose underlying implementation uses the MinIO SDK (the AWS S3-compatible object
+storage server). The same does not apply to the other Flux APIs, so we propose
+introducing only a single optional field `spec.stsEndpoint` instead.
```
matheuscscp, line 597: @stefanprodan please review this part, I think `spec.stsEndpoint` is better for the other APIs, `spec.sts` makes sense only for Bucket

rfcs/0010-multi-tenant-workload-identity/README.md
matheuscscp: Hmmm but people could be using different cloud identities for decryption through secrets and not be using multi-tenant RBAC... Would that justify multi-tenancy for cloud but not for RBAC? :thinking: :thinking: :thinking:

stefanprodan: A common use-case is where the tenants are apps made by the same team. Since apps should never be reconciled as cluster-admin, Flux is configured with ` --default-service-account`. The team uses a single KMS instance for all apps secrets and they would be now forced into creating IAM role binds for all apps SAs in all namespaces. If we chose to reuse `spec.serviceAccountName` and the default SA setting, we can never switch the `UseServiceAccountForDecryption` to `true` as it would break this commonly use-case where a team has a shared KMS for all apps running in a cluster.

matheuscscp: Okay, let's forget this idea then :+1:

rfcs/0010-multi-tenant-workload-identity/README.md
matheuscscp: (By the way, the only provider that supports this is AWS. Azure and GCP don't support custom STS endpoints.)

stefanprodan: Ok let's not add the sts setting in here. Let's see if anyone needs it after we implement the SA.

matheuscscp: We don't need to implement this until somebody asks for it, but since we already have `spec.sts.endpoint` for the Bucket API anyway maybe we should at least agree on a plan here for when somebody asks for it, so I updated the RFC with what we decided here. Please let me know if you prefer otherwise (removing this paragraph about STS endpoint altogether and the update I just added to the RFC)

matheuscscp: Paragraph removed as discussed via DM :ok_hand:

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

Poltergeisen:
@matheuscscp was this ever successfully added to bootstrap command? or is there an issue i can follow to track that?

stefanprodan:
@Poltergeisen this RFC has no proposed changes for the CLI nor the bootstrap procedure. Enabling multi-tenancy during bootstrap is the same as before: https://fluxcd.io/flux/installation/configuration/multitenancy/

Once we have complemented the RFC-0010 implementation, we'll document the new controller flags on that page.

PR #5287, opened 2025-04-03
add: OSSF scorecard configuration file - ignore false-positive

### LINKED ISSUE #5257: Draft a recommendation for fixing the failing Scorecard check of `Dangerous-Workflow`
According to [OpenSSF Scorecard report](https://scorecard.dev/viewer/?uri=github.com/fluxcd/flux2) the GitHub Workflow we use for backporting changes made with PRs to `release/vX.Y` branches is unsecure.

https://github.com/fluxcd/flux2/blob/main/.github/workflows/backport.yaml

```
Reason
dangerous workflow patterns detected

Details
Warn: untrusted code checkout '${{ github.event.pull_request.head.sha }}': .github/workflows/backport.yaml:18
``` 

We should assess this report and find a way to tell the OpenSSF that this workflow does not run untrusted code, the bot will only backport if the a PR was approved and merged into main. Alternatively  we could find a diffrent backport bot that doesn't trigger the OpenSSF scanner.

piontec:
According to the [docs](https://github.com/ossf/scorecard/tree/main/config), we should be able to put in place annotations about why the check should be ignored. Unfortunately, I failed to confirm this with a local run - maybe it will work fine when it runs for a real repository.
closes #5257

### DIFF
--- a/.scorecard.yml
+++ b/.scorecard.yml
@@ -1,0 +1,5 @@
+annotations:
+  - checks:
+      - dangerous-workflow
+    reasons:
+      - reason: not-applicable # This workflow does not run untrusted code, the bot will only backport a code if the a PR was approved and merged into main.


matheuscscp:
Thanks @piontec!

stefanprodan (approved):
Awesome

PR #5292, opened 2025-04-08
[RFC-0012] External Artifact API

stefanprodan:
This RFC proposes the introduction of a new API called `ExternalArtifact` that would allow 3rd party controllers to act as a source of truth for the cluster desired state. In effect, the `ExternalArtifact` API acts as an extension of the existing `source.toolkit.fluxcd.io` APIs that enables Flux `kustomize-controller` and `helm-controller` to consume artifacts from external source types that are not natively supported by `source-controller`.

Preview: https://github.com/fluxcd/flux2/blob/rfc-external-artifact/rfcs/0012-external-artifact/README.md

### DIFF
--- a/rfcs/0000-external-artifact/README.md
+++ b/rfcs/0000-external-artifact/README.md
@@ -4,7 +4,7 @@
 
 **Creation date:** 2025-04-08
 
-**Last update:** 2025-07-15
+**Last update:** 2025-08-30
 
 ## Summary
 
@@ -161,6 +161,53 @@ spec:
   values:
     replicaCount: 2
 ```
+
+### Security Considerations
+
+With the introduction of the `ExternalArtifact` API, the trust boundary of Flux is extended
+to include 3rd party controllers that are capable of creating and managing `ExternalArtifact`
+custom resources in the cluster. This means that the security posture of the cluster
+is now dependent on the security of these 3rd party controllers.
+
+To mitigate potential security risks, it is recommended to implement the following measures 
+when developing 3rd party source controllers:
+
+- **Authentication and Authorization**: Ensure that the controller uses proper authentication
+  and authorization mechanisms to interact with upstream sources and avoid embedding sensitive
+  information directly in the custom resource specifications. Following source-controller
+  best practices for managing credentials is highly recommended: use `serviceAccountName` to
+  integrate with Kubernetes Workload Identity for short-lived credentials, use `secretRef` to
+  reference long-lived credentials, never cache long-lived credentials on disk or in-memory.
+- **TLS Encryption**: Use TLS encryption for all communications between the controller
+  and upstream sources to protect sensitive data in transit. Following source-controller
+  best practices for TLS is highly recommended: use `certSecretRef` to reference
+  custom CA certificates and client certificates, prefer Mutual TLS authentication, never
+  allow skipping TLS verification.
+- **Provenance and Integrity**: Ensure that the controller verifies the integrity of the
+  artifacts it generates and exposes in-cluster. This can be achieved by using checksums
+  and digital signatures to validate the authenticity of upstream sources. Following
+  source-controller best practices for source integrity is highly recommended:
+  verify the provenance of upstream artifacts using Sigstore Cosign or Notary
+  Notation signatures, prefer keyless verification using OIDC identity tokens and
+  public transparency logs.
+- **Access Control**: Implement access control mechanisms to restrict cross-namespace
+  generation of `ExternalArtifact` custom resources. Following source-controller
+  best practices for access control is highly recommended: expose a `--no-cross-namespace-refs`
+  flag to restrict the controller from generating `ExternalArtifact` resources in a different
+  namespace than the one where the source custom resource is located. Use Kubernetes owner
+  references to establish a clear ownership relationship between the source custom resource
+  and the `ExternalArtifact` resource, allowing Kubernetes garbage collection to clean up
+  the `ExternalArtifact` when the source resource is deleted.
+- **Least Privilege**: Run the controller with the least privilege necessary to perform
+  its functions. Following source-controller best practices for least privilege is highly recommended:
+  use a dedicated Kubernetes service account with minimal RBAC permissions, avoid running
+  the controller as a cluster-admin or with wildcard permissions, conform with the restricted pod security
+  standard (e.g., disallow running as root, disallow host network access, read-only rootfs).
+- **Artifact persistent storage integrity**: Ensure that the controller can be configured to use
+  persistent storage for storing artifacts, to avoid data loss in case of controller restarts
+  or failures. Following source-controller best practices for artifact storage is highly recommended:
+  at startup, ensure that the artifacts in-storage have not been tampered with by verifying
+  the checksums of all stored artifacts against the `ExternalArtifact` digests in the cluster.
 
 ### User Stories
 
--- a/rfcs/0000-external-artifact/README.md
+++ b/rfcs/0000-external-artifact/README.md
@@ -1,0 +1,166 @@
+# RFC External Artifacts
+
+**Status:** provisional
+
+**Creation date:** 2025-04-08
+
+**Last update:** 2025-04-08
+
+## Summary
+
+This RFC proposes the introduction of a new API called `ExternalArtifact` that would allow
+3rd party controllers to act as a source of truth for the cluster desired state. In effect,
+the `ExternalArtifact` API acts as an extension of the existing `source.toolkit.fluxcd.io` APIs
+that enables Flux `kustomize-controller` and `helm-controller` to consume artifacts from external
+source types that are not natively supported by `source-controller`.
+
+## Motivation
+
+Over the years, we've received requests from users to support other source types besides the
+ones natively supported by `source-controller`. For example, users have asked for support of
+downloading Kubernetes manifests from GitHub/GitLab releases, Omaha protocol, SFTP protocol,
+and other remote storage systems.
+
+Another common request is to run transformations on the artifacts fetched by source-controller.
+For example, users want to be able to generate YAML manifests from jsonnet, cue, and other
+templating engines before they are consumed by Flux `kustomize-controller`.
+
+In order to support these use cases, we need to define a standard API that allows 3rd party
+controllers to expose artifacts in-cluster (in the same way `source-controller` does)
+that can be consumed by Flux  `kustomize-controller` and `helm-controller`.
+
+### Goals
+
+Define a standard API for 3rd party controllers to expose artifacts that can be consumed by
+Flux controllers in the same way as the existing `source.toolkit.fluxcd.io` APIs.
+
+Allow Flux users to transition from using `source-controller` to using 3rd party source controllers
+with minimal changes to their existing `Kustomizations` and `HelmReleases`.
+
+### Non-Goals
+
+Allow arbitrary custom resources to be referenced in Flux `Kustomization` and `HelmRelease` as `sourceRef`.
+
+Extend the Flux controllers permissions to access custom resources that are not part of the
+`source.toolkit.fluxcd.io` APIs.
+
+## Proposal
+
+Assuming we have a custom controller called `release-controller` that is responsible for
+reconciling `GitHubRelease` custom resources. This controller downloads the Kubernetes
+deployment YAML manifests from the GitHub API and stores them in a local file system
+as a `tar.gz` file. The `release-controller` then creates an `ExternalArtifact`
+custom resource that tells the Flux controllers from where to fetch the artifact.
+
+Every time the `release-controller` reconciles a `GitHubRelease` custom resource,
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
Hey there 👋 Thanks for sharing this in Draft State, cool to see such a development!
(FYI @Skarlso @fabianburth who also worked on the proposal)

This looks very similar to https://github.com/openfluxcd / https://github.com/fluxcd/flux2/discussions/5058 original proposal with the main difference being that you use the ExternalArtifact instead of any arbitrary SourceRef.

I am wondering how this would be different to having a CRD that pulls OCI artifacts and then referencing those...

Maybe its even worth thinking about standardizing ExternalArtifacts completely to OCI? I think we talked about this in the Flux community call when the original proposal of openflux was discussed.

The main drawback we mentioned back then still persists:
If we have an ExternalArtifact, Flux will always need to reference ExternalArtifact, and users will not be able to use their origin CRDs. In your Example `release-controller` would work on a `GitHubRelease` but the Kustomization still contains `ExternalArtifact`.

At this point, why not think about having `release-controller` directly download OCIArtifacts, expose a registry endpoint, and then use the already available Flux types?

This is ultimately what caused us to not follow up on the design.

If there are cases when working with an OCIRegistry is not desired, or unfeasible (e.g. working with SFTP and converting that to OCI Artifacts may be unfeasible / undesirable), then I still see value in this CRD even with the indirection in place!

Looking forward to see where this goes! 🎉

stefanprodan:
> If we have an ExternalArtifact, Flux will always need to reference ExternalArtifact, and users will not be able to use their origin CRDs. 

@jakobmoellerdev an arbitrary sourceRef would imply opening up RBAC and the Flux CRD to unknown kinds, which would imply a v2 of Flux Kustomization and a v3 of HelmRelease. With `ExternalArtifact` the Flux RBAC does not changes, as Flux controllers already have all the permissions needed for the `source.toolkit.fluxcd.io` APIs and instead of dropping the allow list in the Flux CRDs we will just add another owned Kind to that which does not changes the security stance of Flux.

> At this point, why not think about having release-controller directly download OCIArtifacts, expose a registry endpoint, and then use the already available Flux types?

Hosting an OCI registry inside the cluster is way more challenging and involving than embedding a Go file server in a custom controller. Given that source-controller contains a [storage](https://github.com/fluxcd/source-controller/blob/main/internal/controller/storage.go) implementation that can easily be copied in any other controller, the effort to expose artifacts to Flux is significant lower. If running a registry inside the cluster is feasible for you, then using `OCIRepository` is right way to go, this RFC is for offering an alternative solution which IMO is easier to implement for 3rd party controllers.

matheuscscp:
> There may be other reasons why we want a shim API within our own namespace (maybe so Notification controller only has to watch a single Kind of list), but RBAC may not be a sufficient reason on its own.

Not sure what you mean by "NC watching a single kind of list", but NC does not watch any resources :thinking: The other Flux controllers send the event payload to NC via HTTP and external webhooks cause the Receiver controller to annotate the Flux resources for requesting out-of-schedule reconciliations. But all of that can be kustomized to work with Flux-external types by just patching the CRDs and RBAC like we do for flux-operator, no code changes are required:

https://github.com/controlplaneio-fluxcd/flux-operator/blob/07a37e692ed2d7ee6b16cd3c8f8231a134a9fbdc/internal/builder/profiles.go#L65-L138

External source controller authors must document how to do this :point_up_2: for their controllers/CRDs.

> There could be separate security concerns with the code of multiple controller authors all having write access to the same status objects of a shared shim API. ex: if ControllerA for SourceTypeA writes to the status of ExternalArtifacts for SourceTypeB, a flux controller could be confused into reconciling an incorrect or malicious set of resources.

Not sure I follow this example. I think the proposal here is that ControllerA for SourceTypeA will create an ExternalArtifact object with ownerRef/controllerRef set to an object of SourceTypeA, and only ControllerA will manage this ExternalArtifact object, no other controllers will. Each external controller will manage its own source types and ExternalArtifact object instances for its source types. If you install a malicious controller in the cluster that messes with it, I think that's on you, and that's the risk of opening up Flux for external source controllers regardless of particular details of the implementation we choose?

matheuscscp:
Actually, thinking again, I think I understand your concern. By accepting kinds that are external to Flux directly in `.sourceRef` cluster admins can have a fine control over what kinds are allowed in the Kustomization CRD, while with a single built-in kind `ExternalArtifact` this kind of fine control is not possible. I wonder, however, how much protection this really brings e.g. if a malicious controller starts patching the status of a CRD object whose kind is allowed in the Kustomization CRD, I suppose the honest controller will start fighting with the malicious controller and reconcile the object, then the malicious controller will reconcile again, and so on, so the protection we're providing here is actually this DDoS on the objects? :thinking:

phoban01:
cc @mandelsoft 🙂

stefanprodan:
@matheuscscp @stealthybox I've added a policy example for protecting a cluster against malicious tenants trying to hijack `ExternalArtifacts`.

rfcs/0000-external-artifact/README.md
```diff
@@ -0,0 +1,260 @@
...
+  namespace: apps
+spec:
+  # SourceRef points to the Kubernetes custom resource for
+  # which the artifact is generated.
+  # +required
+  sourceRef:
+    apiVersion: source.example.com/v1alpha1
+    kind: GitHubRelease
+    name: podinfo
+    namespace: apps
```
matheuscscp: I'd prefer not to add a `namespace` field here. This is my only concern, everything else LGTM

stefanprodan: This is standard sourceRef from Flux APIs which by default allow cross-namespace source access.

matheuscscp: Will kustomize-controller and helm-controller validate this spec and enforce the --no-cross-namespace-refs? Why do we even need this field (I assume it's for flux trace)? Can't the external controller simply add annotations it deems relevant for source tracking? What if the external source is not even associated with a Kubernetes object? To me it seems that Flux should care only about status.artifact

stefanprodan: Yes this is for `flux trace` and other tools that would discover the origin source & lookup its ready status. The `ExternalArtifact` is not meant to be created by users, so there must be some other CR backing it.

matheuscscp: Actually, it makes immense sense not to add a namespace field here. Think about Flux source APIs. They report the status of the source in the status field. The status field is a subresource that is tightly coupled to the object, it's a part of the object. Now imagine a world where Flux was initially designed with an Artifact API and the Flux source APIs would report the status in an Artifact object. That Artifact object would **always** be in the same namespace of the original Flux source object that produced it. The same reasoning can applied to an external source controller. Such a controller would never have a reason to report an ExternalArtifact object in a namespace different than the one where the external source object producing this ExternalArtifact object lives.

Furthermore, by not adding a namespace field here, this spec matches the spec of Kubernetes RBAC APIs, where the scope is defined by the kind. External source controllers could implement cluster-scoped kinds.

matheuscscp: And `flux trace` can figure out whether or not the reference is cluster-scoped using the apiVersion and kind as well, so all good. I'd not add namespace in the first iteration of this feature. If somebody shows up with a use case that makes a lot of sense to support, we can consider it.

stefanprodan: The use-case I have in mind is the same as with Flux HR, you have a `targetNamespace` / `storageNamespace` in the CR and you generate the artifact there.

stefanprodan: I made the `sourceRef` optional, if a 3rd party controller choses not to use it, then `flux trace` will fail.

matheuscscp: If we're gonna add the namespace field here then we need to enforce the --no-cross-namespace-refs flag on kustomize-controller and helm-controller when consuming the artifact and when the sourceRef field is set. The reconciliation must fail if spec.sourceRef.namespace is set to a value different than metadata.namespace.

`flux trace` will still need to reflect the type definition to determine whether or not to fill the optional namespace field of the sourceRef with the object namespace. If the type is namespaced and the namespace is empty, for a good debugging experience we should fill the empty namespace field with the namespace of the ExternalArtifact.

matheuscscp: Can we somehow document for users showing how to restrict cross-namespace references in the `ExternalArtifact` API field `.spec.sourceRef` using policy controller tools, e.g. the built-in one? If yes, I'm happy with this RFC as is.

matheuscscp: For external source controller authors, we should create [here](https://fluxcd.io/flux/guides/) a page called `Authoring an external source controller` including all these major guidelines, `--no-cross-namespace-refs` should be there, very visible.

matheuscscp (approved):
LGTM

bgrant0607:
The source controller seems like the most needed extensible integration point in Flux. I'd even put other configuration format renders as lower priority, because other kinds of automation, notably CI tools, can perform rendering (c.f., the rendered manifest pattern). Being able to pull configuration from arbitrary repositories, cache the files persistently, and then apply from the unpacked files would open up new possibilities.

matheuscscp:
> The source controller seems like the most needed extensible integration point in Flux. I'd even put other configuration format renders as lower priority, because other kinds of automation, notably CI tools, can perform rendering (c.f., the rendered manifest pattern). Being able to pull configuration from arbitrary repositories, cache the files persistently, and then apply from the unpacked files would open up new possibilities.

The ideal solution for artifacts produced in CI is `OCIRepository`. The use case of this RFC is for when using `OCIRepository` is impossible.

ebourgeois:
> > The source controller seems like the most needed extensible integration point in Flux. I'd even put other configuration format renders as lower priority, because other kinds of automation, notably CI tools, can perform rendering (c.f., the rendered manifest pattern). Being able to pull configuration from arbitrary repositories, cache the files persistently, and then apply from the unpacked files would open up new possibilities.
> 
> The ideal solution for artifacts produced in CI is `OCIRepository`. The use case of this RFC is for when using `OCIRepository` is impossible.

That is correct, that is what we are thinking and needing.

stefanprodan:
@bgrant0607 @ebourgeois are you using some kind of proprietary storage and you can't run `flux push artifact` in your system? Trying to understand why would you need a 3rd-party source instead of using a container registry.

bgrant0607:
@stefanprodan @matheuscscp 

Correct, we have other storage. A more generic HTTPS client than the S3 implementation which supported more general authentication methods might also work.

Another issue is that we're breaking up configuration artifacts into more granular pieces and haven't found an acceptable way to package and deploy them as OCI artifacts for Flux. It doesn't look like it supports applying multiple layers from a single image or watching a whole repo prefix, for instance.

S3 bucket is better in that regard because they can be pushed individually but reconciled as a group. S3 has other issues, though. For instance, I only just started to look at it, but I didn't see a way to sign artifacts and verify signatures.

We need to pull groups of files from our storage, persistently cache the files, and verify signatures before applying them.

matheuscscp:
> Correct, we have other storage.

Where do you store the container images of your apps?

matheuscscp:
> S3 bucket is better in that regard because they can be pushed individually but reconciled as a group.

Not sure I consider this an advantage

bgrant0607:
> Where do you store the container images of your apps?

This is not a point you need to debate.
https://itnext.io/advantages-of-storing-configuration-in-container-registries-rather-than-git-b4266dc0c79f

We have N files that are updated independently. We want O(1) configuration complexity in the deployment mechanism. Right now that has N^2 cost to update if we pack them into a single OCI image but update them independently. If we could add them as layers, this would not be an issue. 

We could cache them in some other storage and push a new image after they are all updated, but at that point we don't need the OCI repository. The history is redundant and a management burden. We're trying to remove dominos from the Rube Goldberg Machine.

I have been an open-source maintainer and understand the burden of accepting new features. Good extension APIs reduce maintainer burden by pushing whole categories of features into the ecosystem. Honestly with the toolkit approach of Flux v2 I was surprised to find that pluggable source controllers weren't supported already. 

We could use an alternative to Flux, but since there was a proposal that appeared to address our needs, I wanted to support it and find out if it was going anywhere. I'm not attached to this specific proposal, however.

cappyzawa (approved):
LGTM

rfcs/0012-external-artifact/README.md
```diff
@@ -0,0 +1,321 @@
...
+  interval: 10m
+  releaseName: podinfo
+  chartRef:
+    kind: ExternalArtifact
+    name: podinfo
+  values:
+    replicaCount: 2
+```
+
+### Security Considerations
```
souleb, line 165: Can we also have a paragraph about recommendations on how to secure the access to the artifact storage endpoint?

stefanprodan: Added, thanks for the suggestion!

PR #5297, opened 2025-04-11
Add an en var to use the namespace declared in the kubeconfig context

jybp:
Allow for `flux` to work seamlessly with tools like `kubens` to avoid having to specify `-n`.

Adds `FLUX_SYSTEM_NAMESPACE_FROM_CONTEXT=true` as suggested in https://github.com/fluxcd/flux2/discussions/2768 to avoid any breaking changes.

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -206,19 +206,7 @@ func main() {
 func configureDefaultNamespace() {
 	*kubeconfigArgs.Namespace = rootArgs.defaults.Namespace
 
-	// First check if we should use the namespace from the current context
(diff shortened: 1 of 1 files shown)


matheuscscp:
There's been more discussions around this, @kingdonb do you remember which issue was it/which conclusions we reached?

kingdonb:
Yes: https://github.com/fluxcd/flux2/pull/5028

jybp:
Added my 2c on #3453

jybp:
Welp dupe https://github.com/fluxcd/flux2/pull/5831

PR #5299, opened 2025-04-11
Allow to pull/push artifacts to insecure registries without TLS

mottetm:
This PR relates to the [following discussion](https://github.com/fluxcd/flux2/discussions/5296#discussioncomment-12801631).

If merged, this PR will introduce a new `--insecure-registry` flag to the following commands: `push artifacts`, `pull artifact`, `diff artifact` and `list artifacts`. When used the flag will lead to the option `crane.Insecure` being passed to the `crane` client allowing the use of insecure repositories.

### DIFF
--- a/cmd/flux/diff_artifact.go
+++ b/cmd/flux/diff_artifact.go
@@ -23,6 +23,7 @@ import (
 
 	oci "github.com/fluxcd/pkg/oci/client"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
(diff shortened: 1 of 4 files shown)


stefanprodan:
Can you please update the PR description and correct the flag name it’s `insecure-registry`

mottetm:
fixed 🤦

stefanprodan (approved):
LGTM

Thanks @mottetm

PR #5302, opened 2025-04-15
Reset target struct before decoding

maboehm:


stefanprodan (approved):
LGTLM

Thanks @maboehm 🏅

PR #5321, opened 2025-04-24
[RFC-0011] OpenTelemetry Tracing

adri1197:
Specification for fluxcd/notification-controller#1097

### DIFF
--- a/rfcs/0011-opentelemetry-tracing/README.md
+++ b/rfcs/0011-opentelemetry-tracing/README.md
@@ -1,0 +1,218 @@
+# RFC-0011: OpenTelemetry Tracing
+
+**Status:** provisional
+
+<!--
+Status represents the current state of the RFC.
+Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+-->
+
+**Creation date:** 2025-04-24
+
+**Last update:** 2025-08-13
+
+## Summary
+The aim is to be able to collect traces via OpenTelemetry (OTel) across all Flux related objects, such as HelmReleases, Kustomizations and among others. These may be sent towards a tracing provider where may be potentially stored and visualized. Flux does not have any responsibility on storing and visualizing those, it keeps being completely stateless. Thereby, being seamless for the user, the implementation is going to be part of the already existing `Alert` API Type. Therefore, `EventSources` is going to discriminate the events belonging to the specific sources, which are going to be looked up to and send them out towards the `Provider` set. In this way, it could facilitate the observability and monitoring of Flux related objects.
+
+## Motivation
+This RFC was born out of a need for end-to-end visibility into Flux’s multi-controller GitOps workflow. At the time Flux was one monolithic controller; it has since split into several specialized controllers (source-, kustomize-, helm-, notification-, etc.), which makes tracing the path of a single "Source change → applied resource → notification” much harder. Additionally, users may not have to implement tools/sidecars around to maintain.
+
+Correlate any potential source (GitRepository, OCIRepository, HelmChart or Bucket) with all downstream actions. Therefore, you would like to see a single trace (with multiple spans underneath):
+- Alert reference based on a unique ID (root trace).
+- Any source pulling new content based on a new Digest Checksum.
+- Any subsequent reconciliation that ran.
+- Events emitted and notifications sent by the notification-controller.
+
+On top of this, can be built custom UIs that surface trace timelines alongside Git commit or Docker image tags, so operators can say “what exactly happened when I tagged v1.2.3?” in a single pane of glass. 
+
+### Goals
+- **End-to-end GitOps traceability:** Capture the traces that follows "a Git change" (any source) through all Flux controllers for simply debugging and root-cause analysis.
+- **Declarative, CRD-driven configuration:** Reuse the concept of `Alerts` to be able to populate this feature over, out-of-the-box. Therefore, users can link `EventSources` and `Provider` where trace will be sent.
+- **Notification Controller as the trace-collector:** Leverage the notification-controller's existing event watching pipeline to ingest reconciliation events and turn them into OpenTelemetry spans, being forwarded to an OpenTelemetry-compatible backend - `Provider`.
+- **Cross-controller span correlation:** Ensure spans are emitted from multiple, stateless controller can be stitched together into a single trace by using Flux "revision" annotation.
+
+### Non-Goals
+- **Not a full-tracing backend:** We won't build or bundle a storage/visualization system. Users may have to still rely on a external collector for long-term retention, querying and UI.
+- **Not automatic instrumentation of user workloads:** This integration only captures Flux controller events (Source, Kustomize, Helm, etc.). It won't auto-inject spans into your application pods or third-party controllers running in the same cluster.
+- **Not a replacement for metrics or logs:** Flux's existing Prometheus metrics and structural logging remain the primary way to monitor performance and errors. Tracing is purely for request-flow visibility, not for time-series monitoring or log aggregation.
+- **No deep-code level spans beyond CRUD events:** Will emit spans around high-level reconciliation steps (e.g. "reconcile GitRepository", "dispatch Notification"), but we're not aiming to instrument every internal function call or library method within each controller.
+- **Not a service mesh integration:** It's not plan of the scope tying this into Istio, Linkerd, or other mesh-sidecar approaches. It's strictly a controller-drive, CRD-based model.
+- **No per-span custom enrichment beyond basic metadata:** At least initially, it won't support complex span attributes or tag-enrichment rules. You may have to handle those in your downstream collector/processor if needed.
+- **Not a replacement for user-driven OpenTelemetry SDKs:** If you already have a Go-based operator that embed OpenTelemetry's SDK directly, this feature won't override or duplicate that. Think about it as a complementary, declarative layer for flux controllers.
+
+## Proposal
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.
+
+If the RFC goal is to document best practices,
+then this section can be replaced with the actual documentation.
+-->
+The implementation will extend the notification-controller with OpenTelemetry tracing capabilities by leveraging the existing Alert API object model and adding a new Provider API type called `otel`. This approach maintains Flux's declarative configuration paradigm while adding powerful distributed tracing functionality.
+
+### Core Implementation Strategy
+1. **Extend the notification-controller:** Add OpenTelemetry tracing support to the notification-controller via adding a new type, `otel`. Which already has visibility into events across the Flux ecosystem.
+2. **Leverage existing Alert CRD structure:** Use the Alert Kind API object as the configuration entry point, where:
+     - `EventSources` define which Flux resources to trace (GitRepositories, Kustomizations, HelmReleases, etc.).
+     - `Provider` specifies where to send the trace data (any OpenTelemetry-compatible backends).
+3. **Span generation and correlation:** Generate spans for each reconciliation event from watched resources, ensuring proper parent-child relationships and context propagation using Flux's revision annotations as correlation identifiers.
+4. **Provider compatibility and fallback mechanism:** The implementation supports any provider that implements the OpenTelemetry Protocol (OTLP). When traces are sent to OTLP-compatible providers (like Jaeger or Tempo), they are transmitted as proper OpenTelemetry spans via HTTP(s) requests (no gRPC support at this moment). For non-OTLP providers, the system gracefully degrades by logging trace information as structured warnings in the notification-controller logs, ensuring no alerting functionality is disrupted. This approach maintains system stability while encouraging the use of proper tracing backends.
+
+
+This approach allows users to declaratively configure tracing using familiar Flux patterns, without requiring code changes to their applications or additional sidecar deployments. The notification-controller will handle the collection, correlation, and forwarding of spans to the configured tracing backend.
+
+Example Configuration:
+```yaml
+# Configure the alert
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Alert
+metadata:
+  name: webapp-tracing
+  namespace: default
+spec:
+  providerRef: 
+    name: otel-collector
+  eventSources:
+    - kind: GitRepository  # Source controller resources
+      name: webapp-source
+    - kind: Kustomization  # Kustomize controller resources
+      name: webapp-backend
+    - kind: Kustomization  # Kustomize controller resources
+      name: webapp-frontend
+  eventMetadata:
+    env: staging
+    cluster: cluster-1
+    region: us-east-2
+---
+# Define a tracing provider
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Provider
+metadata:
+  name: otel-collector
+  namespace: default
+spec:
+  type: otel
+    address: http://otel-collector.observability.svc.cluster.local:4318/v1/traces # OTEL Collector endpoint
+  secretRef:
+    name: otel-collector-secret  # Optional: auth + additional headers
+  certSecretRef:
+    name: mtls-certs # Optional: enable mTLS auth
+  proxySecretRef:
+    name: otel-collector-proxy # Optional: proxy configuration
+---
+# OTEL Collector secret
+apiVersion: v1
+kind: Secret
+metadata:
+  name: otel-collector-secret
+  namespace: default
+stringData:
+  # Headers data prevails over auth fields (username/password or token)
+  # Must be used either username/password or token (considers if username is set in order to discriminate bearer token auth or basic auth)
+  username: "<otel-collector-username>"
+  password: "<otel-collector-password>"
+  token: "<otel-collector-api-token>"
+  headers: |
+    X-Forwarded-Proto: https
+---
+# TLS Certificates and keys
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mtls-certs
+  namespace: default
+type: kubernetes.io/tls # or Opaque
+stringData:
+  # All fields are required to enable mTLS
+  tls.crt: |
+    -----BEGIN CERTIFICATE-----
+    <client certificate>
+    -----END CERTIFICATE-----
+  tls.key: |
+    -----BEGIN PRIVATE KEY-----
+    <client private key>
+    -----END PRIVATE KEY-----
+  # Just ca.crt in case of CA-only
+  ca.crt: | 
+    -----BEGIN CERTIFICATE-----
+    <certificate authority certificate>
+    -----END CERTIFICATE-----
+---
+# Proxy configuration
+apiVersion: v1
+kind: Secret
+metadata:
+  name: otel-collector-proxy
+  namespace: default
+stringData:
+  address: "http://<otel-collector-proxy-url>(:<otel-collector-proxy-port>)"
+  username: "<otel-collector-proxy-username>"
+  password: "<otel-collector-proxy-password>"
+```
+
+Based on this configuration, the notification-controller will:
+- Watch for events from the specified resources.
+- Generate OpenTelemetry spans for each reconciliation event.
+- Correlate spans across controllers using Flux's revision annotations.
+- Forward the spans to the configured OTEL Collector endpoint - `Provider`.
+- This implementation maintains Flux's stateless design principles while providing powerful distributed tracing capabilities that help users understand and troubleshoot their GitOps workflows.
+
+### Alternatives
+<!--
+List plausible alternatives to the proposal and explain why the proposal is superior.
+
+This is a good place to incorporate suggestions made during discussion of the RFC.
+-->
+
+## Design Details
+
+### Trace Identity and Correlation
+A key challenge in distributed tracing is establishing a reliable correlation mechanism that works across multiple controllers in a stateless, potentially unreliable environment. Our solution addresses this with a robust span identification strategy.
+
+The Trace ID is generated using a deterministic approach that combines:
+- **Alert Object UID** (guaranteed unique by Kubernetes across all clusters).
+- **Source's revision ID** (extracted from event payloads).
+
+These values are concatenated and passed through a configurable checksum algorithm (SHA-256 by default). This approach ensures:
+- Globally unique trace identifiers across multi-tenant and multi-cluster environments.
+- Consistent trace correlation even when events arrive out of order.
+- Reliable identification of the originating source event.
+
+Example:
+```yaml
+# Input values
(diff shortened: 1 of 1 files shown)


rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,170 @@
...
+  name: webapp
+  namespace: default
+spec:
+  providerRef: 
+    name: jaeger
+  eventSources:
+    - kind: Kustomization
+      name: webapp-backend
+    - kind: HelmRelease
+      name: webapp-frontend
```
stefanprodan, line 128: This does not capture events from source-controller, no GitRepository nor OCIRepository are in the event sources. In this case what will happen as there is no root span for the revision?

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+The user will be able to define a Flux `Trace` custom resource and deploy it to the cluster. The trace may rely on revision annotation as the correlation key due to every Flux reconciliation writes the same `source.toolkit.fluxcd.io/reconcile-revision` annotation onto the object (and is emitted on its Event). That revision string (e.g. a Git commit SHA) is your trace ID surrogate. Therefore, if Source Controller emits an event for `GitRepository` with `revision = SHA123`, later on Kustomize Controller will emit another event for `Kustomization` with the same `revision = SHA123`.
+
+In this way, rather than trying to carry a live span context through controller A -> API server -> controller B:
+1. Watch Flux Events.
+2. Group them by ``event.objectRef.kind`` + ``revision``. Hence, based on `EventSources` field:
+     - First time you see `GitRepository@SHA123`, spawn a root span with traceID = hash(SHA123).
+     - When you see `Kustomization@SHA123`, spawn a child span (parent = root) in that same trace.
+     - And so on for `HelmRelease@SHA123`, `Notification@SHA123`, etc.
```
stefanprodan, line 149: HelmReleases events don't have the Git SHA in their events, the source of a HelmRelease is a HelmChart or an OCIRepository, the revision is the version of the chart. It's not possible for helm-controller events to spawn child spans for Git.

rfcs/0011-opentelemetry-tracing/README.md
adri1197: Good point! We can assume that the root span may come from any of the available sources. It's true that it could be so time consuming iterating over all available sources (specially on multi-tenant setups and large deployments) and tracing back. Therefore, could be set as mandatory to have a source as part of the eventSources list. Every "deployable" entity must have a source.

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,170 @@
...
+    name: jaeger-secret
+```
+
+The user will be able to define a Flux `Trace` custom resource and deploy it to the cluster. The trace may rely on revision annotation as the correlation key due to every Flux reconciliation writes the same `source.toolkit.fluxcd.io/reconcile-revision` annotation onto the object (and is emitted on its Event). That revision string (e.g. a Git commit SHA) is your trace ID surrogate. Therefore, if Source Controller emits an event for `GitRepository` with `revision = SHA123`, later on Kustomize Controller will emit another event for `Kustomization` with the same `revision = SHA123`.
+
+In this way, rather than trying to carry a live span context through controller A -> API server -> controller B:
+1. Watch Flux Events.
+2. Group them by ``event.objectRef.kind`` + ``revision``. Hence, based on `EventSources` field:
+     - First time you see `GitRepository@SHA123`, spawn a root span with traceID = hash(SHA123).
+     - When you see `Kustomization@SHA123`, spawn a child span (parent = root) in that same trace.
```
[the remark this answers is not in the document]
adri1197: In that case, I'd suggest that the notification-controller should be able to assume that there must be a source as a root span in every trace and build synthetic spans. In the end of day, any deployment entity must have a source (could be Git, HelmRepo and OCIRepo). Therefore, once receives from kustomize-controller:
- List Trace's related GitRepository/OCIRepository/HelmRepository objects listed under `EventSources` and rely on `status.artifact.revision`.
- Enqueue a synthetic “root” span for each current revision before processing any buffered Events.

In this way, orphaned as well as "incomplete" spans can be avoided.

rfcs/0011-opentelemetry-tracing/README.md
adri1197: Well.. there could be also the case that there is no Git involved. The config as well as the deployment could come from non-Git sources (not ideal, though). Not sure if it would be good to include as part `ArtifactUpToDate` event the artifact digest SHA to have the same format across (in addition to the version), regardless of the source.

adri1197: Of course, I completely agree with you. Must support all potential sources. One thing that I noticed while this discussion is that it doesn't seem to be a common approach in terms of events and the info populated across sources. Meaning that the ocirepo populates the artifact version, gitrepo commit sha and so on. Perhaps, we should start by standardizing the info populated within the events related to the source. We came up with this table about the data could be potentially populated:

| Source Type    | Recommended Unique Identifier                                             | Additional Useful Information in Trace               |
|----------------|-----------------------------------------------------------------------------|------------------------------------------------------|
| GitRepository  | Git Commit SHA (commit sha)                                                | Repo URL, Branch, Tag, Author, Commit message        |
| OCIRepository  | OCI Manifest Digest (`sha256:<digest>`)                                     | Registry URL, Repository path, Tag                  |
| Bucket (S3)    | Object version ID (if versioning enabled) or Object ETag                   | Bucket name, Object path, Region                     |
| HelmRepository | Index.yaml digest or a combination of URL + timestamp                       | Helm repo URL, Chart name, Chart version             |
| HelmChart      | Helm Chart digest (`sha256:<digest>`) if pulled via OCI; otherwise, Chart version and URL | Chart name, Helm repo source, Chart version |

Which are your thoughts on this?

stefanprodan: The revision format cannot be changed as it would break the whole Flux ecosystem, this was standardized in Flux 2.0 GA in RFC-0005. The format is documented for each source type. Docs here: https://fluxcd.io/flux/components/source/

adri1197: As far as I see in RFC-0005, the current format may fit our use case. As a root span, we could have the `artifact:revision` which should be available in most of the main sources: HelmRepository, OCIRepository, GitRepository and Bucket . Then, HelmChart may have a `sourceRef` as a top source, therefore there would be always a unique ID (root span) that, in this case,  may cascade over with the chart version following the event (as a child span).

As an aside comment, I've seen that `artifact:revision` doesn't get properly populated if you have an OCI type for a HelmRepository. Is it intended?

stefanprodan: HelmRepository is a pointer to an index.yaml, but for OCI there is no such things. In any case, the actual source of a HelmRelease is always a Helm chart not the repo, and the revision of the HelmRelease is in the format `<chart-version>-<oci shot digest>`.

adri1197: My point was that HelmChart must always have a top source. Therefore, the sequence may be: <GitRepository/HelmRepository/Bucket/OCIRepository> > HelmChart > HelmRelease, just as an example.  Thereby, by having this `<oci short digest>` (or any other checksum if other source is set), we could correlate over from source to final HelmRelease. Following the previous example: `<branch>@<algo>:<git-hash>`/`<algo>@<index.yaml pointer (checksum)>`/`<algo>:<checksum>`/`<oci repo name>@<algo>:<oci short digest>`/ > `chart-version` (checking source) > `<chart-version>-<source digest checksum>`.

In this way, by having an event on HelmRelease following kind of this structure - `<chart version>:<source digest checksum>` on any kind of source. Following the same approach as OCI, but in a generic way for the remaining sources. We should be capable of tracing the whole sequence.

Oh.. I see, we're currently using a HelmRepo pointing to an OCI type registry, just to have version control under the HelmRelease. We'll investigate further to move towards the recommended approach.

matheuscscp: When you use an OCIRepository for a HelmRelease there's no HelmChart involved, the HelmRelease references the OCIRepository object directly

adri1197: Well.. in that scenario, the root span may be the `<oci repo name>@<algo>:<oci short digest>` (OCI revision), cascading over with child spans about the HelmRelease, which may contain the corresponding chart name, version and `artifact digest`.

In the end of the day, we need a unique ID which does have a correlation with the source and its reconciliation reference to work as the root span. Then, we can have multiple objects which uses that source as child spans. Thereby, we need to have that ID across event's "source consumers" in order to be able to traceback to the source and full traceability. May it be possible?

stefanprodan: How would notification-controller know the repo name from `<oci repo name>@<algo>:<oci short digest>`, the event issued by a HelmRelease contains the revision which is `<oci-tag>@<algo>:<oci short digest>`.

adri1197: Oh... I see. Then, the only solution I can think of is to additional check either HelmRelease spec to get that information (repo name) or filter by these two values `<oci-tag>`  and/or `<oci short digest>` (which must be unique) to get the proper source and therefore, get the repo name. As far as I see, there is no other way to get it "out-of-the-box" by the info populated over the event issued.

adri1197:
RFC updated with the approach discussed on Flux Dev Meeting 2025-05-07.
@stefanprodan @matheuscscp

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,163 @@
...
+Example Configuration:
+```yaml
+# Define a tracing provider
+apiVersion: notification.toolkit.fluxcd.io/v1beta1
+kind: Provider
+metadata:
+  name: jaeger
+  namespace: default
+spec:
+  type: jaeger
```
stefanprodan:
```suggestion
  type: otel
```

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,163 @@
...
+4. **Provider compatibility and fallback mechanism:** The implementation supports any provider that implements the OpenTelemetry Protocol (OTLP). When traces are sent to OTLP-compatible providers (like Jaeger or Tempo), they are transmitted as proper OpenTelemetry spans. For non-OTLP providers, the system gracefully degrades by logging trace information as structured warnings in the notification-controller logs, ensuring no alerting functionality is disrupted. This approach maintains system stability while encouraging the use of proper tracing backends.
+
+This approach allows users to declaratively configure tracing using familiar Flux patterns, without requiring code changes to their applications or additional sidecar deployments. The notification-controller will handle the collection, correlation, and forwarding of spans to the configured tracing backend.
+
+Example Configuration:
+```yaml
+# Define a tracing provider
+apiVersion: notification.toolkit.fluxcd.io/v1beta1
```
stefanprodan:
```suggestion
apiVersion: notification.toolkit.fluxcd.io/v1beta3
```

stefanprodan (commented):
I think we should introduce a Provider type called `otel` that should work with any OpenTelemetry-compatible backends.

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,162 @@
...
+  providerRef: 
+    name: otel
+  eventSources:
+    - kind: GitRepository  # Source controller resources
+      name: webapp-source
+    - kind: Kustomization  # Kustomize controller resources
+      name: webapp-backend
+    - kind: Kustomization  # Kustomize controller resources
+      name: webapp-frontend
+```
```
matheuscscp, line 153: Let's add here also a YAML for the Secret using the `.stringData` write-only field, please :pray: 

Let's show in this YAML all the supported fields inside the Secret

adri1197: Sure! Secrets added: auth, proxy, mTLS and headers.

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,224 @@
...
+  password: "proxy_password"
+---
+# Additional headers to appended into HTTP request
+apiVersion: v1
+kind: Secret
+metadata:
+  name: provider-headers
+  namespace: default
+stringData:
+  headers: |
```
matheuscscp: Looks like this key should be in the `jaeger-secret` secret? Let's show here a single secret per reference field in the API, and let's also add comments close to the keys showing which ones are required and which ones are optional. Also, the `jaeger-secret` secret with both username+password and token without any comments it's a bit strange right now, should add some comment saying that it's one or the other (if that's really the case?)

adri1197: Yes, you're absolutely right, added as part of `jaeger-secret`. I made a mistake by considering that username/password or token secrets gets populated as headers. I'm going to add some logic under notification-controller PR (https://github.com/fluxcd/notification-controller/issues/1097)

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,218 @@
...
+- Alert reference based on a unique ID (root span).
+- Any source pulling new content based on a new Digest Checksum.
+- Any subsequent rencolitiation that ran.
+- Events emitted and notifications sent by the notification-controller.
+
+On top of this, can be built custom UIs that surface trace timelines alongside Git commit or Docker image tags, so operators can say “what exactly happened when I tagged v1.2.3?” in a single pane of glass. 
+
+### Goals
+- **End-to-end GitOps traceability:** Capture the traces that follows "a Git change" (any source) through all FluxCD controllers for simply debugging and root-cause analysis.
+- **Declarative, CRD-drive configuration:** Reuse the concept of `Alerts` to be able to populate this feature over, out-of-the-box. Therefore, users can link `EventSources` and `Provider` where trace will be sent.
```
matheuscscp:
```suggestion
- **Declarative, CRD-driven configuration:** Reuse the concept of `Alerts` to be able to populate this feature over, out-of-the-box. Therefore, users can link `EventSources` and `Provider` where trace will be sent.
```

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,218 @@
...
+- **Notification Controller as the trace-collector:** Leverage the notification-controller's existing event watching pipeline to ingest reconciliation events and turn me into OpenTelemetry spans, being forwarwed to an OLTP-compatible backend - `Provider`.
```
matheuscscp:
```suggestion
- **Notification Controller as the trace-collector:** Leverage the notification-controller's existing event watching pipeline to ingest reconciliation events and turn them into OpenTelemetry spans, being forwarwed to an OLTP-compatible backend - `Provider`.
```

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,218 @@
...
+- **Cross-controller span correlation:** Ensure spans are emitted from multiple, stateless controller can be stiched together into a single trace by using Flux "revision" annotation.
+
+### Non-Goals
+- **Not a full-tracing backend:** We won't build or bundle a storage/visualization system. Users may have to still rely on a external collector for long-term retention, querying and UI.
+- **Not automatic instrumentation of user workloads:** This integration only captures FluxCd controller events (Source, Kustomize, Helm, etc.). It won't auto-inject spans into your application pods or third-party controllers running in the same cluster.
```
matheuscscp:
```suggestion
- **Not automatic instrumentation of user workloads:** This integration only captures Flux controller events (Source, Kustomize, Helm, etc.). It won't auto-inject spans into your application pods or third-party controllers running in the same cluster.
```

matheuscscp: Let's use Flux everywhere instead of FluxCD

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,218 @@
...
+- **Not a replacement for metrics or logs:** Flux's existing Prometheus metrics and structural logging remain the primary way to monitor performance and errors. Tracing is purely for request-flow visibility, not for time-series monitoring or log aggregation.
+- **No deep-code lelve spans beyond CRUD events:** Will emit spans around high-level reconciliation steps (e.g. "reconcile GitRepository", "dispatch Notification"), but we're not aiming to instrument every internal function call or library method within each controller.
```
matheuscscp:
```suggestion
- **No deep-code level spans beyond CRUD events:** Will emit spans around high-level reconciliation steps (e.g. "reconcile GitRepository", "dispatch Notification"), but we're not aiming to instrument every internal function call or library method within each controller.
```

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,218 @@
...
+- **Not a service mesh integration:** It's not plan of the scope tieing this into Istio, Linkerd, or other mesh-sidecar approaches. It's strictly a controller-drive, CRD-based model.
```
matheuscscp:
```suggestion
- **Not a service mesh integration:** It's not plan of the scope tying this into Istio, Linkerd, or other mesh-sidecar approaches. It's strictly a controller-drive, CRD-based model.
```

matheuscscp (commented):
I think it would be good to use a spell-checker in the document

adri1197:
@matheuscscp You're absolutely right. I ran a spell-checker and many typos cropped up. Should look a bit better now 😄

matheuscscp (approved):
LGTM

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,217 @@
...
+- **Asynchronous Event Processing:** Since events may arrive in any order due to the distributed nature of Flux controllers, the system doesn't assume sequential processing. Each event can independently locate its parent span or create a new root span as needed.
+- **Fault Tolerance:** If the notification-controller experiences downtime or latency issues, it implements a recovery mechanism:
+  - When processing an event, it first attempts to locate an existing root span based on the calculated ID.
+  - If found, it attaches the new span as a child to maintain the trace hierarchy.
+  - If not found (due to previous failures or out-of-order processing), it automatically creates a new root span
+- Span Hierarchy Maintenance: All subsequent spans related to the same revision are properly attached to their parent spans, creating a coherent trace visualization regardless of when events are processed.
+
+This design ensures trace continuity even in challenging distributed environments while maintaining Flux's core principles of statelessness and resilience.
+
```
yuriolisa, line 210: I believe we should include here a sub-topic with a recommendation about OTEL [Trace semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/trace/).

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,217 @@
...
+2. **Leverage existing Alert CRD structure:** Use the Alert Kind API object as the configuration entry point, where:
+     - `EventSources` define which Flux resources to trace (GitRepositories, Kustomizations, HelmReleases, etc.).
+     - `Provider` specifies where to send the trace data (any OpenTelemetry-compatible backends).
+3. **Span generation and correlation:** Generate spans for each reconciliation event from watched resources, ensuring proper parent-child relationships and context propagation using Flux's revision annotations as correlation identifiers.
+4. **Provider compatibility and fallback mechanism:** The implementation supports any provider that implements the OpenTelemetry Protocol (OTLP). When traces are sent to OTLP-compatible providers (like Jaeger or Tempo), they are transmitted as proper OpenTelemetry spans. For non-OTLP providers, the system gracefully degrades by logging trace information as structured warnings in the notification-controller logs, ensuring no alerting functionality is disrupted. This approach maintains system stability while encouraging the use of proper tracing backends.
+
```
yuriolisa, line 63: I'm assuming we're gonna use the HTTP channel to submit the Traces to the OTEL Collector, but it would be great to have this definition here. The reason for this is that many users configure the gRPC by default.

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,217 @@
...
+---
+# Define a tracing provider
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Provider
+metadata:
+  name: jaeger
+  namespace: default
+spec:
+  type: otel
+  address: http://jaeger-collector.jaeger-system.svc.cluster.local:9411 # Provider endpoint
```
yuriolisa: I'd recommend using a generic OTEL Collector configuration instead of a Jaeger one, once the Collector receives the data points and so delivers to a Jaeger, Tempo, etc.

rfcs/0011-opentelemetry-tracing/README.md
```diff
@@ -0,0 +1,217 @@
...
+  address: "http://proxy_url:proxy_port"
+  username: "proxy_username"
+  password: "proxy_password"
+```
+
+Based on this configuration, the notification-controller will:
+- Watch for events from the specified resources.
+- Generate OpenTelemetry spans for each reconciliation event.
+- Correlate spans across controllers using Flux's revision annotations.
+- Forward the spans to the configured Jaeger endpoint - `Provider`.
```
yuriolisa: If we decide to move forward to above OTEL generic configuration, we have to adjust here.

rfcs/0011-opentelemetry-tracing/README.md
adri1197: Good point! Let's clarify it just from the beginning. In principle, as you correctly pointed out, it's going to only support HTTP

rfcs/0011-opentelemetry-tracing/README.md
adri1197: I've been thinking about this... I'm not particularly sure how this semantic conventions may fit in this use case as well as the resource/attributes conventions. Could be potentially added as a note, but not stated as supported, as of now.

stefanprodan:
@adri1197 please rebase with upstream main

rfcs/0011-opentelemetry-tracing/README.md
yuriolisa: The semantic conventions should be used in the trace/span signal to have a good instrumentation of OTEL in place. I don't believe we should document everything here, but mention that all the FluxCD traces should obey this.

adri1197:
> @adri1197 please rebase with upstream main

Done! 😃

rfcs/0011-opentelemetry-tracing/README.md
stefanprodan: The closest semantic convention for what Flux does would be CloudEvents / CDEvents, but it would only fit the Helm install/upgrade/uninstall events. Given this, I'm not for imposing a convention now, we should probably have our own conventions specific to the reconciliation process.

stefanprodan (approved):
LGTM

Thanks @adri1197 🏅

PR #5332, opened 2025-05-03
[RFC-0010] Add RBAC for creating service account tokens

matheuscscp:
Part of: #5022

All the six controllers will eventually implement object-level workload identity so this should be fine.


stefanprodan (approved):
LGTM

PR #5333, opened 2025-05-04
Upgrade fluxcd/pkg auth, oci, git and git/gogit

matheuscscp:
Depends on: https://github.com/fluxcd/pkg/pull/919

Part of: https://github.com/fluxcd/flux2/issues/5022


stefanprodan (approved):
LGTM

PR #5338, opened 2025-05-06
Fix exit code handling in get command

### LINKED ISSUE #5330: `flux get` exits with exit code 0 even if the resource does not exist
### Describe the bug

I'm using a script to bootstrap Flux from Jenkins. Due to organization processes and branch protections, our script effectively replicates the normal `flux bootstrap` process manually with `flux install` and does some secret creations, and whatnot to get Flux up and running.

I've noticed that when I do `flux get kustomization flux-system -n flux-system` before bootstrapping, the Flux CLI still returns an exit code of 0 instead of an exit code indicating an error.

```
❯ flux get kustomization flux-system -n flux-system --timeout 1m0s
✗ Kustomization object 'flux-system' not found in "flux-system" namespace
❯ echo $?
0
❯
```

As a last resort, I've fallen back to using `kubectl get kustomization` instead, but that has its own issues because the non-zero exit code 1 from kubectl can be either due to missing CRDs such as the Kustomization CRD, or because of the Flux kustomization being actually missing. Whereas, with the Flux CLI, I would expect these two scenarios to result in different error codes.

Please make the Flux CLI return proper exit error codes depending on the response from the API regarding whether the `kind: Kustomization` simply doesn't exist, or whether the Kustomization CRD has not been installed.

Currently if the Kustomization CRD (or any other) has not been installed, the CLI does properly return a non-zero exit code, however the code is 1, and it should probably be something other than that when/if this is implemented.

### Steps to reproduce

See above

### Expected behavior

Flux CLI returns 1 for resources that aren't found, and a different exit code for supported resource Kinds that aren't found.

_No response_

### OS / Distro

N/A

### Flux version

v2.3.0

### Flux check

N/A

### Git provider

_No response_

_No response_

### Additional context

_No response_

### Code of Conduct

dgunzy:
## What this PR does

Fixes #5330 - Changes the error handling in the `get` command to properly return non-zero exit codes when resources are not found, instead of just logging errors.

## How I tested it

I tested both the original and modified behaviour with the following test script (I compiled before to compare to the version in .bin/flux):

```
#!/bin/bash

echo "Test 1: Specific object not found - original flux command (should return 0 exit code)"
flux get kustomization non-existent-resource
echo "Exit code: $?"

echo -e "\nTest 2: No objects found in namespace - original flux command (should return 0 exit code)"
flux get helmrelease -n kube-system
echo "Exit code: $?"

echo -e "\nTest 3: Specific object not found - modified ./bin/flux command (should return 1 exit code)"
./bin/flux get kustomization non-existent-resource
echo "Exit code: $?"

echo -e "\nTest 4: No objects found in namespace - modified ./bin/flux command (should return 1 exit code)"
./bin/flux get helmrelease -n kube-system 
echo "Exit code: $?"
```

## Output

```
Test 1: Specific object not found - original flux command (should return 0 exit code)
✗ Kustomization object 'non-existent-resource' not found in "flux-system" namespace
Exit code: 0

Test 2: No objects found in namespace - original flux command (should return 0 exit code)
✗ no HelmRelease objects found in "kube-system" namespace
Exit code: 0

Test 3: Specific object not found - modified ./bin/flux command (should return 1 exit code)
✗ Kustomization object 'non-existent-resource' not found in "flux-system" namespace
Exit code: 1

Test 4: No objects found in namespace - modified ./bin/flux command (should return 1 exit code)
✗ no HelmRelease objects found in "kube-system" namespace
Exit code: 1
```

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -184,13 +184,13 @@ func (get getCommand) run(cmd *cobra.Command, args []string) error {
 
 	if get.list.len() == 0 {
 		if len(args) > 0 {
-			logger.Failuref("%s object '%s' not found in %s namespace",
+			return fmt.Errorf("%s object '%s' not found in %s namespace",
 				get.kind,
 				args[0],
 				namespaceNameOrAny(getArgs.allNamespaces, *kubeconfigArgs.Namespace),
 			)
 		} else if !getAll {
-			logger.Failuref("no %s objects found in %s namespace",
+			return fmt.Errorf("no %s objects found in %s namespace",
 				get.kind,
 				namespaceNameOrAny(getArgs.allNamespaces, *kubeconfigArgs.Namespace),
 			)
--- a/cmd/flux/get_test.go
+++ b/cmd/flux/get_test.go
@@ -19,7 +19,10 @@ limitations under the License.
 
 package main
 
-import "testing"
+import (
+	"fmt"
+	"testing"
+)
 
 func Test_GetCmd(t *testing.T) {
 	tmpl := map[string]string{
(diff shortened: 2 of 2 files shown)


dgunzy:
> @dgunzy can you please add tests for this in the `get_test.go`, you can use the `assertError` and `assertSuccess` helper functions.

Yes! I will update the PR today with some tests for this.

dgunzy:
@stefanprodan  I added some tests for this. Let me know if you would like any more changes.

stefanprodan:
@dgunzy can you please squash all commits into one and force push

dgunzy:
done!

stefanprodan (approved):
LGTM

Thanks @dgunzy

PR #5407, opened 2025-06-13
refactor: Use `normalize.UnstructuredList` instead of `ssa.SetNativeKindsDefaults`

cappyzawa:
ref: https://github.com/fluxcd/pkg/blob/3d6f759b7671a1ab0b3da5ee856de681d6f7532f/ssa/utils.go#L26-L30


matheuscscp (approved):
LGTM

PR #5440, opened 2025-07-04
manifests: Add `app.kubernetes.io/part-of: flux` label to controller pods

pinkavaj:
Ensure also pods contain the relevant labels inherited from parent Deployment object, this makes it easier to select and filter the pods using the labels eg. when scraping for metrics.

### DIFF
--- a/manifests/bases/helm-controller/labels.yaml
+++ b/manifests/bases/helm-controller/labels.yaml
@@ -8,3 +8,6 @@ labels:
 fieldSpecs:
   - path: metadata/labels
     create: true
+  - kind: Deployment
+    path: spec/template/metadata/labels
+    create: true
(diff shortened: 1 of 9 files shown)


stefanprodan:
See the scraping config here, each pod is labeled with the controller name: https://github.com/fluxcd/flux2-monitoring-example/blob/main/monitoring/configs/podmonitor.yaml

pinkavaj:
> See the scraping config here, each pod is labeled with the controller name: https://github.com/fluxcd/flux2-monitoring-example/blob/main/monitoring/configs/podmonitor.yaml

Yep, each pod is labeled with `app: <CONTROLLER_NAME>`, and the other labels from Deployment are not present. Please can you be more elaborate on how this related to this MR?

Btw one of the goals of this MR is to propagate the `app.kubernetes.io/*`, those labels are suggested as best practice by K8s and having them will make our live easier, because most of the common apps already use them and thus we can use the same tooling for all the apps ..

stefanprodan (approved):
LGTM

PS. Please rebase with upstream main and force push

Issue #5447, opened 2025-07-09
Cluster Authentication via Workload Identity for Generic Provider


Preisschild:
Reference: https://github.com/fluxcd/pkg/pull/960, https://github.com/fluxcd/flux2/blob/main/rfcs/0010-multi-tenant-workload-identity/README.md#api-changes-and-feature-gates, https://github.com/fluxcd/kustomize-controller/pull/1476
Slack thread: https://cloud-native.slack.com/archives/C01A402P0R2/p1752066566113149

# Description

In the currently proposed RFC-0010 implementation, [it is not supported to authenticate against the target cluster using workload identity if you are using a `generic` kubernetes cluster](https://github.com/fluxcd/kustomize-controller/pull/1476/files#diff-f3a1c537d2c89e0a43fdf16b23b9612d96ff9b2997d0f97d55acf4331057d4ccR275)

It would be useful to cover this use case for `generic` clusters too. This could be possible by creating a serviceaccount token with the audience set to the target cluster's address and then using this token to authenticate against the target cluster. The target cluster needs to have the source cluster's [service account issuer discovery endpoints](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery) configured as an [additional kube-apiserver jwt issuer in the target cluster](https://kubernetes.io/docs/reference/config-api/apiserver-config.v1alpha1/#apiserver-k8s-io-v1alpha1-Issuer).

Example kube-apiserver AuthenticationConfiguration:

```yaml
---
apiVersion: apiserver.config.k8s.io/v1beta1
kind: AuthenticationConfiguration
jwt:
- issuer:
    url: https://<hub cluster domain>:6443
    audiences:
    - https://<target cluster domain>:6443
    certificateAuthority: |-
      <hub cluster apiserver CA>
  claimMappings:
    username:
    claim: "sub"
    prefix: "hub-cluster:"
```

If Flux controllers then create a token for the serviceaccount referenced in `.spec.kubeConfig.serviceAccountName` of the kustomization / helmrelease, with the audience set to `.spec.kubeConfig.address` using the `TokenRequest` API in the source cluster it can then authenticate against the target cluster using this serviceaccount.

For example, if you'd create a Kustomization like this:

```yaml
---
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: sync
  namespace: tenant-a
spec:
  kubeConfig:
[...]

PR #5452, opened 2025-07-10
[RFC-0010] Add workload identity support for remote generic clusters

matheuscscp:
Part of: #5022

xref: #5447
xref: #4548


stefanprodan (approved):
LGTM

Issue #5458, opened 2025-07-18
CI Bug: `update` workflow fails a lot


matheuscscp:
At some point I momentarily disabled and re-enabled the `update` workflow, so now I'm the one receiving emails on failures.

This workflow fails a lot, this morning I saw 6 failure emails in my mailbox. 

<img width="704" height="443" alt="Image" src="https://github.com/user-attachments/assets/378ecff4-cf33-44b3-9bee-60739b0832ef" />

Here's an example:

https://github.com/fluxcd/flux2/actions/runs/16357001233/job/46217603206

```
jq: error (at <stdin>:1): Cannot index string with string "published_at"
go: -require=github.com/fluxcd/helm-controller/api@: invalid version ""
```

The only place where `jq` is used in the workflow is here:

https://github.com/fluxcd/flux2/actions/runs/16357001233/workflow#L35

```bash
          bump_version() {
            local LATEST_VERSION=$(curl -s https://api.github.com/repos/fluxcd/$1/releases | jq -r 'sort_by(.published_at) | .[-1] | .tag_name')
```

Looking at the last 7 failed workflows, it seems like the failure is always for `bump_version` being called for `helm-controller`.

My first guess is that `curl -s https://api.github.com/repos/fluxcd/$1/releases` is getting rate-limited by GitHub, but it's super weird that it only happens for helm-controller??

In any case, this seems to be the only call we are making to GitHub in this workflow. Would it be worth doing something to mitigate these failures? (Thinking retries with exp. backoff)

stefanprodan:
GitHub has put a tight leash on their APIs recently. We need to pass the token to that curl or switch to `gh`.

matheuscscp:
Oh, and the mistery of helm-controller it's because it's the first :) 🕵️

matheuscscp:
This is back, I assume this time because of GitHub's poor availability recently? But the error is in `go mod` commands, so not sure... Maybe the Go package server tries to reach GitHub every time for some reason? Probably nothing we can do here 😢 

https://github.com/fluxcd/flux2/actions/runs/16932748390/job/47982038243

PR #5460, opened 2025-07-18
ci: Use GITHUB_TOKEN for API calls in update workflow

stefanprodan:
Fix: #5458


matheuscscp:
This definitely seem to have fixed the issue, thanks very much!

PR #5473, opened 2025-08-08
Implement `flux migrate` command

stefanprodan:
The migrate command must be run before a Flux minor version upgrade. The command migrates the Flux custom resources stored in Kubernetes etcd to their latest API version, ensuring the Flux components can continue to function correctly after the upgrade.

### DIFF
--- a/cmd/flux/migrate.go
+++ b/cmd/flux/migrate.go
@@ -1,0 +1,184 @@
+/*
+Copyright 2025 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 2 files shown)


cmd/flux/migrate.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (m *Migrator) Run(ctx context.Context) error {
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+
+	if err := m.kubeClient.List(ctx, crdList, m.labelSelector); err != nil {
+		return fmt.Errorf("failed to list CRDs: %w", err)
+	}
+
+	for _, crd := range crdList.Items {
+		if err := m.migrateCRD(ctx, crd.Name); err != nil {
```
matheuscscp, line 96: Why not just pass `crd` to the function? `.List()` will already return the CRD object, no need to `.Get()`

stefanprodan: Went with this as it's safer, to process a single CRD it can take minutes and by that time the storage version may change.

cmd/flux/migrate.go
```diff
@@ -0,0 +1,184 @@
...
+func (m *Migrator) migrateCR(ctx context.Context, crd *apiextensionsv1.CustomResourceDefinition, version string) error {
+	list := &unstructured.UnstructuredList{}
+
+	apiVersion := crd.Spec.Group + "/" + version
+	listKind := crd.Spec.Names.ListKind
+
+	list.SetAPIVersion(apiVersion)
+	list.SetKind(listKind)
+
+	err := m.kubeClient.List(ctx, list, client.InNamespace(""))
```
matheuscscp, line 146: It's curious that `.List()` returns objects in previous versions when the specified `version` is just the latest 🤔 

🤷

stefanprodan: CRs are not versioned in etcd, thr API will allways serve the latest API version

matheuscscp (commented):
GitHub is forcing me to write this comment to submit a review lol

matheuscscp (approved):
LGTM! 🚀

Issue #5411, opened 2025-06-16
Minor improvements to image automation APIs before GA


matheuscscp:
Just to track a few things we need to do before making the image APIs GA:

- [ ] Add better kubectl columns with status and message to the CRDs
- [ ] Add some aliases like `k get imagerepo` and `k get imagepol`
- [ ] Remove deprecated code paths (e.g. auto-login flags in IRC)

lukas8219:
I can have this one >Add better kubectl columns with status and message to the CRDs

cappyzawa:
I'd like to work on the second task: "Add some aliases like k get imagerepo and k get imagepol"

cappyzawa:
Hi! I'd like to work on the third task: **"Remove deprecated code paths"** in the image-reflector-controller.

Note: I'm also working on the second task (shortNames addition) in the same issue.

## Implementation Plan

### 1. Auto-login Flags ([deprecated in v0.25.0](https://github.com/fluxcd/image-reflector-controller/blob/b65e8a0/CHANGELOG.md#0250), [shows error logs](https://github.com/fluxcd/image-reflector-controller/blob/b65e8a0/main.go#L134-L138))
**Flags to remove:**
- `--aws-autologin-for-ecr`
- `--gcp-autologin-for-gcr` 
- `--azure-autologin-for-acr`

**Migration:** Use `.spec.provider` field in ImageRepository objects

I'll start with this since it's mentioned as an example.

### 2. Deprecated TLS Key Names ([deprecated in v1beta2](https://github.com/fluxcd/image-reflector-controller/blob/b65e8a0/api/v1beta2/imagerepository_types.go#L84-L85), [shows warning logs](https://github.com/fluxcd/image-reflector-controller/blob/b65e8a0/internal/registry/options.go#L163-L165))
**Keys to remove:**
- `caFile` → use `ca.crt`
- `certFile` → use `tls.crt`
- `keyFile` → use `tls.key`

### 3. ImageRepositoryFinalizer Constant ([marked deprecated](https://github.com/fluxcd/image-reflector-controller/blob/b65e8a0/api/v1beta2/imagerepository_types.go#L30-L31))
Remove `ImageRepositoryFinalizer` constant (alias to `ImageFinalizer`), unused in codebase.

I'll proceed with items 2 and 3 after confirmation that this approach works for preparing the APIs for GA.

Would this plan work?

matheuscscp:
[...]

PR #5492, opened 2025-08-26
Implement `flux [reconcile|suspend|resume] image policy` commands

lukas8219:
Closes: #5411

### DIFF
--- a/cmd/flux/image.go
+++ b/cmd/flux/image.go
@@ -17,6 +17,8 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	autov1 "github.com/fluxcd/image-automation-controller/api/v1"
@@ -77,6 +79,34 @@ func (a imagePolicyAdapter) asClientObject() client.Object {
 	return a.ImagePolicy
 }
 
+func (a imagePolicyAdapter) deepCopyClientObject() client.Object {
+	return a.ImagePolicy.DeepCopy()
+}
+
+func (a imagePolicyAdapter) isStatic() bool {
+	return false
+}
+
+func (a imagePolicyAdapter) lastHandledReconcileRequest() string {
+	return a.Status.GetLastHandledReconcileRequest()
+}
+
+func (a imagePolicyAdapter) isSuspended() bool {
+	return a.Spec.Suspend
+}
+
+func (a imagePolicyAdapter) setSuspended() {
+	a.Spec.Suspend = true
+}
+
+func (a imagePolicyAdapter) successMessage() string {
+	return fmt.Sprintf("selected ref %s", a.Status.LatestRef.String())
+}
+
+func (a imagePolicyAdapter) setUnsuspended() {
+	a.Spec.Suspend = false
+}
+
 // imagev1.ImagePolicyList
 
 type imagePolicyListAdapter struct {
@@ -89,6 +119,18 @@ func (a imagePolicyListAdapter) asClientList() client.ObjectList {
 
 func (a imagePolicyListAdapter) len() int {
 	return len(a.ImagePolicyList.Items)
+}
+
+func (a imagePolicyListAdapter) resumeItem(i int) resumable {
+	return &imagePolicyAdapter{&a.ImagePolicyList.Items[i]}
+}
+
+func (obj imagePolicyAdapter) getObservedGeneration() int64 {
+	return obj.ImagePolicy.Status.ObservedGeneration
+}
+
+func (a imagePolicyListAdapter) item(i int) suspendable {
+	return &imagePolicyAdapter{&a.ImagePolicyList.Items[i]}
 }
 
 // autov1.ImageUpdateAutomation
--- a/cmd/flux/image_test.go
+++ b/cmd/flux/image_test.go
@@ -53,6 +53,18 @@ func TestImageScanning(t *testing.T) {
 			"get image policy podinfo-regex",
 			"testdata/image/get_image_policy_regex.golden",
 		},
+		{
+			"suspend image policy podinfo-semver",
+			"testdata/image/suspend_image_policy.golden",
+		},
+		{
+			"resume image policy podinfo-semver",
+			"testdata/image/resume_image_policy.golden",
+		},
+		{
+			"reconcile image policy podinfo-semver",
+			"testdata/image/reconcile_image_policy.golden",
+		},
 	}
 
 	for _, tc := range cases {
--- a/cmd/flux/reconcile_image_policy.go
+++ b/cmd/flux/reconcile_image_policy.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2025 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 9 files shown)


cmd/flux/image.go
```diff
@@ -77,6 +79,34 @@ func (a imagePolicyAdapter) asClientObject() client.Object {
...
+func (a imagePolicyAdapter) isSuspended() bool {
+	return a.Spec.Suspend
+}
+
+func (a imagePolicyAdapter) setSuspended() {
+	a.Spec.Suspend = true
+}
+
+func (a imagePolicyAdapter) successMessage() string {
+	return fmt.Sprintf("scan fetched %d tags", a.Status.LatestRef.Tag)
```
cappyzawa: Tag is a string type.

Since ImagePolicy selects a ref from ImageRepository scan results based on the policy:

```suggestion
	return fmt.Sprintf("selected ref %s", a.Status.LatestRef.String())
```

This aligns with the pattern used by other source commands (e.g., `fetched revision %s`) and accurately reflects what ImagePolicy does.

cmd/flux/reconcile_image.go
```diff
@@ -26,6 +27,19 @@ var reconcileImageCmd = &cobra.Command{
...
+	Use:   "policy [name]",
+	Short: "Reconcile an ImagePolicy",
+	Long:  `The reconcile image policy command triggers a reconciliation of an ImagePolicy resource and waits for it to finish.`,
+	Example: `  # Trigger an scan for an existing image policy
+  flux reconcile image policy alpine`,
+	RunE: reconcileCommand{
+		apiType: imagePolicyType,
+		object:  imagePolicyAdapter{&imagev1.ImagePolicy{}},
+	}.run,
+}
```
cappyzawa: The new commands are missing `ValidArgsFunction` for tab completion. All other similar commands have this for better UX:

```suggestion
 var reconcileImagePolicyCmd = &cobra.Command{
     Use:   "policy [name]",
     Short: "Reconcile an ImagePolicy",
     Long:  `The reconcile image policy command triggers a reconciliation of an ImagePolicy resource and waits for it to finish.`,
     Example: `  # Trigger a scan for an existing image policy
   flux reconcile image policy alpine`,
     ValidArgsFunction: resourceNamesCompletionFunc(imagev1.GroupVersion.WithKind(imagev1.ImagePolicyKind)),
     RunE: reconcileCommand{
         apiType: imagePolicyType,
         object:  imagePolicyAdapter{&imagev1.ImagePolicy{}},
     }.run,
 }
```

cmd/flux/image.go
cappyzawa, line 1: The `resumable` interface requires `statusable`, which in turn requires `getObservedGeneration()`. This method is already implemented in `create_image_policy.go`:

```go
// create_image_policy.go L84-88
// getObservedGeneration is implemented here, since it's not
// (presently) needed elsewhere.
func (obj imagePolicyAdapter) getObservedGeneration() int64 {
    return obj.ImagePolicy.Status.ObservedGeneration
}
```

Now that you've added resume functionality, the comment "not presently needed elsewhere" is outdated. Consider moving this method to `image.go` alongside the other ImagePolicy adapter methods for better organization.

cmd/flux/reconcile_image.go
```diff
@@ -26,6 +27,19 @@ var reconcileImageCmd = &cobra.Command{
 	Long:  `The reconcile sub-commands trigger a reconciliation of image automation objects.`,
 }
 
+var reconcileImagePolicyCmd = &cobra.Command{
+	Use:   "policy [name]",
+	Short: "Reconcile an ImagePolicy",
+	Long:  `The reconcile image policy command triggers a reconciliation of an ImagePolicy resource and waits for it to finish.`,
+	Example: `  # Trigger an scan for an existing image policy
```
cappyzawa: The example has a typo and could be more consistent with other reconcile commands:

```suggestion
	Example: `  # Trigger a reconciliation for an existing image policy
```

Note: ImageRepository has the same typo, might want to fix that too!

cappyzawa (commented):
Since we already have E2E tests for image commands in `image_test.go`, it would be great to add test cases for the new commands. Other resources like HelmRelease, Kustomization, and OCIRepository all have suspend/resume/reconcile tests:

```go
// Additional test cases in TestImageScanning
{
    "suspend image policy podinfo-semver",
    "testdata/image/suspend_image_policy.golden",
},
{
    "resume image policy podinfo-semver", 
    "testdata/image/resume_image_policy.golden",
},
{
    "reconcile image policy podinfo-semver",
    "testdata/image/reconcile_image_policy.golden",
},
```

This would ensure consistency with other resources and help catch any regressions.

cmd/flux/reconcile_image.go
matheuscscp: Also, the resource-specific commands are defined inside their own `.go` files. We should introduce `suspend_image_policy.go`, `resume_image_policy.go` and `reconcile_image_policy.go`

cmd/flux/image_test.go
```diff
@@ -53,6 +53,18 @@ func TestImageScanning(t *testing.T) {
...
+			"testdata/image/suspend_image_policy.golden",
+		},
+		{
+			"resume image policy podinfo-semver",
+			"testdata/image/resume_image_policy.golden",
+		},
+		{
+			"reconcile image policy podinfo-semver",
+			"testdata/image/reconcile_image_policy.golden",
+		},
```
lukas8219, line 67: Is this really enough for integration test? I am afraid running `make test` is not working completely on my env.  Bunch of `covdata` related issue and `resume image policy` doesn't seem to be matching `Use:   "policy [name]"`

stefanprodan: The `covdata` errors are due to using stale Go tooling. Please make sure you are using Go 1.25 then clear the Go cache.

cmd/flux/reconcile_image_policy.go
```diff
@@ -0,0 +1,24 @@
+package main
```
matheuscscp, line 17: Please add the license header to all new files

cmd/flux/get_image_policy.go
```diff
@@ -74,7 +74,7 @@ func init() {
 func (s imagePolicyListAdapter) summariseItem(i int, includeNamespace bool, includeKind bool) []string {
 	item := s.Items[i]
 	status, msg := statusAndMessage(item.Status.Conditions)
-	return append(nameColumns(&item, includeNamespace, includeKind), item.Status.LatestImage, status, msg)
+	return append(nameColumns(&item, includeNamespace, includeKind), item.Status.LatestRef.Name, status, msg)
```
matheuscscp:
```suggestion
	return append(nameColumns(&item, includeNamespace, includeKind), item.Status.LatestRef.Name, item.Status.LatestRef.Tag, status, msg)
```

cmd/flux/get_image_policy.go
```diff
@@ -74,7 +74,7 @@ func init() {
...
 }
 
 func (s imagePolicyListAdapter) headers(includeNamespace bool) []string {
```
matheuscscp, line 85: We also need to change the column names from:

```go
headers := []string{"Name", "Latest image", "Ready", "Message"}
```

To:

```go
headers := []string{"Name", "Image", "Tag", "Ready", "Message"}
```

This matches what we did in the kubectl printer columns here:

https://github.com/fluxcd/image-reflector-controller/blob/fe23c393a458eef24c4d7a100efd47a695e58e8c/api/v1beta2/imagepolicy_types.go#L204-L205

matheuscscp:
@lukas8219 image-reflector-controller `v1.0.0` is released and updated in `flux2`: https://github.com/fluxcd/flux2/pull/5517

This PR is now unblocked. You can rebase and wrap up the work, there will be less things to do in the PR now.

matheuscscp (approved):
LGTM! :rocket: 

Thanks very much!

PR #5508, opened 2025-09-08
ci: Align azure e2e tests secret names with fluxcd/pkg

matheuscscp:
This aligns all the secrets between flux2 and pkg. @stefanprodan with this you can copy everything under `Stefan Azure Int Tests` in 1password to this repo, just like we did for pkg.


stefanprodan:
All done now

matheuscscp:
Passed: https://github.com/fluxcd/flux2/actions/runs/17561102671/job/49877455790

PR #5509, opened 2025-09-09
`fluxcd/flux2/action@main`: Determine latest version without using GitHub API

RussellAult:
Currently, creating a workflow that `uses: fluxcd/flux2/action@main` requires a GitHub API token, even the workflow is being run on a platform other than GitHub (e.g. [Forgejo](https://forgejo.org/docs/latest/user/actions/overview/)). This is because the GitHub API is called to determine the latest version (and only to determine the latest version), and GitHub rate-limits access to their API endpoint.

This PR modifies `action.yml` so that the action gathers latest version information from `github.com` instead of from `api.github.com`, which avoids the rate-limiting problem entirely. This, in turn, allows for the removal of the token-related input/variable, and removes a usability pain-point for non-GitHub platforms


action/action.yml
```diff
@@ -15,9 +15,6 @@ inputs:
   bindir:
     description: "Alternative location for the Flux binary, defaults to path relative to $RUNNER_TOOL_CACHE."
     required: false
-  token:
-    description: "Token used to authentication against the GitHub.com API. Defaults to the token from the GitHub context of the workflow."
-    required: false
```
stefanprodan, line 20: This is a breaking change that we can't do. Please restore the input with the description: `Token used to authentication against the GitHub.com API.`.

In the script, we can remove the default `TOKEN=${{ github.token }}`. If `${TOKEN}` is empty, then we can call the curl command you added, if it's set, then we call the GH API.

RussellAult: > This is a breaking change that we can't do.

Ah shoot, I was wondering about that.
 
> In the script, we can remove the default `TOKEN=${{ github.token }}`. If `${TOKEN}` is empty, then we can call the curl command you added, if it's set, then we call the GH API.

This should now do what you're requesting!

stefanprodan:
All tests are failing with  `line 4: conditional binary operator expected`

RussellAult:
> All tests are failing with `line 4: conditional binary operator expected`

Sorry, forgot to put quotes around `${{ inputs.token }}` in the conditional; should pass now.

stefanprodan:
@RussellAult can you please squash and force push.

RussellAult:
> @RussellAult can you please squash and force push.

Done!

stefanprodan (approved):
LGTM

Thanks @RussellAult

action/action.yml
```diff
@@ -26,13 +26,12 @@ runs:
...
         VERSION=${{ inputs.version }}
 
-        TOKEN=${{ inputs.token }}
-        if [[ -z "$TOKEN" ]]; then
-          TOKEN=${{ github.token }}
-        fi
-
         if [[ -z "$VERSION" ]] || [[ "$VERSION" = "latest" ]]; then
-          VERSION=$(curl -fsSL -H "Authorization: token ${TOKEN}" https://api.github.com/repos/fluxcd/flux2/releases/latest | grep tag_name | cut -d '"' -f 4)
+          if [[ "${{ inputs.token }}" != '' ]]; then
```
matheuscscp: Interpolating values from GitHub Actions in shell scripts is a bad practice. Can you please pass the token through an environment variable to the shell script? Like this:

```yaml
      env:
        TOKEN: ${{ inputs.token }}
```

No need to fix the other inputs now, but it would be good to fix the one you are touching in this PR :pray:

RussellAult: That's done; it was easy enough to move the other two as well (but I did that as a separate commit if you'd rather that go to a separate PR).

FWIW, after my "all tests are failing" experience earlier in PR, I decided to `""` the interpolations in the `env` key, just because I wasn't sure that an empty `input.` without the quotes would still produce schema-compliant YAML

action/action.yml
```diff
@@ -16,23 +16,24 @@ inputs:
     description: "Alternative location for the Flux binary, defaults to path relative to $RUNNER_TOOL_CACHE."
     required: false
   token:
-    description: "Token used to authentication against the GitHub.com API. Defaults to the token from the GitHub context of the workflow."
+    description: "Token used to authentication against the GitHub.com API."
```
matheuscscp:
```suggestion
    description: "Token used to authenticate against the GitHub.com API."
```

RussellAult: Hah, whoops, didn't bother to read the rest of the line.

stefanprodan:
@RussellAult can you please squash, rebase with upstream main and force push.

Issue #5516, opened 2025-09-15
Prepare for Flux v2.7.0


stefanprodan:
Flux 2.7.0 promotes the Image Automation APIs to v1 (GA) and comes with new features, bug fixes and performance improvements. 

Flux 2.7.0 comes with a new set of APIs (`ExternalArtifact` & `ArtifactGenerator`) and a new controller [source-watcher](https://github.com/fluxcd/source-watcher) that extends Flux with advanced source composition and decomposition patterns.

:date: **Target dates:** 29-30 September   

🚀  **Roadmap milestone:** https://fluxcd.io/roadmap/#v27-q3-2025

### Highlights

- General availability release for the Flux Image Automation APIs (`ImagePolicy`, `ImageRepository`, `ImageUpdateAutomation`)
- Watch for changes in ConfigMaps and Secrets references (`Kustomization`, `HelmRelease`)
- Extend the readiness evaluation of dependencies with CEL expressions (`Kustomization`, `HelmRelease`)
- Introduce `RetryOnFailure` lifecycle management strategy (`HelmRelease`)
- Support mTLS for sending alerts to external systems (`Provider`)
- Support mTLS for GitHub App transport (`GitRepository`, `ImageUpdateAutomation`, `Provider`)
- Support for remote cluster authentication using Workload Identity (`Kustomization`, `HelmRelease`)
- Support object-level workload identity auth for Azure DevOps, Azure Blob, GCP Pub/Sub, GCP Storage, AWS S3
- Support for 3rd-party source controllers (`ExternalArtifact`)
- Support for source composition and decomposition patterns (`ArtifactGenerator`)

### Dependencies updates

- [x] Go 1.25
- [x] Kubernetes 1.34
- [x] Controller Runtime 0.22.1
- [x] Kustomize 5.7.1
- [x] Helm 3.19.0

### Release checklist

- [ ] source-controller v1.7.0
- [ ] source-watcher v2.0.0
- [ ] kustomize-controller v1.7.0
- [ ] notification-controller v1.7.0
- [ ] helm-controller v1.4.0
- [ ] image-reflector-controller v1.0.0
- [ ] image-automation-controller v1.0.0
- [ ] flux-benchmark v2.7.0
- [ ] flux2 v2.7.0
- [ ] terraform-provider-flux v1.7.0

### Conformance testing

| Distribution | Versions |
|:--------|:--------|
| Kubernetes KinD | 1.32, 1.33, 1.34  |
[...]

PR #5533, opened 2025-09-24
Set Kubernetes 1.32 as min supported version

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/5516


matheuscscp (approved):

stefanprodan:
@matheuscscp going to merge this, the GCP KMS tests is failing, we need to investigate if this is flake or not.

Issue #5545, opened 2025-09-30
Extend `flux migrate` to work with local files


stefanprodan:
Currently, the `flux migrate` command updates the Flux custom resources to their latest API version in the target cluster. We should extend this command with a `-f` flag to help users migrate their YAML manifests stored in Git repositories.

The `flux migrate -f /path/to/manifests` should do the following:
- scans recursively the specified directory and sub-dirs for `*.yaml` files
- reads each file and identifies Flux custom resources by their `apiVersion` and `kind`
- the files can contain multiple resources separated by `---`, the command should correctly parse and handle them
- for each identified Flux custom resource, it should check if there is a newer API version available for that resource kind
- if a newer API version is found, it should update the `apiVersion` field in the manifest
- the `apiVersion` update should be done using string replacement to preserve formatting and comments

matheuscscp:
This new `flux migrate` variant will be released soon in Flux 2.7.1.

bestrocker221:
Using flux cli v2.7.2 and the `-f .` does not work, it returns "✔ no custom resources found that require migration"
When running `flux migrate` it will correctly update the versions in the  cluster, but local to the git repo it does not seem to work for me. Even pointing the path to the exact  file where there is a v1beta1 is returning empty.

matheuscscp:
Works for me:

```
flux migrate -f .
► starting migration of custom resources
✚ manifest.yaml:1: ImageRepository v1beta2 -> v1
? Are you sure you want to proceed with the above upgrades? [y/N] y
✔ file manifest.yaml migrated successfully
✔ custom resources migrated successfully
```

matheuscscp:
@bestrocker221 How does your manifest file look like?

matheuscscp:
Mine looks like this:

```
cat manifest.yaml 
apiVersion: image.toolkit.fluxcd.io/v1beta2
kind: ImageRepository
...
```

bestrocker221:
One example:
```
$ head -n 1 kustomization.yaml 
apiVersion: kustomize.config.k8s.io/v1beta1
$ flux migrate -v 2.6 -f . (or direct path to kustomization.yaml)                 
► starting migration of custom resources
✔ no custom resources found that require migration
```

matheuscscp:
Bingo. `kustomize.config.k8s.io/v1beta1` is not a Flux Kustomization, it's a kustomize Kustomization.

bestrocker221:
Thank you, I noticed my mistake there. 
[...]

PR #5554, opened 2025-10-01
Extend flux migrate to work with local files

matheuscscp:
Closes: #5545 

TODO:

- [ ] Add tests.


cmd/flux/migrate.go
```diff
@@ -52,17 +189,46 @@ func runMigrateCmd(cmd *cobra.Command, args []string) error {
...
 	cfg, err := utils.KubeConfig(kubeconfigArgs, kubeclientOptions)
 	if err != nil {
-		return fmt.Errorf("Kubernetes client initialization failed: %s", err.Error())
+		return fmt.Errorf("the Kubernetes client initialization failed: %w", err)
 	}
 
 	kubeClient, err := client.New(cfg, client.Options{Scheme: utils.NewScheme()})
 	if err != nil {
 		return err
 	}
```
stefanprodan, line 211: I think all this code should be run only if the live cluster migration is specified, as this could fail if no kubeconfig is present on the machine.

cmd/flux/migrate.go
```diff
@@ -182,3 +354,225 @@ func (m *Migrator) getStorageVersion(crd *apiextensionsv1.CustomResourceDefiniti
...
+		idx := strings.Index(apiVersionLine, apiVersionPrefix)
+		if idx == -1 {
+			continue
+		}
+		apiVersion := strings.TrimSpace(apiVersionLine[idx+len(apiVersionPrefix):])
+		gv, err := schema.ParseGroupVersion(apiVersion)
+		if err != nil {
+			// Not sure what to do here. Erroring out could be too harsh.
+			continue
+		}
```
stefanprodan, line 592: Let's log the filename, line number and error, then continue

cmd/flux/migrate.go
```diff
@@ -182,3 +354,225 @@ func (m *Migrator) getStorageVersion(crd *apiextensionsv1.CustomResourceDefiniti
...
+
+	for _, upgrade := range fileUpgrades.Upgrades {
+		line := lines[upgrade.Line]
+		line = strings.ReplaceAll(line, upgrade.OldVersion, upgrade.NewVersion)
+		lines[upgrade.Line] = line
+	}
+
+	b = []byte(strings.Join(lines, "\n"))
+
+	if err := os.WriteFile(fileUpgrades.File, b, 0644); err != nil {
```
stefanprodan: We could preserve the original file permission

cmd/flux/migrate.go
```diff
@@ -182,3 +354,225 @@ func (m *Migrator) getStorageVersion(crd *apiextensionsv1.CustomResourceDefiniti
...
+	}
+	lines := strings.Split(string(b), "\n")
+
+	for _, upgrade := range fileUpgrades.Upgrades {
+		line := lines[upgrade.Line]
+		line = strings.ReplaceAll(line, upgrade.OldVersion, upgrade.NewVersion)
+		lines[upgrade.Line] = line
+	}
+
+	b = []byte(strings.Join(lines, "\n"))
```
stefanprodan: Are we preserving empty lines and Windows line endings?

cmd/flux/migrate.go
```diff
@@ -182,3 +354,225 @@ func (m *Migrator) getStorageVersion(crd *apiextensionsv1.CustomResourceDefiniti
...
+	if !fileInfo.IsDir() {
+		return []string{f.path}, nil
+	}
+	var files []string
+	err = fs.WalkDir(f.fileSystem, f.path, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if !d.IsDir() {
+			files = append(files, path)
```
stefanprodan, line 498: This would append OS devices and symlinks, please use https://github.com/fluxcd/pkg/blob/edd6c51a6ac04bcb275bb909e823c87de704e275/oci/build.go#L79-L82

cmd/flux/migrate.go
```diff
@@ -52,17 +189,46 @@ func runMigrateCmd(cmd *cobra.Command, args []string) error {
...
-		"app.kubernetes.io/part-of": "flux",
-	})
+	var migrator Migrator
+
+	if migrateFlags.path == "" {
+		migrator = NewClusterMigrator(kubeClient, client.MatchingLabels{
+			"app.kubernetes.io/part-of": "flux",
+		})
+	} else {
+		fileSystem := os.DirFS(".")
```
stefanprodan: Let's use `os.OpenRoot`

Form go docs:

> DirFS is therefore not a general substitute for a chroot-style security mechanism when the directory tree contains arbitrary content.
> Use [Root.FS](https://pkg.go.dev/os#Root.FS) to obtain a fs.FS that prevents escapes from the tree via symbolic links.

cmd/flux/migrate.go
matheuscscp: > Are we preserving empty lines

Definitely.

> and Windows line endings?

If a line is always delimited by `\r\n` in Windows, then definitely as well.

matheuscscp: Windows tested, both things work.

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
```diff
@@ -19,48 +19,198 @@ package main
...
 	Long: `The migrate command must be run before a Flux minor version upgrade.
 The command migrates the Flux custom resources stored in Kubernetes etcd to their latest API version,
 ensuring the Flux components can continue to function correctly after the upgrade.
+
+The command has two modes of operation:
+
+- Cluster mode (default): migrates all the custom resources in the cluster for
+the CRDs matching the label selector "app.kubernetes.io/part-of=flux".
+- File system mode: migrates the custom resources defined in the manifests
+located in the specified path.
```
stefanprodan:
```suggestion
The command has two modes of operation:

- Cluster mode (default): migrates all the custom resources stored in Kubernetes etcd to their latest API version.
- File system mode (-f): migrates the custom resources defined in the manifests located in the specified path.
```

cmd/flux/migrate.go
```diff
@@ -182,3 +366,244 @@ func (m *Migrator) getStorageVersion(crd *apiextensionsv1.CustomResourceDefiniti
...
+	return nil
+}
+
+func (f *FileSystemMigrator) listFiles() ([]string, error) {
+	fileInfo, err := fs.Stat(f.fileSystem, f.path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat path %s: %w", f.path, err)
+	}
+	if !fileInfo.IsDir() {
+		return []string{f.path}, nil
```
stefanprodan: This can be a symlink or device, please add the IsRegular() check

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

cmd/flux/migrate.go
matheuscscp: Fixed, thanks!

PR #5566, opened 2025-10-05
refactor: convert `Kustomization` resource into unstructured map only once during variable substitution

ramasai1:
The `Kustomization` resource only needs to be converted into its unstructured form once.

The unstructured form is not being mutated, only accessed, so converting it once for all resources should suffice.

Existing tests for this should be good enough in my view since I am not introducing new behavior.


stefanprodan (approved):
LGTM

Thanks @ramasai1

Issue #5504, opened 2025-09-03
[RFC-0012] Implement ExternalArtifact API and SDK


stefanprodan:
This issue tracks the implementation of [RFC-0012](https://github.com/fluxcd/flux2/blob/main/rfcs/0012-external-artifact/README.md).

API Tasks:
- [x] [Extract the `Artifact` type from source-controller to `github.com/fluxcd/pkg/apis/meta` package](https://github.com/fluxcd/pkg/pull/1018)
- [x] Refactor source-controller to use the `Artifact` type from the meta package
- [x] Add the `ExternalArtifact` API to the source-controller API package
- [x] Add documentation for `ExternalArtifact` to source-controller API docs
- [x] Add the `ExternalArtifact` API as a supported `sourceRef` in kustomize-controller including watchers
- [x] Add the `ExternalArtifact` feature gate to kustomize-controller
- [x] Add the `ExternalArtifact` as a supported `chartRef` in helm-controller including watchers
- [x] Add the `ExternalArtifact` feature gate to helm-controller
- [x] Add documentation for the feature gate and supported sources in kustomize-controller and helm-controller

SDK Tasks:
- [x] [Extract the source-controller storage component to a dedicated package](https://github.com/fluxcd/source-controller/pull/1864)
- [x] Refactor source-controller to use the  `github.com/fluxcd/pkg/artifact` package
- [ ] Define the public interfaces for the SDK
- [ ] Implement the server, client and storage components
- [ ] Add development guide for the SDK to the GitOps Toolkit documentation

Artifact Generators Tasks:
- [x] [Implement artifact generators in source-watcher controller using the ExternalArtifact API and SDK](https://github.com/fluxcd/source-watcher/pull/255)
- [ ] Develop conformance testing for ExternalArtifact generators and consumers

CLI Tasks:
- [x] Add support for ExternalArtifact source type to the following commands:
   - `flux trace` (display `ExternalArtifact.spec.sourceRef`)
   - `flux reconcile ks/hr --with-source` (skip external artifact annotation)
- [x] Add command `flux get source external`
- [x] Add command `flux export source external`

dgunzy:
@stefanprodan if there is a small part of this I can take on I would be keen to help 😃

stefanprodan:
SDK docs: https://fluxcd.io/flux/gitops-toolkit/external-artifact-sdk/

PR #5583, opened 2025-10-10
[RFC-0012] Add command `flux export source external`

dgunzy:
Part of: https://github.com/fluxcd/flux2/issues/5504
[RFC-0012] Add command flux export source external

@stefanprodan  - I added this, one thing I am not sure about is there is no `SecretRef` so I am just returning `nil` 
let me know if there is a better way to go about this. 

```
~/OpenSource/flux/flux2-fork add-export-source-external ?1 ❯ ./bin/flux export source external --all -n decomposition                                                     ⎈ kind-flux-test
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: ExternalArtifact
metadata:
  labels:
    app.kubernetes.io/managed-by: source-watcher
    source.extensions.fluxcd.io/generator: 4b83e39f-4032-4808-a66f-3adea8878a37
  name: podinfo-backend
  namespace: decomposition
spec:
  sourceRef:
    apiVersion: source.extensions.fluxcd.io/v1beta1
    kind: ArtifactGenerator
    name: app-gen
    namespace: decomposition
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: ExternalArtifact
metadata:
  labels:
    app.kubernetes.io/managed-by: source-watcher
    source.extensions.fluxcd.io/generator: 4b83e39f-4032-4808-a66f-3adea8878a37
  name: podinfo-cache
  namespace: decomposition
spec:
  sourceRef:
    apiVersion: source.extensions.fluxcd.io/v1beta1
    kind: ArtifactGenerator
    name: app-gen
    namespace: decomposition
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: ExternalArtifact
metadata:
  labels:
    app.kubernetes.io/managed-by: source-watcher
    source.extensions.fluxcd.io/generator: 4b83e39f-4032-4808-a66f-3adea8878a37
  name: podinfo-frontend
  namespace: decomposition
spec:
  sourceRef:
    apiVersion: source.extensions.fluxcd.io/v1beta1
    kind: ArtifactGenerator
    name: app-gen
    namespace: decomposition
```

### DIFF
--- a/cmd/flux/export_source_external.go
+++ b/cmd/flux/export_source_external.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2025 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+)
+
+var exportSourceExternalCmd = &cobra.Command{
+	Use:   "external [name]",
+	Short: "Export ExternalArtifact sources in YAML format",
+	Long:  "The export source external command exports one or all ExternalArtifact sources in YAML format.",
+	Example: `  # Export all ExternalArtifact sources
+  flux export source external --all > sources.yaml
+
+  # Export a specific ExternalArtifact
+  flux export source external my-artifact > source.yaml`,
+	ValidArgsFunction: resourceNamesCompletionFunc(sourcev1.GroupVersion.WithKind(sourcev1.ExternalArtifactKind)),
+	RunE: exportWithSecretCommand{
+		list:   externalArtifactListAdapter{&sourcev1.ExternalArtifactList{}},
+		object: externalArtifactAdapter{&sourcev1.ExternalArtifact{}},
+	}.run,
+}
+
+func init() {
+	exportSourceCmd.AddCommand(exportSourceExternalCmd)
+}
+
+func exportExternalArtifact(source *sourcev1.ExternalArtifact) any {
+	gvk := sourcev1.GroupVersion.WithKind(sourcev1.ExternalArtifactKind)
+	export := sourcev1.ExternalArtifact{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       gvk.Kind,
+			APIVersion: gvk.GroupVersion().String(),
+		},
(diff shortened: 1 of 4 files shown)


cmd/flux/export_source_external.go
```diff
@@ -0,0 +1,84 @@
...
+		list:   externalArtifactListAdapter{&sourcev1.ExternalArtifactList{}},
+		object: externalArtifactAdapter{&sourcev1.ExternalArtifact{}},
+	}.run,
+}
+
+func init() {
+	exportSourceCmd.AddCommand(exportSourceExternalCmd)
+}
+
+func exportExternalArtifact(source *sourcev1.ExternalArtifact) interface{} {
```
matheuscscp:
```suggestion
func exportExternalArtifact(source *sourcev1.ExternalArtifact) any {
```

cmd/flux/export_source_external.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (ex externalArtifactAdapter) secret() *types.NamespacedName {
+	return getExternalArtifactSecret(ex.ExternalArtifact)
+}
+
+func (ex externalArtifactListAdapter) secretItem(i int) *types.NamespacedName {
+	return getExternalArtifactSecret(&ex.ExternalArtifactList.Items[i])
+}
+
+func (ex externalArtifactAdapter) export() interface{} {
```
matheuscscp:
```suggestion
func (ex externalArtifactAdapter) export() any {
```

cmd/flux/export_source_external.go
```diff
@@ -0,0 +1,84 @@
...
+	return exportExternalArtifact(ex.ExternalArtifact)
+}
+
+func (ex externalArtifactListAdapter) exportItem(i int) interface{} {
```
matheuscscp:
```suggestion
func (ex externalArtifactListAdapter) exportItem(i int) any {
```

dgunzy:
@matheuscscp  - updated to `any` 

I agree, do you think I should change export to also export the status? It is a bit of a different pattern then I have seen elsewhere, but happy to do it!

matheuscscp:
> @matheuscscp - updated to `any`
> 
> I agree, do you think I should change export to also export the status? It is a bit of a different pattern then I have seen elsewhere, but happy to do it!

Maybe only for ExternalArtifact... @stefanprodan Thoughts?

matheuscscp:
On a second thought, I'm not quite sure if exporting `.status` makes sense depending on the use cases of this command. I think `flux export` is supposed to be used for getting resources out of the cluster for checking them into Git, and `.status` is always managed by something else (e.g. a controller), and not by the Flux appliers, as they don't patch the status subresource of the applied objects.

One could also end up thinking: "Well, `ExternalArtifact` is automatically created by `source-watcher` as a result of reconciling `ArtifactGenerator`, so what's the point of exporting `ExternalArtifact` objects anyway?" And I guess the use case here lies in the `.spec.sourceRef` field. An external source controller could respond to an `ExternalArtifact` object with `.spec.sourceRef.kind` set to a kind that it manages by patching the status of this EA.

@stefanprodan I guess this is what we discussed before, right?

stefanprodan:
The export functions shouldn't print the status. In the case of EA, export is useful when the generation and reconciliation is decoupled, e.g. there is a dedicated controller that watcher external artifacts and sets their status.

dgunzy:
@stefanprodan @matheuscscp  - I added a test in `export_test.go`  - if there is a good place to add some additional tests let me know!

PR #5589, opened 2025-10-13
fix: handle error when writing password prompt to stdout

akshatsinha0:
The readPasswordFromStdin funcn ignored the error from fmt.Fprint,which could cause silent failures if stdout is closed or redirected.

Signed-off-by: akshatsinha0

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -180,7 +180,7 @@ func main() {
 
 	// This is required because controller-runtime expects its consumers to
 	// set a logger through log.SetLogger within 30 seconds of the program's
-	// initalization. If not set, the entire debug stack is printed as an
+	// initialization. If not set, the entire debug stack is printed as an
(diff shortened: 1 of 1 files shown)


cmd/flux/main.go
```diff
@@ -225,7 +225,9 @@ func configureDefaultNamespace() {
 func readPasswordFromStdin(prompt string) (string, error) {
 	var out string
 	var err error
-	fmt.Fprint(os.Stdout, prompt)
+	if _, err:=fmt.Fprint(os.Stdout,prompt); err!=nil{
+		return "",fmt.Errorf("failed to write prompt:  %w",err)
+	}
```
stefanprodan: This code doesn't look properly formatted, please run `make test` and force push.

akshatsinha0:
Should I edit the pr title as well?

akshatsinha0:
@stefanprodan 
I squashed.

PR #5590, opened 2025-10-13
fix(cli): correct typo initalization to initialization

akshatsinha0:
Just randomly saw this mistake (not necessary, !:)) while searching


stefanprodan:
add this to the previous PR you opened, squash all commits there into a single one please.

akshatsinha0:
Ok

PR #5597, opened 2025-10-21
Allow option to skip tenant namespace creation

anshuishere:
## Issue
When flux onboards tenants in multitenant setup, it always creates manifest for namespace with no option to skip it.

## When this is needed?
In a shared cluster, for example in a customer environment, namespaces are managed at infra level by cluster admin. 
This means we receive a namespace already created and we don't want flux platform-admin to manage it or delete it in case of tenant removal. 

## What's in this PR?
This allows a flag to flux create tenant command to optionally skip namespace creation by flux. Default behaviour remains unchanged.

### DIFF
--- a/cmd/flux/create_tenant.go
+++ b/cmd/flux/create_tenant.go
@@ -58,9 +58,10 @@ const (
 )
 
 type tenantFlags struct {
-	namespaces  []string
-	clusterRole string
-	account     string
+	namespaces    []string
+	clusterRole   string
+	account       string
+	skipNamespace bool
 }
 
 var tenantArgs tenantFlags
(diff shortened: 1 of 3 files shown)


anshuishere:
Hello, 
@matheuscscp @stefanprodan 
Can you please review this when you find sometime.

cmd/flux/create_tenant.go
```diff
@@ -69,6 +70,7 @@ func init() {
 	createTenantCmd.Flags().StringSliceVar(&tenantArgs.namespaces, "with-namespace", nil, "namespace belonging to this tenant")
 	createTenantCmd.Flags().StringVar(&tenantArgs.clusterRole, "cluster-role", "cluster-admin", "cluster role of the tenant role binding")
 	createTenantCmd.Flags().StringVar(&tenantArgs.account, "with-service-account", "", "service account belonging to this tenant")
+	createTenantCmd.Flags().BoolVar(&tenantArgs.skipNamespace, "skip-namespace-create", false, "skip namespace creation (namespace must exist already)")
```
stefanprodan: This can just be `--skip-namespace` same flag we use for uninstall

anshuishere: The flag is updated now.

stefanprodan:
Please rebase your fork with upstream main and squash the commits into one and force push. This PR should contain a single commit.

Issue #5066, opened 2024-11-08
Change flags with fixed allowed values to show the actual values


darkowlzz:
Currently, a lot of the flags that accept various fixed values show arbitrary type name instead of the actual values they accept. For example, in `flux create source oci -h`

```
--ignore-paths strings                      set paths to ignore resources (can specify multiple paths with commas: path1,path2)
--insecure                                  for when connecting to a non-TLS registries over plain HTTP
--provider sourceOCIProvider                the OCI provider name, available options are: (generic, aws, azure, gcp) (default generic)
--proxy-secret-ref string                   the name of an existing secret containing the proxy address and credentials
```

`--provider` accepts a fixed set of values. But unlike the other flags that show the known type of what's accepted, it shows `sourceOCIProvider`. This type is not known to the user, unlike int or string. It would be much better to list all the values. This would also shorten the flag description.

This has already been implemented in `flux create source git` for `--provider` flag

```
--private-key-file string                path to a passwordless private key file used for authenticating to the Git SSH server
--provider generic|azure                 the Git provider name
--proxy-secret-ref string                the name of an existing secret containing the proxy address and credentials
```

Refer https://github.com/fluxcd/flux2/blob/v2.4.0/internal/flags/source_git_provider.go#L51-L53 for the implementation.

The pflag [`Values.Type()`](https://pkg.go.dev/github.com/spf13/pflag#Value) implementation should append the allowed values and return it, instead of returning an internal data type name that's not known to the user.

All such flags in the CLI subcommands should be changed accordingly.

This was discussed and agreed upon in a Flux dev meeting.

richyeki:
Hello @darkowlzz may I work on it?

jaxels10:
Since nothing has happened for a long time on this issue, I have given it a try in this [pull request](https://github.com/fluxcd/flux2/pull/5602) @darkowlzz

PR #5602, opened 2025-10-26
fix: return accepted values for flags when calling Values.Type()

jaxels10:
Resolves: #5066 which requested each flags implementation of Values.Type() should return the accepted values.

Feedback welcome.

Notes:
1. I have not done any changes to flags which only has a single accepted value. This is to not confuse the user. Let me know if I need to change this
2. I could not find a reference for `internal/flags/local_helm_chart_source.go` and it seems the `supportedHelmChartSourceKinds` is never defined in the file

### DIFF
--- a/internal/flags/gitlab_visibility.go
+++ b/internal/flags/gitlab_visibility.go
@@ -58,9 +58,13 @@ func (d *GitLabVisibility) Set(str string) error {
 }
 
 func (d *GitLabVisibility) Type() string {
-	return "gitLabVisibility"
+	keys := make([]string, 0, len(supportedGitLabVisibilities))
+	for v := range supportedGitLabVisibilities {
+		keys = append(keys, string(v))
+	}
+	return strings.Join(keys, "|")
 }
 
 func (d *GitLabVisibility) Description() string {
-	return fmt.Sprintf("specifies the visibility of the repository. Valid values are public, private, internal")
+	return "specifies the visibility of the repository"
 }
--- a/internal/flags/helm_chart_source.go
+++ b/internal/flags/helm_chart_source.go
@@ -64,7 +64,7 @@ func (s *HelmChartSource) Set(str string) error {
 }
 
 func (s *HelmChartSource) Type() string {
-	return "helmChartSource"
+	return "string"
 }
 
 func (s *HelmChartSource) Description() string {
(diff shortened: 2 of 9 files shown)


internal/flags/gitlab_visibility.go
```diff
@@ -58,9 +58,13 @@ func (d *GitLabVisibility) Set(str string) error {
...
+	keys := make([]string, 0, len(supportedGitLabVisibilities))
+	for v := range supportedGitLabVisibilities {
+		keys = append(keys, string(v))
+	}
+	return strings.Join(keys, "|")
 }
 
 func (d *GitLabVisibility) Description() string {
-	return fmt.Sprintf("specifies the visibility of the repository. Valid values are public, private, internal")
+	return "specifies the visibility of the repository. Valid values are public, private, internal"
```
matheuscscp:
```suggestion
	return "specifies the visibility of the repository"
```

internal/flags/helm_chart_source.go
```diff
@@ -64,13 +64,9 @@ func (s *HelmChartSource) Set(str string) error {
...
+	return strings.Join(supportedHelmChartSourceKinds, "|")
 }
 
 func (s *HelmChartSource) Description() string {
-	return fmt.Sprintf(
-		"source that contains the chart in the format '<kind>/<name>.<namespace>', "+
-			"where kind must be one of: (%s)",
-		strings.Join(supportedHelmChartSourceKinds, ", "),
-	)
+	return "source that contains the chart in the format '<kind>/<name>.<namespace>'"
```
matheuscscp: I'm not sure this fix make sense for this argument since the value has to match this format and the type is returning only the allowed kinds

internal/flags/kustomization_source.go
```diff
@@ -70,13 +70,9 @@ func (s *KustomizationSource) Set(str string) error {
...
+	return strings.Join(supportedKustomizationSourceKinds, "|")
 }
 
 func (s *KustomizationSource) Description() string {
-	return fmt.Sprintf(
-		"source that contains the Kubernetes manifests in the format '[<kind>/]<name>.<namespace>', "+
-			"where kind must be one of: (%s), if kind is not specified it defaults to GitRepository",
-		strings.Join(supportedKustomizationSourceKinds, ", "),
-	)
+	return "source that contains the Kubernetes manifests in the format '[<kind>/]<name>.<namespace>'. If kind is not specified it defaults to GitRepository"
```
matheuscscp: Same comment as for helm chart source

PR #5625, opened 2025-11-12
diff: report if object is skipped

hown3d:
If an object is skipped during `flux diff` it is not clearly communicated that the object is skipped.
This adds a warning output so that a user sees which resources are skipped.

### DIFF
--- a/internal/build/diff.go
+++ b/internal/build/diff.go
@@ -116,6 +116,9 @@ func (b *Builder) diff() (string, bool, error) {
 			diffErrs = append(diffErrs, err)
 			continue
 		}
+		if change.Action == ssa.SkippedAction {
+			output.WriteString(writeString(fmt.Sprintf("► %s skipped\n", change.Subject), bunt.Orange))
+		}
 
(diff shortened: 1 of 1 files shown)


internal/build/diff.go
```diff
@@ -116,6 +116,9 @@ func (b *Builder) diff() (string, bool, error) {
 			diffErrs = append(diffErrs, err)
 			continue
 		}
+		if change.Action == ssa.SkippedAction {
+			output.WriteString(writeString(fmt.Sprintf("⚠️ %s skipped\n", ssautil.FmtUnstructured(obj)), bunt.Orange))
```
stefanprodan:
```suggestion
			output.WriteString(writeString(fmt.Sprintf("► %s skipped\n", change.Subject), bunt.Orange))
```

stefanprodan:
Can you please rebase with upstream main and force push

matheuscscp:
@hown3d We'll release a patch tomorrow and we can include this. Can you rebase pls? :pray:

PR #5656, opened 2025-12-10
fix: #5654 by checking if both --chart and --chart-ref are set

### LINKED ISSUE #5654: `flux create helmrelease`: Clarify that `--chart` cannot be used with OCIRepository
When combining `--chart-ref=OCIRepository` with the `--chart` flag, `flux create helmrelease` creates an invalid resource.

```
flux create helmrelease my-hr \
    --namespace=my-namespace \
    --chart-ref=OCIRepository/my-oci-repository.my-namespace \
    --chart my-chart \
    --export
```

Result:

```
---
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: my-hr
  namespace: my-namespace
spec:
  chart:
    spec:
      chart: my-chart
      reconcileStrategy: ChartVersion
      sourceRef:
        name: ""
  interval: 1m0s
```

Applying this resource will result in an error message:

> ✗ HelmRelease.helm.toolkit.fluxcd.io "my-hr" is invalid: [spec.chart.spec.sourceRef.name: Invalid value: "": spec.chart.spec.sourceRef.name in body should be at least 1 chars long, spec.chart.spec.sourceRef.kind: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]

## Suggestion

The CLI should fail early with an error message indicating that `--chart` cannot be used in combination with `--chart-ref`.

jaxels10:
Add a test to see if --chart and --chart ref are set at the same time.
Implement check in `createHelmReleaseCmdRun` to make sure CLI returns error if both flags are set at the same time.
Fixes #5654

### DIFF
--- a/cmd/flux/create_helmrelease.go
+++ b/cmd/flux/create_helmrelease.go
@@ -182,6 +182,10 @@ func createHelmReleaseCmdRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("chart or chart-ref is required")
 	}
 
+	if helmReleaseArgs.chart != "" && helmReleaseArgs.chartRef != "" {
+		return fmt.Errorf("cannot use --chart in combination with --chart-ref")
+	}
+
 	sourceLabels, err := parseLabels()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


marians:
Thank you @jaxels10 !

jaxels10:
No worries @marians . Glad I could help

PR #5659, opened 2025-12-11
Added retry logic with delays to the Flux CLI download

ivan-munteanu:
Added retry logic with delays to the Flux CLI download. 
We run into failed to download the cli binary multiple times. changes the following
- Downloads now retry up to 5 times with 5-second delays between attempts
- Applied to both the Flux binary and checksums file downloads
- Clear feedback on retry attempts and failures

Why This Matters
- Improves Reliability: Network hiccups, rate limiting, or temporary outages no longer cause immediate failures. The action automatically recovers from transient issues

Impact
- No breaking changes
- No performance impact (delays only occur on actual failures)
- Fully backward compatible

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -77,9 +77,37 @@ runs:
 
           FLUX_DOWNLOAD_URL="https://github.com/fluxcd/flux2/releases/download/v${VERSION}/"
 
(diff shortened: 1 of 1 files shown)


stefanprodan:
Can you please rename the commit's first line to "Added retry logic to install.sh"

ivan-munteanu:
done

PR #5674, opened 2026-01-12
fix: normalize path for Windows compatibility

### LINKED ISSUE #5673: `flux build kustomization` fails on Windows
### Describe the bug

`flux build kustomization` fails on Windows when using a `--path` other than `.`. On Linux, all invocations with `.`, a relative path or an absolute path for `--path` succeed and correctly render the manifests.

### Steps to reproduce

1. Create a local folder `test` with a `kustomization.yaml`
    ```apiVersion: kustomize.config.k8s.io/v1beta1
    kind: Kustomization
    
    resources:
    - test.yaml
    ```
2. Create `test.yaml` in the `test` folder with some test manifest like:
    ```apiVersion: v1
    kind: Pod
    metadata:
      name: test-pod
      namespace: default
    spec:
      containers:
      - name: test-container
        image: nginx:latest
    ```
3. Create a `test.yaml` with:
    ```apiVersion: kustomize.toolkit.fluxcd.io/v1
    kind: Kustomization
    metadata:
      name: test
      namespace: default
    ```
4. `cd` into the test folder on Windows and run `flux build kustomization test --namespace default --kustomization-file ..\test.yaml --path . --dry-run`
5. The commands succeeds and prints the pod manifest
6. `cd ..` and then `flux build kustomization test --namespace default --kustomization-file .\test.yaml --path .\test\ --dry-run`
7. Observe the error `✗ failed to generate kustomization.yaml: failed to save original kustomization.yaml: open test\test\kustomization.yaml.original: The system cannot find the path specified. <nil> <nil>`
8. Run `flux build kustomization test --namespace default --kustomization-file C:\<absolute-path>\test.yaml --path C:\<absolute-path>\test\ --dry-run`
9. Observe the error `✗ failed to generate kustomization.yaml: failed to save original kustomization.yaml: open C:\<absolute-path>\test\C:\<absolute-path>\test\kustomization.yaml.original: The filename, directory name, or volume label syntax is incorrect. <nil> <nil>`

### Expected behavior

All invocations with `.`, a relative path or an absolute path return the rendered manifest

_No response_

### OS / Distro

Windows 11

### Flux version

v2.7.5

### Flux check

N/A

### Git provider

_No response_

_No response_

### Additional context

_No response_

### Code of Conduct

sibasispadhi:
This commit fixes issue #5673 where 'flux build kustomization' fails on Windows when using --path with relative or absolute paths.

The bug manifested as duplicate path components:
  C:\path\test\C:\path\test\kustomization.yaml.original

Root cause: The path argument was passed directly without normalization, causing issues in downstream path joining.

Fix: Normalize the path using filepath.Abs() and filepath.Clean() before passing it to the builder. This ensures:
  - Relative paths are converted to absolute paths
  - Windows path separators are handled correctly
  - Path components are cleaned and deduplicated

Added unit tests to verify path normalization works correctly on both Linux and Windows platforms.

Fixes #5673

### DIFF
--- a/cmd/flux/build_kustomization_test.go
+++ b/cmd/flux/build_kustomization_test.go
@@ -218,3 +218,65 @@ spec:
 		})
 	}
 }
+
+// TestBuildKustomizationPathNormalization verifies that absolute and complex
+// paths are normalized to prevent path concatenation bugs (issue #5673).
+// Without normalization, paths could be duplicated like: /path/test/path/test/file
+func TestBuildKustomizationPathNormalization(t *testing.T) {
+	// Get absolute path to testdata to test absolute path handling
+	absTestDataPath, err := filepath.Abs("testdata/build-kustomization/podinfo")
+	if err != nil {
+		t.Fatalf("failed to get absolute path: %v", err)
+	}
+
+	tests := []struct {
+		name       string
+		args       string
+		resultFile string
+		assertFunc string
+	}{
+		{
+			name:       "build with absolute path",
+			args:       "build kustomization podinfo --path " + absTestDataPath,
+			resultFile: "./testdata/build-kustomization/podinfo-result.yaml",
+			assertFunc: "assertGoldenTemplateFile",
+		},
+		{
+			name:       "build with complex relative path (parent dir)",
+			args:       "build kustomization podinfo --path ./testdata/build-kustomization/../build-kustomization/podinfo",
+			resultFile: "./testdata/build-kustomization/podinfo-result.yaml",
+			assertFunc: "assertGoldenTemplateFile",
+		},
+		{
+			name:       "build with path containing redundant separators",
+			args:       "build kustomization podinfo --path ./testdata//build-kustomization//podinfo",
+			resultFile: "./testdata/build-kustomization/podinfo-result.yaml",
+			assertFunc: "assertGoldenTemplateFile",
+		},
+	}
+
+	tmpl := map[string]string{
+		"fluxns": allocateNamespace("flux-system"),
+	}
+	setup(t, tmpl)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var assert assertFunc
+
+			switch tt.assertFunc {
+			case "assertGoldenTemplateFile":
+				assert = assertGoldenTemplateFile(tt.resultFile, tmpl)
+			case "assertError":
+				assert = assertError(tt.resultFile)
+			}
+
+			cmd := cmdTestCase{
+				args:   tt.args + " -n " + tmpl["fluxns"],
+				assert: assert,
+			}
+
+			cmd.runTestCmd(t)
+		})
+	}
+}
--- a/cmd/flux/build_kustomization.go
+++ b/cmd/flux/build_kustomization.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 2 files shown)


stefanprodan:
Please run `make fmt` and force push.

cmd/flux/build_kustomization_path_test.go
matheuscscp, line 1: I don't see why create a new file here, we can just use the one we already have

cmd/flux/build_kustomization_path_test.go
```diff
@@ -0,0 +1,105 @@
...
+		{
+			name:    "relative path",
+			input:   "test/path",
+			wantAbs: true,
+			wantErr: false,
+		},
+	}
+
+	// Only run Windows-specific tests on Windows
+	if runtime.GOOS == "windows" {
```
matheuscscp: Not sure how much value these tests have, not sure we have CI running on Windows or if any maintainers use Windows to develop. Maybe equivalent tests that just assert the usage of `filepath.Clean(filepath.Abs())` would be better, regardless of Windows, if possible

sibasispadhi: updated.

cmd/flux/build_kustomization_test.go
```diff
@@ -218,3 +218,48 @@ spec:
...
+
+			// Verify the result is an absolute path
+			if !filepath.IsAbs(cleanPath) {
+				t.Errorf("Expected absolute path, got %q", cleanPath)
+			}
+
+			// Verify Clean doesn't break the path
+			if cleanPath == "" {
+				t.Errorf("filepath.Clean() returned empty path")
+			}
```
matheuscscp: The tests should be testing the Flux code, not the Go standard llibrary `filepath`

sibasispadhi: Got it. Updated. please take a look.

cmd/flux/build_kustomization_test.go
```diff
@@ -218,3 +218,53 @@ spec:
...
+	}{
+		{
+			name:       "build with relative path",
+			args:       "build kustomization podinfo --path testdata/build-kustomization/podinfo",
+			resultFile: "./testdata/build-kustomization/podinfo-result.yaml",
+			assertFunc: "assertGoldenTemplateFile",
+		},
+		{
+			name:       "build with dot-slash relative path",
+			args:       "build kustomization podinfo --path ./testdata/build-kustomization/podinfo",
```
matheuscscp: These paths are too simple and don't test what's changing here. The initial paths from Windows seemed much better in  that regard. Maybe we can adapt them to Linux paths?

P.S.: The vibe-coding is wasting a bit of my time :upside_down_face:

sibasispadhi: Yeah, future is vibe coding. It will get used to soon. Let me recheck & get back.

sibasispadhi: Updated. please take a look when possible. The git should be integrated with a code reviewer agent, thats flush out initial issues before going to actual reviewers. That will save time. I am sure CNCF & other teams would be already thinking about it as that is the the vibe coding future. Have fixed it again, please take a look when possible :) . This time my code-assistant said, you will be happy & sorry :). Lets see !!  :--)

sibasispadhi: Or, are you suggesting to take out the test case all together ?

matheuscscp:
Please squash, this PR should have a single commit

sibasispadhi:
thank you for the review, @matheuscscp .done

matheuscscp (approved):
LGTM

cmd/flux/build_kustomization_test.go
matheuscscp: Using AI effectively and with other people's time in mind is good. Vibe-coding is not.

PR #5682, opened 2026-01-19
Introduce support for looking up GH app installation ID

matheuscscp:
Includes: https://github.com/fluxcd/pkg/pull/1074, https://github.com/fluxcd/pkg/pull/1076


cmd/flux/create_secret_github_app.go
```diff
@@ -70,29 +72,18 @@ func createSecretGitHubAppCmdRun(cmd *cobra.Command, args []string) error {
...
-	}
-
-	if secretGitHubAppArgs.appInstallationID == "" {
-		return fmt.Errorf("--app-installation-id is required")
-	}
-
-	if secretGitHubAppArgs.privateKeyFile == "" {
-		return fmt.Errorf("--app-private-key is required")
-	}
-
```
matheuscscp, line 84: Make it like the `flux-operator` CLI: let the library validate.

PR #5692, opened 2026-01-25
fix: return error immediately on failed reconciliation status

Aman-Cool:
## Flux reconcile does not fail fast on Failed status

`flux reconcile` does not fail fast when a resource is already in `Failed` status.

The polling condition treats `FailedStatus` the same as `InProgressStatus`, so the command waits until the full timeout (default 5 minutes) before reporting an error.

This causes unnecessary delays and poor UX. The fix is to return an error for terminal failure states, matching the behavior of `isObjectReady()`.

### DIFF
--- a/cmd/flux/reconcile.go
+++ b/cmd/flux/reconcile.go
@@ -152,7 +152,14 @@ func reconciliationHandled(kubeClient client.Client, namespacedName types.Namesp
 			return false, err
 		}
 
-		return result.Status == kstatus.CurrentStatus, nil
+		switch result.Status {
+		case kstatus.CurrentStatus:
+			return true, nil
+		case kstatus.InProgressStatus:
+			return false, nil
+		default:
+			return false, fmt.Errorf("%s", result.Message)
+		}
 	}
 }
 


Aman-Cool:
@matheuscscp @stefanprodan ,This makes `flux reconcile` fail fast when a resource is already in `Failed` status, instead of waiting for the full timeout. It aligns the behavior with `isObjectReady()` and improves CLI UX without affecting successful reconciliations.

cmd/flux/reconcile.go
```diff
@@ -152,7 +152,14 @@ func reconciliationHandled(kubeClient client.Client, namespacedName types.Namesp
...
 		}
 
-		return result.Status == kstatus.CurrentStatus, nil
+		switch result.Status {
+		case kstatus.CurrentStatus:
+			return true, nil
+		case kstatus.InProgressStatus:
+			return false, nil
+		default:
+			return false, fmt.Errorf(result.Message)
```
matheuscscp:
```suggestion
			return false, fmt.Errorf("%s", result.Message)
```

Aman-Cool: I’ve updated this to use `fmt.Errorf("%s", result.Message)` to avoid treating the message as a format string. Thanks!

matheuscscp:
Please squash

Aman-Cool:
@matheuscscp , Squashed into a single commit. All checks are green.

Aman-Cool:
@stefanprodan ,Please see it when you have time, happy to help if it needs any updates. Thanks!

stefanprodan:
With the health check cancellation we've shipped in 2.8 this change would make things worse. We should be waiting and not bailout if `Ready=False` but the reason is https://github.com/fluxcd/pkg/blob/357bbcc3462ccf78cc17fd0b9b7d498b1848bf14/apis/meta/conditions.go#L162

PR #5699, opened 2026-01-31
Add example to push artifacts to a registry with a self-signed certif…

nagyv:
…icate

As discussed over Slack, the SSL_CERT_FILE env var can be used to set the path to the certificate, and it will be consumed properly by golang.

### DIFF
--- a/cmd/flux/push_artifact.go
+++ b/cmd/flux/push_artifact.go
@@ -98,6 +98,13 @@ The command can read the credentials from '~/.docker/config.json' but they can a
 	--source="$(git config --get remote.origin.url)" \
 	--revision="$(git tag --points-at HEAD)@sha1:$(git rev-parse HEAD)" \
 	--creds flux:$DOCKER_PAT
+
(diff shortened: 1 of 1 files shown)


stefanprodan:
@nagyv can you please amend the commit message and PR title to "Add SSL_CERT_FILE example to push artifact"

cmd/flux/push_artifact.go
```diff
@@ -98,6 +98,13 @@ The command can read the credentials from '~/.docker/config.json' but they can a
 	--source="$(git config --get remote.origin.url)" \
 	--revision="$(git tag --points-at HEAD)@sha1:$(git rev-parse HEAD)" \
 	--creds flux:$DOCKER_PAT
+
+  # Push manifests to a registry with a self-signed cert
+  export SSL_CERT_FILE=/path/to/your/certificate.crt
```
matheuscscp, line 103: Should be specific that this cert is a CA for server auth

PR #5701, opened 2026-02-01
Fix/resume exit code

Aman-Cool:
### What’s going on
Right now, `flux resume` prints reconciliation errors but still exits with code `0`.  
That makes scripts and CI pipelines think everything worked, even when it didn’t.

### What this changes
If any resource fails to reconcile, the command now exits with a non-zero code.

### Why it matters
Automation relies on exit codes, not logs.  
This makes `flux resume` behave the way people expect and prevents silent failures.

### DIFF
--- a/cmd/flux/reconcile.go
+++ b/cmd/flux/reconcile.go
@@ -152,7 +152,14 @@ func reconciliationHandled(kubeClient client.Client, namespacedName types.Namesp
 			return false, err
 		}
 
-		return result.Status == kstatus.CurrentStatus, nil
+		switch result.Status {
+		case kstatus.CurrentStatus:
+			return true, nil
+		case kstatus.InProgressStatus:
+			return false, nil
+		default:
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@matheuscscp @stefanprodan ,`flux resume` could fail but still exit successfully.  
This makes it return a non-zero exit code on reconciliation failure so automation doesn’t miss it.

Aman-Cool:
Hi @matheuscscp @stefanprodan ,Just bumping this in case it slipped through, all checks are passing.
Let me know if you’d like any changes or clarifications. Thanks!

Aman-Cool:
Squashed into a single commit.
Thanks for the review!
Requesting @stefanprodan for approval when you have a moment.

stefanprodan (approved):
LGTM

matheuscscp:
@Aman-Cool Please rebase and force-push so we can merge :pray:

PR #5702, opened 2026-02-02
[RFC-XXXX] Advanced Support for Short-Lived Cryptographic Material

### LINKED ISSUE #5681: [RFC-XXXX] Advanced Support for Short-Lived Cryptographic Material
In [RFC-0010](https://github.com/fluxcd/flux2/tree/main/rfcs/0010-multi-tenant-workload-identity) we implemented object-level workload identity for cloud providers leveraging ServiceAccount tokens. However, Flux can still improve in terms of workload identity/short-lived-credentials forms of authentication that do not depend on any cloud providers, but rather on open standards (e.g. OIDC, SPIFFE).

A couple of examples:

* The Kustomization and HelmRelease APIs support vendor-agnostic workload identity through the `generic` provider inside `.spec.kubeConfig.configMapRef` -> `.data.provider`. This generic provider issues a ServiceAccount token and uses it for authentication with the remote cluster. The remote cluster must be configured with external OIDC authentication for this to work.
* This [discussion](https://github.com/fluxcd/flux2/discussions/5679) was recently started asking for SPIFFE integration with the Zot container registry. While the specific form with which the request was made does not align well with the Flux multi-tenancy model, it does inspire the idea I'd like to discuss here.

I've been [working](https://github.com/project-zot/zot/issues/3704) with the Zot container registry to introduce OIDC authentication for workloads (Zot already supports OIDC for humans going through the OAuth2 authorization flow in a browser, which is an entirely different thing). With this in mind, I thought about introducing a new authentication pattern for short-lived credentials across Flux, starting with the OCIRepository and ImageRepository APIs.

## Proposal

I'd like to introduce the spec field `.spec.credential`. Valid values would be:

* `ServiceAccountToken`. When the field is set to this value, we do the exactly the same we did for Kustomization and HelmRelease. Users can optionally specify also `.spec.serviceAccountName` in order not to use the ServiceAccount of the respective controller. The open standard here is obviously OIDC. Again, we leverage Kubernetes for this.
* `SpiffeJWT`. When the field is set to this value, we issue a short-lived SPIFFE SVID JWT. Here, (@stealthybox will love it) the identity **will not be based on a ServiceAccount**! Kubernetes can only issue OIDC JWTs for ServiceAccounts, but we have freedom to issue a SPIFFE JWT where the identity is the Flux object **itself**, including kind, name and namespace. The cluster will be identified by the helper flag `--spiffe-issuer`, which will be the `iss` claim of the JWT.
* `SpiffeCertificate`. When the field is set to this value, we issue a short-lived SPIFFE SVID x509 certificate for client authentication via mTLS. Both OCIRepository and ImageRepository already support client authentication via mTLS, so this should be easy to implement.

For the JWT credentials, i.e. `ServiceAccountToken` and `SpiffeJWT`, we can also introduce the field `.spec.audiences`, which will be required for these credentials.

For both `Spiffe*` values of the field, we can add the binary flag `--spiffe-secret-name` for watching a secret containing the private key required for issuing the short-lived credentials, and also the CA in case of certificates.

Note that this proposal introduces a pattern that is significantly different from the configuration API we used for Kustomization and HelmRelease. However, these two APIs have their own very specific circumstances:

* They are the only ones that use a ConfigMap reference.
* They target only Kubernetes clusters, which is a third-party service that already supports the OIDC standard and therefore will not require support for SPIFFE.

### Why a new field and not `.spec.provider`

* The word provider calls the idea of a cloud provider, which is not what we are doing here. Here we are implementing authentication standards, and not targeting cloud vendors. Credentials is a better word for this.
* I want the values for the new field to follow PascalCase, like most enums in Kubernetes (especially because of the `ServiceAccountToken` value). The `.spec.provider` field uses lowercase initials, e.g. `aws`, `gcp`.
* Say we wanted to use `.spec.provider: generic` like in the Kustomization and HelmRelease APIs. This is a strong no-go because this already has a very specific behavior in the OCI APIs.

### How trust should be established for SPIFFE

For JWTs, users must host the issuer and JWKS documents somewhere reachable from the third-party services they are integrating with (e.g. from Zot in this first use case).

For certificates, users must configure the third-party service (e.g. Zot) with the CA provided via `--spiffe-secret-name`.

matheuscscp:
Closes: #5681

Preview: https://github.com/fluxcd/flux2/blob/rfc-creds/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md

### DIFF
--- a/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
+++ b/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
@@ -1,0 +1,144 @@
+# RFC-XXXX Advanced Support for Short-Lived Cryptographic Material
+
+**Status:** provisional
+
+<!--
+Status represents the current state of the RFC.
+Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+-->
+
+**Creation date:** 2026-09-07
+
+**Last update:** 2026-09-07
+
+## Summary
+
+In [RFC-0010](https://github.com/fluxcd/flux2/tree/main/rfcs/0010-multi-tenant-workload-identity)
+we introduced object-level workload identity for cloud provider integrations
+leveraging Kubernetes ServiceAccount tokens. That was our first step towards
+advanced workload identity features. In this RFC we propose a new set of
+closely-related enhancements through a new set of APIs, where the central
+topic is advanced support for short-lived cryptographic material. These
+enhancements cover not only advanced workload identity support, but also
+other scenarios that benefit from short-lived cryptographic material.
+Namely, these scenarios are secure communication between Flux controllers
+and third-party services, and between Flux controllers themselves.
+The main goal of the RFC is to define the shape of these APIs so that we
+can capture all these use cases in a consistent way all across Flux. The
+implementation of the APIs across all the Flux components will be split
+into multiple releases.
+
+## Motivation
+
+Many Flux users have advanced security requiments and use cases. This RFC
+is motivated by the subset of security requirements that can benefit from
+a unified solution for short-lived cryptographic material. They can be
+split into the following categories:
+
+- *Authentication*: Secure authentication with external systems, be it
+  from a cloud provider or a vendor-neutral infrastructure components such
+  as free open-source projects.
+- *Private Communication*: Secure communication among the Flux controllers,
+  and between Flux controllers and external systems.
+- *Centralized Management*: Integration with central infrastructure that
+  is responsible for providing short-lived cryptographic material for the
+  rest of the stack of the Flux user.
+
+The Secure Production Identity Framework For Everyone (SPIFFE) CNCF project
+aims to provide a standardized framework for issuing and managing short-lived
+cryptographic identities for widely used standards such as JWT and x509
+certificates. The project is graduated and adoption grows steadily, from both
+large technology players such as
+[Uber](https://www.uber.com/us/en/blog/solving-the-agent-identity-crisis/)
+and open-source projects, e.g. several service meshes and policy engines.
+Also, Flux users have manifested interest in integrations with SPIFFE from the
+Flux side, such as [#3368 (comment)](https://github.com/fluxcd/flux2/pull/3368#discussion_r1040899292)
+and [#5679](https://github.com/fluxcd/flux2/discussions/5679) and several
+offline discussions in conferences and meetups and lost Slack threads.
+
+This RFC takes RFC-0010 to the next level in two dimensions:
+
+- Flux will be able to support workload identity for more vendor-neutral
+  infrastructure components, such as container registries like Harbor and
+  Zot (both CNCF projects) that have implemented support for workload
+  identity. RFC-0010 introduced workload identity for remote Kubernetes
+  clusters and Flux 2.9 introduced workload identity for OpenBao (and Vault).
+  By supporting workload identity for container registries Flux will be
+  covering workload identity for all the vendor-neutral infrastructure
+  components that are core to Flux: Kubernetes, container registries
+  and key management systems for decryption of secrets / transit engine.
+- Flux will support both an out-of-the-box workload identity backbone,
+  which is Kubernetes itself through ServiceAccount tokens, and a more
+  advanced solution that covers more security scenarios beyond workload
+  identity, like private communication, and aims specifically at centralized
+  management of short-lived cryptographic material. By using the JWT PKI
+  provided by Kubernetes a Flux user can go a long way, but each cluster
+  will have its own key pair and federating several clusters in a consistent
+  way can be a challenge. Because SPIFFE's main goal, on the other hand, is
+  to solve this specific problem, it has come up with established ways for
+  different SPIFFE runtimes to federate and provide a consistent layer of
+  trust across clusters and the entire infrastructure of an organization.
+
+### Goals
+
+The main goal of this RFC is defining the shape of the APIs that will support
+the use cases described in this RFC in a way that they can be uniformly
+implemented across all the Flux components (given enough time and demand).
+
+### Non-Goals
+
+It's not a goal of this RFC to make Flux become a SPIFFE runtime,
+like SPIRE (the reference runtime implementation). The whole point of
+SPIFFE is providing a simple interface through which applications can
+acquire short-lived cryptographic material from the infrastructure,
+allowing the ownership of the signing keys to remain with the
+infrastructure rather than with the application. From this RFC's
+perspective, Flux is the application, and whatever SPIFFE runtime
+the user has deployed is the infrastructure. There are both free
+(SPIRE) and commercial options available.
+
+## Proposal
+
+TODO
+
+### API fields
+
+TODO
+
+### Controller Flags
+
+TODO
+
+### CLI Support
+
+TODO
+
+### User Stories
+
+TODO
+
+## Design Details
+
+TODO
+
+### Validation Rules for the API fields
+
+TODO
+
+### Caching
+
+The existing token cache from RFC-0010 will be reused. Following
+the same principle established there, any new fields that affect
+the credential will be part of the key computed during a
+reconciliation to look up the credential cache.
+
+## Implementation History
+
+TODO
+
+<!--
+Major milestones in the lifecycle of the RFC such as:
+- The first Flux release where an initial version of the RFC was available.
+- The version of Flux where the RFC graduated to general availability.
+- The version of Flux where the RFC was retired or superseded.
+-->
--- a/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
+++ b/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
@@ -523,14 +523,23 @@ to authenticate incoming connections from Ingress and Gateway controllers:
 --authorized-receiver-spiffe-id=spiffe://<trust domain>/my-gateway-controller
 ```
 
+Finally, all the controllers can authorize the Kubernetes API Server
+if it serves an X509-SVID in the same trust domain:
+
+```sh
+--kube-apiserver-spiffe-id=spiffe://<trust domain>/kube-apiserver
+```
+
 This covers the entire communication matrix between the Flux controllers.
 Every peer can authorize the peer on the other side of the TCP connection,
-for all types of HTTP communications that exist inside Flux.
+for all types of HTTP traffic existing inside Flux.
 
-The presence of at least one occurrence of these flags gates the feature
-in the controller. Note that artifact traffic, event traffic and receiver
-traffic can all be enabled separately, since each traffic type has its own
-flags.
+The presence of at least one occurrence of these flags gates the respective
+feature in the respective controller. Note that not all types of traffic
+need to be enabled together. The flags allow protecting each type of traffic
+independently, e.g. artifact traffic can be protected while Kubernetes
+API Server traffic can remain using the default Kubernetes CA mounted into
+pods.
 
 ### New Dependencies and Bootstrap
 
--- a/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
+++ b/rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
@@ -9,7 +9,7 @@ Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejec
 
 **Creation date:** 2026-09-07
 
-**Last update:** 2026-09-07
+**Last update:** 2026-09-10
 
 ## Summary
 
@@ -35,9 +35,9 @@ is motivated by the subset of security requirements that can benefit from
 a unified solution for short-lived cryptographic material. They can be
 split into the following categories:
 
-- *Authentication*: Secure authentication with external systems, be it
-  from a cloud provider or a vendor-neutral infrastructure components such
-  as free open-source projects.
+- *Identification*: Secure identification with external systems, such as
+  cloud providers or vendor-neutral infrastructure components such as
+  free open-source projects.
 - *Private Communication*: Secure communication among the Flux controllers,
   and between Flux controllers and external systems.
 - *Centralized Management*: Integration with central infrastructure that
@@ -47,14 +47,22 @@ split into the following categories:
 The Secure Production Identity Framework For Everyone (SPIFFE) CNCF project
 aims to provide a standardized framework for issuing and managing short-lived
 cryptographic identities for widely used standards such as JWT and x509
-certificates. The project is graduated and adoption grows steadily, from both
-large technology players such as
+certificates. The standard name defined by SPIFFE for the short-lived
+cryptographic material it provides is SPIFFE Verifiable Identity Document
+(SVID). The standard defines both JWT-SVID and X509-SVID. Recently, WIT-SVID
+was also introduced, but it still lacks real-world adoption due to being so
+new, so we choose to leave it out of the scope of this RFC. All the standards
+referenced in this RFC are defined
+[here](https://github.com/spiffe/spiffe/tree/main/standards).
+
+The SPIFFE project is graduated and adoption grows steadily,
+from both large technology players such as
 [Uber](https://www.uber.com/us/en/blog/solving-the-agent-identity-crisis/)
 and open-source projects, e.g. several service meshes and policy engines.
 Also, Flux users have manifested interest in integrations with SPIFFE from the
 Flux side, such as [#3368 (comment)](https://github.com/fluxcd/flux2/pull/3368#discussion_r1040899292)
 and [#5679](https://github.com/fluxcd/flux2/discussions/5679) and several
-offline discussions in conferences and meetups and lost Slack threads.
+offline discussions in conferences and meetups, and in lost Slack threads.
 
 This RFC takes RFC-0010 to the next level in two dimensions:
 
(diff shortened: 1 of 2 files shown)


rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,632 @@
...
+  flag `--spiffe-secret-name`. Both `OCIRepository` and `ImageRepository`
+  already support client certificate authentication via mTLS, so this
+  integrates naturally with the existing transport layer. The third-party
+  service must be configured with the CA certificate for trust. Note that
+  `.spec.certSecretRef` may still be optionally used alongside
+  `SpiffeCertificate` for specifying a CA to trust the server's certificate.
+
+### Additional Fields
+
+- **`.spec.audiences`** (list of strings): Optional. Specifies the audiences
```
stefanprodan: Since the `audiences` is a property of the credential, maybe grouping them would be better?

```yaml
credential:
   type: ServiceAccountToken
   audiences:
      - https://registry1.exmaple.com
      - https://registry2.exmaple.com
```

matheuscscp: That's a great idea. In Kustomization and HelmRelease everything is on the same level, but they can be considered exceptions... For them it's a ConfigMap after all. What do you think?

matheuscscp: There's also the fact that spec.serviceAccountName is also a property of the credential...

stefanprodan: Since this is about Flux sources only, I would go with grouping. It's a better UX and allows us to extend this feature with other options without polluting the main spec.

stefanprodan: Also, if we decide to have in the notification API, a shared struct in fluxcd/pkg/apis would fit better than lots of fields in spec.

matheuscscp: > Since this is about Flux sources only

At the moment it is. But as mentioned in the Goals, this RFC is here to establish a pattern to be followed everywhere, like most security-related fields are in Flux.

> Also, if we decide to have in the notification API, a shared struct in fluxcd/pkg/apis would fit better than lots of fields in spec.

Indeed :+1: 

In any case, `.spec.serviceAccountName` has multiple usages and it's not just `.spec.credential`, so it makes sense for it to be a top-level field. This won't be the case for `.spec.audiences` so I think the struct approach is the best :+1:

matheuscscp: Fixed, thanks!

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,658 @@
...
+[here](#why-a-new-field-and-not-specprovider)). The mutual exclusivity is enforced
+via a CEL validation rule:
+
+```yaml
+x-kubernetes-validations:
+- message: spec.credential can only be used with spec.provider 'generic'
+  rule: '!has(self.credential) || !has(self.provider) || self.provider == ''generic'''
+```
+
+### Credential Types
```
hiddeco: The flat struct with a type discriminator doesn't follow Kubernetes API conventions and I think it will cause issues as the feature evolves.

For example, `audiences` applies to `ServiceAccountToken` and `SpiffeJWT` but not `SpiffeCertificate`. This requires "field X only valid when type is Y" validation that grows with each type-specific field added later.

I worry that when `SpiffeJWT` needs custom claims or `SpiffeCertificate` needs a key algorithm preference, they end up in the same flat struct with conditional applicability.

If we would make the presence of a sub-field the discriminator, the type field would become redundant. This discriminated union pattern can also be found elsewhere in Kubernetes (e.g. `VolumeSource`, `PersistentVolumeSource`, `IngressBackend`):

```go
type Credential struct {
    ServiceAccountToken *ServiceAccountTokenCredential `json:"serviceAccountToken,omitempty"`
    SpiffeJWT           *SpiffeJWTCredential           `json:"spiffeJWT,omitempty"`
    SpiffeCertificate   *SpiffeCertificateCredential   `json:"spiffeCertificate,omitempty"`
}

type ServiceAccountTokenCredential struct {
    Audiences []string `json:"audiences,omitempty"`
}

type SpiffeJWTCredential struct {
    Audiences []string `json:"audiences,omitempty"`
}

type SpiffeCertificateCredential struct {}
```

Mutual exclusivity via CEL:

```yaml
x-kubernetes-validations:
- message: exactly one credential type must be specified
  rule: '(has(self.serviceAccountToken) ? 1 : 0) + (has(self.spiffeJWT) ? 1 : 0) + (has(self.spiffeCertificate) ? 1 : 0) == 1'
```

This is slightly more verbose, but would long-term (and with possible future extensions), not cause you to be stuck in a corner.

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,658 @@
...
+- It's not a goal to support all Flux APIs in the first iteration. The initial
+  implementation targets `OCIRepository` and `ImageRepository`, with other APIs
+  to follow in subsequent releases.
+
+## Proposal
+
+We propose introducing a new spec field `.spec.credential` to the `OCIRepository`
+and `ImageRepository` APIs. This field specifies the type of vendor-agnostic
+short-lived credential to use for authentication. The field is mutually exclusive
+with `.spec.provider` (when set to a cloud provider) because "provider" conveys
```
hiddeco: Should we provide another CEL rule example for this? Given "mutual exclusivity".

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,658 @@
...
+spec:
+  url: oci://registry.example.com/secure-app
+  credential:
+    type: SpiffeCertificate
+```
+
+The controller is started with `--spiffe-issuer=https://spiffe.example.com`.
+It issues a short-lived x509 certificate with the SPIFFE ID
+`spiffe://spiffe.example.com/ocirepositories/production/secure-app`
+in the SAN URI field, and uses it for mTLS authentication with the registry.
```
hiddeco: Elsewhere we state:

>  When `.spec.credential.type` is `SpiffeCertificate`, the controller flags   `--spiffe-trust-domain` and `--spiffe-secret-name` must be set, otherwise a terminal error is returned.

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,658 @@
...
+  url: oci://registry.example.com/my-app
+  credential:
+    type: SpiffeJWT
+    audiences:
+      - registry.example.com
+```
+
+The controller is started with `--spiffe-issuer=https://spiffe.example.com`.
+The SPIFFE ID for this object would be
+`spiffe://spiffe.example.com/ocirepositories/production/my-app`,
```
hiddeco: Isn't the SPIFFE ID derived from the trust domain (and not the issuer URL)?

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
stefanprodan: We use `type: x` in all the Flux APIs for sources and notification. While I agree that the proposed structure would be better, doesn't fit with how the rest of the API looks. We can write CEL functions and reject field combinations that we don't support.

matheuscscp: Agreed. I think this suggestion would be great for a new project without established patterns, but I think in Flux we have already done a few things according to the currently proposed structure as Stefan said

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: Not sure I follow the suggestion :thinking:

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: Good catch!

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: At the end, my proposal is to have one flag for the issuer, and one for the trust domain, because the SPIFFE spec does not establish any relationship between them. I believe the recommended way would be having the issuer host be used as the trust domain, but the spec does not enforce that. So for allowing flexibility upfront (and not considering breaking changes in the future), I think the best is to provider a flag for each and use them as is.

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: Fixed, thanks!

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: Fixed, thanks!

matheuscscp:
@hiddeco I fixed the inconsistencies you found around the SPIFFE flags in this diff: https://github.com/fluxcd/flux2/compare/f9632fd9d7c118763ba7bfbd2e88cbe8a9dbf8e3..792631f49699aa8813cc24f2875f3a75fe1a826c

They were an artifact of my late decision to introduce a separate flag for the trust domain.

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: I think the current CEL expression already covers the entire mutual exclusivity, no? :thinking:

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,660 @@
...
+  [SPIFFE ID](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md)
+  in the format
+  `spiffe://<trust-domain>/<resource>/<namespace>/<name>`, where `<resource>`
+  is the lowercase plural form of the Kubernetes resource type, i.e.
+  `ocirepositories` for `OCIRepository` and `imagerepositories` for
+  `ImageRepository`.
+  The trust domain for the SPIFFE ID is configured via the controller flag
+  `--spiffe-trust-domain`. The `iss` claim is set to the controller flag
+  `--spiffe-issuer`, which third-party services use for OIDC discovery. The
+  JWT is signed using a private key provided via `--spiffe-secret-name`. Unlike
```
stealthybox: do we just read this secret every time and let the client decide how long to cache it?

matheuscscp: We don't cache secrets in Flux, it's always a live lookup

matheuscscp:
@stefanprodan @hiddeco @stealthybox The last week to work on `ServiceAccountToken` for Flux 2.8 is the next one, let's try to approve this RFC ASAP :pray:

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
```diff
@@ -0,0 +1,660 @@
...
+  `ServiceAccountToken`, this credential type does **not** depend on a
+  Kubernetes `ServiceAccount` for identity. The third-party service must be
+  configured to trust the SPIFFE issuer by having access to the corresponding
+  JWKS document.
+
+- **`SpiffeCertificate`**: The controller issues a short-lived SPIFFE SVID
+  x509 certificate for client authentication via mTLS. The certificate encodes
+  the same SPIFFE ID as `SpiffeJWT` in the SAN URI field. The certificate is
+  signed using a CA private key and certificate provided via the controller
```
SpectralHiss: I don't think having to supply CA key as secret will be acceptable to many security teams to be honest..

stefanprodan: All Flux APIs require the CA be specified in a secret.

SpectralHiss: To be clear, I mean the private key of the CA which as I understand is required for issuance for this proposal to work, not the CA (public) certificate (aka root of trust).

matheuscscp: @SpectralHiss As we discussed in our call today, this proposal indeed needs to be clearer and to better frame why we are proposing to provide a CA key to Flux: The idea is replacing the PKI from the Kubernetes ServiceAccount token issuer with one dedicated for Flux. There are two major advantages:

* This CA would be dedicated for Flux. Only Flux can/should issue short-lived creds with it. This is better than Kube's SA token issuer because there are much less components issuing creds with it: it's just Flux, and not Kube itself (for its own built-in purposes) plus anything else that has the `create` RBAC verb for the `serviceaccounts/token` resource in the cluster.
* Flux can issue identities mapping 1:1 with objects, the identity is the object itself. With Kube SA token issuer we're limited to SAs, which can be reused by different objects.

The justification for this is that Flux is normally the most trusted component in the cluster, deploying absolutely everything inside it, including CNI (e.g. https://timoni.sh/flux-aio/).

vicaya: > The justification for this is that Flux is normally the most trusted component in the cluster, deploying absolutely everything inside it, including CNI

Unless Flux can be easily run on isolated control-plane nodes for components like kube-apiserver and etcd, any long-lived secrets stored anywhere on worker nodes should be assumed compromised by one of those Linux kernel zero-days. Even for on-prem clusters, it's possible for kube-apiserver to use --service-account-signing-endpoint to delegate signing to HSM/KMS. Until Flux has an equivalent external signing mechanism, the SPIFFE support should be considered experimental only.

I suggest split the PR and ship common machinery and KSA first and let SPIFFE bake for a while.

matheuscscp: @vicaya This thread is ultra stale. The plan is no longer making Flux issue the crypto material itself using private keys, but rather rely on the SPIFFE Broker API:

* SPIFFE: https://github.com/spiffe/spiffe/pull/340
* SPIRE: https://github.com/spiffe/spire/pull/6915

Flux will connect to the SPIFFE Workload API via UDS by mounting the SPIFFE CSI Driver, request an x509-SVID for itself, then use this x509-SVID bundle to connect over TCP to the SPIFFE Broker API. It will then use Broker API RPCs to request SVIDs using a KubernetesObjectReference that points exactly to the object Flux is reconciling and let the SPIFFE server (SPIRE or competitors) deal with attesting and issuing the crypto material.

This solution is fully secret-less and is the primary way for SPIFFE to cover use cases like Flux. I took over the implementation of the Broker API in SPIRE and am constantly in touch with the SPIFFE/SPIRE maintainers, Flux is now a primary use case of the Broker API along Istio.

I'm going to resolve this thread.

rfcs/xxxx-vendor-agnostic-short-lived-credentials/README.md
matheuscscp: Stale, we no longer need secrets for the SPIFFE integration as per https://github.com/fluxcd/flux2/pull/5702#discussion_r3366902649.

matheuscscp:
As explained here: https://github.com/fluxcd/flux2/pull/5702#discussion_r3366902649

The SPIFFE integration currently described in this proposal is stale. This RFC will soon undergo lots of changes.

Also, this RFC is delayed to Flux 2.10 (planned for somewhere in Q4), and it will be the main topic of the release. The Flux 2.9 release is unfortunately coming very soon: we will start releasing the controllers on June 15. Flux 2.9 is also very stuffed and even if we had more time we decided not to ship more things, otherwise the time we planned for dealing with fallouts and bugs after the release will not be enough and we will compromise other things in our calendars.

cc @vicaya

matheuscscp:
SPIFFE merged the Broker API today!

https://github.com/spiffe/spiffe/pull/340

matheuscscp:
@stefanprodan @stealthybox @hiddeco This is now fully ready for review.

@SpectralHiss Appreciate another round!

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
```diff
@@ -0,0 +1,677 @@
...
+
+Some options apply to both, some apply only to one of the two categories mentioned above.
+
+#### The `go-spiffe` Environment Variables
+
+The first set of controller options that will be added are the environment
+variables used by the `go-spiffe` library:
+
+```sh
+SPIFFE_ENDPOINT_SOCKET=unix:///spiffe-workload-api/spire-agent.sock
```
stefanprodan, line 385: Can you please explain what this entails and what options we have for not breaking the Restricted Pod Security Standard which Flux subscribes to.

matheuscscp: Done, thanks!

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
```diff
@@ -0,0 +1,783 @@
...
+
+```sh
+--authorized-receiver-spiffe-id=spiffe://<trust domain>/my-ingress-controller
+--authorized-receiver-spiffe-id=spiffe://<trust domain>/my-gateway-controller
+```
+
+Finally, all the controllers can authorize the Kubernetes API Server
+if it serves an X509-SVID in the same trust domain:
+
+```sh
```
SpectralHiss, line 542: I think I have not seen this that much, if at all, in the wild. Using spiffe for the kubernetes apiserver certs is quite tricky. might not be worth including as it might complicate an already complex picture

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
```diff
@@ -0,0 +1,783 @@
...
+networking between the Flux controller pods). The only supported
+bootstrap method that already handles both CNI and CSI dependencies
+is the
+[`flux-operator-bootstrap`](https://github.com/controlplaneio-fluxcd/terraform-kubernetes-flux-operator-bootstrap/blob/main/scripts/e2e-critical-components.sh)
+Terraform module.
+
+#### SPIFFE Broker API
+
+Only the object-level features will require access to the SPIFFE Broker API.
+This means inter-controller private communication does not require access to
```
SpectralHiss, line 602: This will also need strict authz , the approach above with the flags + spiffe mTLS should be sufficient but might be good to clarify

SpectralHiss (commented):
Small suggestions, nice work on this Matheus!

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
matheuscscp: Great insight! I will reword the text to communicate this, and keep the implementation as a non-prio in the backlog until users ask for this / k8s support improves. I will however state that the flags / API fields in this RFC are still the ones to be implemented / supported when it happens 🙏

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
matheuscscp: Cool, I will update the `#### Inter-Controller Communication` section to emphasize that the flags are for implementing strict authz of the controller SPIFFE IDs by all the peers 👍

matheuscscp: Hmm actually I already put something about this in `#### Inter-Controller Communication`:

> However, in addition to acquiring its own X509-SVID, the controller
also needs to authorize SPIFFE IDs of the other controllers. This
warrants a new set of flags in specific controllers.

Did I misunderstand the comment?

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
matheuscscp: Addressed here: https://github.com/fluxcd/flux2/pull/5702/commits/e0c9a1a1068020b5aec61a200547f73d86f7884f

Thanks!

rfcs/xxxx-advanced-support-for-short-lived-cryptographic-material/README.md
SpectralHiss: yes the mechanism to do the authz between controller instances is very well explained in that section, but what not clear to me is which component will host the Broker api tcp server and which controllers it needs to allow for object level wi to work

SpectralHiss: I probably misunderstood how the brokerAPI will be hosted tbh :)

matheuscscp: SPIFFE does not define how authorization for Broker workloads to fetch SVIDs for objects is achieved, SPIFFE leaves this as an implementation concern for the SPIFFE runtime. SPIRE, for example, supports doing this via Kubernetes RBAC. SPIRE defined a special RBAC verb `impersonate-via-spire` for allowing SPIFFE IDs of Broker workloads (used as the `username` field in Kubernetes RBAC) to be authorized for KubernetesObjectReferences.

Thanks for the clarification! I will push a commit with this.

matheuscscp: Addressed here: https://github.com/fluxcd/flux2/pull/5702/commits/2d31b1a2ea3257d41f627399c50218779025c397

I also emphasized that Broker API is part of the SPIFFE runtime, just like Workload API. This means that the way it gets deployed in your environment is out of scope here, as we aim to cleanly integrate with SPIFFE without any particular assumptions of any particular SPIFFE runtimes.

Thanks!

Issue #5683, opened 2026-01-19
`flux events --all-namespaces` is hanging


towolf:
### Describe the bug

With flux v2.7.5 I get a hanging command `flux events --all-namespaces` with no output, also no more output with `--verbose`.

If I go through all namespaces with `--namespace` I get results.

If I use `kubectl events --all-namepace` I get all results.

### Steps to reproduce

With flux v2.7.5 use

```
flux events --all-namespaces
```

### Expected behavior

List of events returned

### OS / Distro

Linux

### Flux version

2.7.5

### Flux check

$ flux check
► checking prerequisites
✔ Kubernetes 1.33.5-gke.2118000 >=1.32.0-0
► checking version in cluster
✔ distribution: flux-2.7.5
✔ bootstrapped: false
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v1.4.5
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v1.7.3
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v1.7.5
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v1.7.4
✔ source-watcher: deployment ready
► ghcr.io/fluxcd/source-watcher:v2.0.3
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta3
✔ artifactgenerators.source.extensions.fluxcd.io/v1beta1
✔ buckets.source.toolkit.fluxcd.io/v1
✔ externalartifacts.source.toolkit.fluxcd.io/v1
✔ gitrepositories.source.toolkit.fluxcd.io/v1
✔ helmcharts.source.toolkit.fluxcd.io/v1
✔ helmreleases.helm.toolkit.fluxcd.io/v2
✔ helmrepositories.source.toolkit.fluxcd.io/v1
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1
✔ ocirepositories.source.toolkit.fluxcd.io/v1
✔ providers.notification.toolkit.fluxcd.io/v1beta3
✔ receivers.notification.toolkit.fluxcd.io/v1
✔ all checks passed

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

matheuscscp:
Works for me

towolf:
I have three colleagues, two of them can repro, one of them can't (he is on flux 2.6.x).

[...]

PR #5721, opened 2026-02-18
Fix event listing ignoring pagination token

matheuscscp:
Fixes: #5683

We were ignoring the pagination token, so namespaces with more events than the page size were getting stuck in infinite loop.

The test I added fails without the fix.


stefanprodan (approved):
LGTM

PR #5724, opened 2026-02-19
Add --resolve-symlinks flag to build and push artifact commands

### LINKED ISSUE #5055: `flux build artifact` doesn't follow symlinks
### Describe the bug

When using `flux build artifact --path=src --output=result.tgz`, Flux seems to generate an empty OCI artifact if `src` contains symlinks to YAML manifest outside of `src`.

The use case is using Flux to generate OCI artifacts from a symlink tree built by Nix, with YAML files pointing to the read-only Nix store.

### Steps to reproduce

```sh
mkdir ro rw
echo >ro/namespace.yaml 'apiVersion: v1
kind: Namespace
metadata:
  name = test'
chmod -w -R ro
ln -s ../ro/namespace.yaml rw/namespace.yaml
flux build artifact --path=rw --output=rw.tgz
tar tvf rw.tgz
```

### Expected behavior

Expected output:

```
drwxr-xr-x 0/0               0 1970-01-01 01:00 .
-rw-r--r-- 0/0              56 1970-01-01 01:00 namespace.yaml
```

Actual output:

```
drwxr-xr-x 0/0               0 1970-01-01 01:00 .
```

Ideally Flux should resolve symlinks when creating OCI artifacts. Additionally, it should use a temporary directory (or allow specifying one) for any temporary files, which would allow working with source files located in the Nix store.

_No response_

### OS / Distro

NixOS 24.05

### Flux version

v2.4.0

### Flux check

N/A

### Git provider

N/A

GitHub (ghcr.io), but irrelevant here

### Additional context

_No response_

### Code of Conduct

rohansood10:
When building OCI artifacts from directories containing symlinks (e.g., symlink trees created by Nix), symlinked files are silently skipped. This adds a --resolve-symlinks flag to both build artifact and push artifact commands that resolves symlinks by copying their targets into a temp directory before archiving. Implements the approach suggested by stefanprodan in the issue discussion. Includes unit tests. Fixes #5055

### DIFF
--- a/cmd/flux/build_artifact.go
+++ b/cmd/flux/build_artifact.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -48,9 +49,10 @@ from the given directory or a single manifest file.`,
 }
 
 type buildArtifactFlags struct {
-	output      string
-	path        string
-	ignorePaths []string
+	output          string
+	path            string
+	ignorePaths     []string
+	resolveSymlinks bool
 }
 
 var excludeOCI = append(strings.Split(sourceignore.ExcludeVCS, ","), strings.Split(sourceignore.ExcludeExt, ",")...)
@@ -61,6 +63,7 @@ func init() {
 	buildArtifactCmd.Flags().StringVarP(&buildArtifactArgs.path, "path", "p", "", "Path to the directory where the Kubernetes manifests are located.")
 	buildArtifactCmd.Flags().StringVarP(&buildArtifactArgs.output, "output", "o", "artifact.tgz", "Path to where the artifact tgz file should be written.")
 	buildArtifactCmd.Flags().StringSliceVar(&buildArtifactArgs.ignorePaths, "ignore-paths", excludeOCI, "set paths to ignore in .gitignore format")
+	buildArtifactCmd.Flags().BoolVar(&buildArtifactArgs.resolveSymlinks, "resolve-symlinks", false, "resolve symlinks by copying their targets into the artifact")
 
 	buildCmd.AddCommand(buildArtifactCmd)
 }
@@ -85,6 +88,15 @@ func buildArtifactCmdRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("invalid path '%s', must point to an existing directory or file", path)
 	}
 
+	if buildArtifactArgs.resolveSymlinks {
+		resolved, cleanupDir, err := resolveSymlinks(path)
+		if err != nil {
+			return fmt.Errorf("resolving symlinks failed: %w", err)
+		}
+		defer os.RemoveAll(cleanupDir)
+		path = resolved
+	}
+
 	logger.Actionf("building artifact from %s", path)
 
 	ociClient := oci.NewClient(oci.DefaultOptions())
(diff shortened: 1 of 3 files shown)


rohansood10:
Addressed the review feedback in the latest push:

1. **Symlinked directory traversal**: Replaced `filepath.Walk` with a recursive `copyDir` function using `os.ReadDir` that properly follows symlinks into directories. `filepath.Walk` uses `Lstat` internally and doesn't descend into symlinked directories — the new approach resolves symlinks and recursively copies directory contents.

2. **File permission preservation**: Updated `copyFile` to use `os.OpenFile` with the source file's mode instead of `os.Create` (which defaults to 0666), ensuring permissions are preserved.

3. **Test coverage**: Added assertions to verify that files inside symlinked directories are properly copied and are regular files (not symlinks).

rohansood10:
Friendly ping — any chance to review this when you get a moment? Happy to make changes if needed.

cmd/flux/build_artifact.go
```diff
@@ -96,6 +108,125 @@ func buildArtifactCmdRun(cmd *cobra.Command, args []string) error {
...
+		return "", err
+	}
+
+	return tmpDir, nil
+}
+
+// copyDir recursively copies the contents of srcDir to dstDir, resolving any
+// symlinks encountered along the way. This handles symlinked directories that
+// filepath.Walk would not descend into (since Walk uses Lstat).
+func copyDir(srcDir, dstDir string) error {
```
stefanprodan: `copyDir` follows symlinks recursively with no cycle detection. If a symlink creates a circular reference (e.g., `link -> ../parent`), `filepath.EvalSymlinks` resolves the intermediate symlink but the resolved real directory can contain further symlinks pointing back up, causing infinite recursion and a stack overflow.

This is particularly relevant for the Nix use case, where symlink trees can be complex. A `visited` map tracking resolved real paths (by device+inode or absolute path) should be threaded through the recursion:

```go
func copyDir(srcDir, dstDir string, visited map[string]bool) error {
    real, _ := filepath.EvalSymlinks(srcDir)
    abs, _ := filepath.Abs(real)
    if visited[abs] {
        return nil // break the cycle
    }
    visited[abs] = true
    // ... rest of the function
}
```

cmd/flux/build_artifact.go
```diff
@@ -96,6 +108,125 @@ func buildArtifactCmdRun(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// resolveSymlinks creates a temporary directory with symlinks resolved to their
+// real file contents. This allows building artifacts from symlink trees (e.g.,
+// those created by Nix) where the actual files live outside the source directory.
+func resolveSymlinks(srcPath string) (string, error) {
```
stefanprodan: When `resolveSymlinks` receives a single file path, it returns a temp **directory** containing the copied file. The caller then sets `path = resolved`, so `oci.Build` now receives a directory instead of a file. While the archive content ends up functionally the same (both cases produce just the base filename in the tarball), the semantics change — and it could diverge if `oci.Build` behavior is ever updated.

A simpler and more correct approach for the single-file case would be to return the path to the resolved file itself (within the temp dir), not the temp dir:

```go
// Instead of: return tmpDir, nil
return dst, nil
```

rohansood10:
Hi @stefanprodan @hiddeco,

Friendly ping on this PR! It's been 8 days since the last update. 

This PR adds the `--resolve-symlinks` flag to the `flux build artifact` and `flux push artifact` commands, which enables proper handling of symbolic links when creating OCI artifacts. This is particularly useful for:

- Kustomization overlays that reference base configurations via symlinks
- Shared configuration patterns where multiple apps reference common configs
- GitOps workflows that use symlinks to avoid duplication

The implementation:
- Adds the flag to both build and push commands
- Properly resolves symlinks during artifact creation
- Includes comprehensive tests
- Updates documentation

All CI checks are passing. Would appreciate a review when you get a chance. Happy to address any feedback!

Thanks!

stefanprodan:
@rohansood10 did you read my last review??

rohansood10:
Hi @stefanprodan, yes I did! I addressed all your review feedback in the commits from 8 days ago:

1. ✅ Fixed symlinked directory traversal by replacing filepath.Walk with a recursive copyDir function
2. ✅ Fixed file permission preservation using os.OpenFile with source file mode
3. ✅ Added test coverage for files inside symlinked directories

The implementation now properly handles all the edge cases you mentioned. All tests are passing. Is there any additional feedback I should address?

stefanprodan:
You really don't see the last review I made around symlinks recursively with no cycle detection and the single file path?

rohansood10:
@stefanprodan Apologies for missing those — I see the two review comments now. Both are addressed in the latest push:

1. **Symlink cycle detection** ([comment](https://github.com/fluxcd/flux2/pull/5724#discussion_r2868659162)): Added a `visited` map (keyed by absolute resolved path) threaded through `copyDir` recursion. If a directory has already been visited, we return early to break the cycle. Added `Test_resolveSymlinks_cycle` that creates a `dir/cycle -> dir` symlink and verifies no infinite recursion or deeply nested paths.

2. **Single file path semantics** ([comment](https://github.com/fluxcd/flux2/pull/5724#discussion_r2868659528)): `resolveSymlinks` now returns the resolved file path (not the temp directory) for the single-file case, preserving file semantics for callers. The function signature was updated to return a separate cleanup path so the temp directory is still properly cleaned up. Added `Test_resolveSymlinks_singleFile` to verify the returned path is a file.

stefanprodan:
@rohansood10 please squash all commits into one, signoff and force push. You also need to rebase your fork.

rohansood10:
@stefanprodan done - squashed into a single commit, signed off, and rebased onto main. Ready for another look when you get a chance!

stefanprodan:
you need to sync your fork with upstream and rebase! take a look at what you pushed in this PR...

rohansood10:
@stefanprodan done - synced the fork and rebased with a single clean commit. Sorry about the messy state earlier.

rohansood10:
fork is synced and rebased with a clean single commit now - ready for another look when you get a chance

Issue #5510, opened 2025-09-09
[RFC-0011] Extend the tracing logic to advanced use cases


adri1197:
Part of: https://github.com/fluxcd/flux2/pull/5321

Recently was introduced a new provider type, otel, in the notification-controller. The initial focus was on ensuring support and performing thorough testing across three main use cases:
- `<source>` -> `Kustomization` -> KRM (Kubernetes resource manifests).
- `OCIRepository` -> multiple `HelmReleases`.
- `HelmRepository` -> `HelmChart` (treated as the actual source) -> `HelmRelease`.

That said, there’s another use case that seems quite common: `<source>` -> `Kustomization` -> `<any-resource>` (including FluxCD’s own resources). Right now, this scenario can lead to separate traces being generated per source applied, depending on what the Kustomization ends up reconciling.

The assumption is that there will always be a source and a Kustomization in the chain. The Kustomization acts as the orchestrator: it decides what gets applied and already carries the top-level source revision as part of its metadata. This opens the door for a neat solution: by appending a `commonAnnotation` with the source revision (using the event convention `event.toolkit.fluxcd.io/`), that annotation could automatically propagate to all applied resources. In turn, notification-controller could then pick this up when building spans and linking traces across resources.

I’ve sketched a tentative implementation for this across the following controllers:
- kustomize-controller - [internal/controller/kustomization_controller.go](https://github.com/adri1197/kustomize-controller/blob/otel-source/internal/controller/kustomization_controller.go#L806-L824).
- notification-controller - [internal/notifier/otel.go](https://github.com/adri1197/notification-controller/blob/otel-source/internal/notifier/otel.go#L239-L245).

Here's how it may look like with the following scenario a single trace (under the same rootSpan/traceID): `GitRepository` -> `Kustomization` -> `OCIRepository` > collection of `HelmRelease`.

<img width="975" height="399" alt="Image" src="https://github.com/user-attachments/assets/63dbc446-d14e-4e01-b02b-870a375a27e8" />

PS: I'd rather discussing this in FluxCD Dev weekly meeting, however I'm not going to be able to join in the upcoming one (Sept 10th).

PR #5753, opened 2026-03-05
Update RFC-0011: Include parent/child orchestration logic

adri1197:
Update RFC-0011 to include implementation history.

Part of: #5510

closes fluxcd/notification-controller#1172


stefanprodan:
I think this is already covered in main.

adri1197:
> I think this is already covered in main.

Oh.. I didn't spot got updated. I would like to include the parent/child orchestration convention and design as part of the RFC, once it's merged. Do you think it's a good to have?

stefanprodan:
> I would like to include the parent/child orchestration convention and design as part of the RFC

Yes 💯

PR #5759, opened 2026-03-07
fix(reconcile): fix HelmChart namespace lookup when HelmRepository is in a different namespace

Aman-Cool:
fix: resolve HelmChart lookup using wrong namespace with --with-source

When a HelmRelease uses spec.chart (HelmChartTemplate) and the HelmRepository lives in a different namespace (e.g. flux-system), `flux reconcile hr <name> --with-source` was failing with a not found error.

The bug was in reconcile_helmrelease.go — the HelmChart lookup was using the HelmRepository's namespace instead of the HelmRelease's namespace. Since helm-controller always creates the HelmChart in the HelmRelease's namespace, the lookup returned a 404.

Fixed by using obj.Namespace for the lookup. The HelmRepository namespace is correctly handled downstream by helmChartAdapter.getSource().

### DIFF
--- a/cmd/flux/reconcile_helmrelease.go
+++ b/cmd/flux/reconcile_helmrelease.go
@@ -98,11 +98,9 @@ func (obj helmReleaseAdapter) getSource() (reconcileSource, sourceReference) {
 			return nil, srcRef
 		}
 	default:
-		// default case assumes the HelmRelease is using a HelmChartTemplate
-		ns = obj.Spec.Chart.Spec.SourceRef.Namespace
-		if ns == "" {
-			ns = obj.Namespace
-		}
+		// default case assumes the HelmRelease is using a HelmChartTemplate.
+		// The HelmChart is always created in the same namespace as the HelmRelease,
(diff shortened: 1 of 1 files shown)


Aman-Cool:
Hey @stefanprodan @matheuscscp, Happy to answer any questions on this; the fix is minimal and self-contained so it should be a straightforward review.

Also wanted to flag: I have a couple of other PRs that have been approved for a while now but haven't been merged yet. Is there anything blocking them or a specific release window I should be aware of? Would love to get those over the line too🙂.

stefanprodan:
The HelmChart is always created in the namespace where the HelmRepository is. See https://github.com/fluxcd/helm-controller/blob/390b7de8dac9445b6507b23aa39d646b79fddfac/api/v2/helmrelease_types.go#L424

Aman-Cool:
@stefanprodan, My bad, the original code is correct. The HelmChart lives in the source's namespace, not the HelmRelease's. Should've traced the controller logic before opening this. Closing.

PR #5763, opened 2026-03-11
ci: add top-level permissions to upgrade-fluxcd-pkg workflow

### LINKED ISSUE #5762: ci: fix excessive GitHub workflow token permissions
## Summary

The OpenSSF Scorecard reports a warning about missing top-level `permissions` in the `upgrade-fluxcd-pkg.yaml` workflow.

## Details

The `upgrade-fluxcd-pkg.yaml` workflow does not declare top-level permissions, which means it gets the default token permissions. Adding explicit `permissions: contents: read` follows the principle of least privilege and improves the repository's OpenSSF Scorecard score.

## References

- [OpenSSF Scorecard Token-Permissions Check](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions)

gaganhr94:
Fixes #5762

## Motivation

The OpenSSF Scorecard reports a Token-Permissions warning for the `upgrade-fluxcd-pkg.yaml` workflow because it lacks top-level `permissions`. Without explicit permissions, the workflow token gets default permissions that may be broader than necessary.

## Modifications

- Added `permissions: contents: read` at the top level of `.github/workflows/upgrade-fluxcd-pkg.yaml` to follow the principle of least privilege.

## Verification

- Verified the YAML syntax is valid.
- The change only restricts default token permissions and does not affect the workflow's functionality, as it uses a separate `BOT_GITHUB_TOKEN` secret.

### DIFF
--- a/.github/workflows/upgrade-fluxcd-pkg.yaml
+++ b/.github/workflows/upgrade-fluxcd-pkg.yaml
@@ -3,6 +3,9 @@ name: upgrade-fluxcd-pkg
 on:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   upgrade-fluxcd-pkg:
     uses: fluxcd/gha-workflows/.github/workflows/upgrade-fluxcd-pkg.yaml@v0.9.0


stefanprodan (approved):
LGTM

Thanks @gaganhr94 

PS. that rule is quite silly as we enforce minimal token access (read-only content) at the org level, there is no need to pollute the workflows with top level permissions. Anyway if we make the OpenSSF scanner happy I'll indulge it.

gaganhr94:
Totally agree @stefanprodan. This PR is mainly to make the OpenSSF Scorecard happy. Currently the score is 8.7, which is pretty good. With these changes, it can go to around 9.6

PR #5782, opened 2026-03-16
feat(flux get): add --output flag, yaml only

mtfurlan:
Copied the flag syntax from kubectl --output so it can be extended later
without breaking changes.

It's not integrated terribly well, but the next step would be to
abstract the printing so we can switch out printers more cleanly and
also pick which information is rendered for each printing format, and I
do not know go or this codebase well enough to do that yet.

Signed-off-by: Mara Furland <mara@fur.land>

### DIFF
--- a/pkg/printers/yaml_printer.go
+++ b/pkg/printers/yaml_printer.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package printers
+
+import (
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/yaml"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/printers/yaml_printer.go
```diff
@@ -0,0 +1,66 @@
...
+						return fmt.Errorf("unsupported type %T", v)
+					}
+					rows = append(rows, s...)
+				}
+			default:
+				return fmt.Errorf("unsupported type %T", arg)
+			}
+		}
+
+		var entries []map[string]string
```
mtfurlan: entries being `[]map[string]string` is a problem because the true/false are not treated as booleans in the yaml but strings

mtfurlan:
> zero rows print nothing—fine if deliberate.

Maybe a warning on stderr would be nice, but if you're doing all resources it's going to be quite hard to check.


> YamlPrinter assumes len(row) >= len(header); a short row would panic

It doesn't panic, but it does assume that header[i] maps to row[i], so if headers is short, then it just won't add the row
You did cause me to notice --no-headers though, so that's fixed

<s>I'm still not sure how to fix the boolean output though</s> EDIT: I did it!

stefanprodan:
For get commands the `--output` should return a `List` type like kubectl does, having the table printed in yaml format is not something I would consider, we need to return all objects in a list.

mtfurlan:
I don't disagree, but the requests for this have existed for a few years and this is still a lot better than the current state.
Could name the output `yamltable` so we could do real yaml later without any breaking change?

If you can help me with a plan I can spend a little more time trying to make it do more real yaml output.
The only idea I have right now is to modify `summarisable` to include a `getRaw(i int) any` and then pass the summarisable to the printer interface, but I don't like any and I don't know go typing well enough to do this on my own.

Issue #5439, opened 2025-07-02
[Feature] Allow `flux diff ks` when the `Kustomization` does not already exist


kitforbes:
### Describe the bug

This is not a bug, and is instead a request to extend the feature set.

`flux diff` is great for identifying changes when the `Kustomization` object already exists on the cluster. If you run `flux diff` when the `Kustomization` does not exist within the cluster, you'll get an error saying:
>✓  Kustomization diffing...
>✗ failed to get kustomization object: [kustomizations.kustomize.toolkit.fluxcd.io](http://kustomizations.kustomize.toolkit.fluxcd.io/) "my-object" not found

This obviously makes sense, but it would be nice if it told me about all of the resources which _would_ be created in this case (similar to `terraform plan`).

### Steps to reproduce

1. Create a `Kustomization` (`kustomize.toolkit.fluxcd.io/v1`) file at `path/to/kustomization.yaml` to manage any resources within `path/to/files`
2. Run `flux diff ks new-kustomization --path path/to/files --kustomization-path path/to/kustomization.yaml`

### Expected behavior

When the `Kustomization` does not exist on the cluster, it would be great if `flux diff` showed:
1. All new resources which would be created by reconciliation (similar to `terraform plan`)
2. differences to existing resources within the cluster which would be managed by Flux. Although this may be harder to achieve.

### OS / Distro

N/A

### Flux version

v2.6.2

### Flux check

N/A

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

cazlo:
I would also like to see this feature.

As a temporary workaround there are a couple options:
- First create the Kustomization in the cluster without any associated resources. Then in subsequent changesets, add resources. In the second changeset, `flux diff` will behave as expected. This has downsides such as requiring 2 changesets which usually comes with additional operational overhead, the ordered operation is easy to forget, etc.
- First do a command like `flux build kustomization --path . --kustomization-file ./clusters/foo/bar.yaml --dry-run > out.yaml`, then run a command like `kubectl diff -f out.yaml`.  This has downsides such as using `--dry-run` does not have the same variable substitution behavior as flux would normally have.

wlaw-drw:
Yeah why is `flux diff ks` attempting to retrieve the Flux kustomization resource from the cluster when I provide it locally using `--kustomization-path`?

rycli:
[...]

PR #5794, opened 2026-03-29
Add `--in-memory-build` to `flux build ks` and `flux diff ks`

### LINKED ISSUE #5781: 'flux build kustomization' corrupts files when run concurrently
### Describe the bug

When multiple instances of the `flux build kustomization` command are run concurrently, there is a chance of corruption due to the file writing of `.original` creation and related logic. This manifests as;
- `kustomization.yaml` files being modified/corrupted
- `kustomization.yaml` files being emptied
- `kustomization.yaml.original` files being left behind
- failing builds, due to the above

A similar build failure also occurs if the repository files/directories are read-only.

### Steps to reproduce

Given that this is a timing-sensitive issue, I've created a set of small reproduction scripts at https://github.com/rycli/fluxcd-kustomize-build-bug-repro . See the README.txt for example output of the failures.

### Expected behavior

One or more of the following:
- `flux kustomize build` does not write any files.
- flux only writes files when necessary (e.g., missing kustomization.yaml file)
- flux writes all of its files to a tempdir/the configured cache dir.

### OS / Distro

macOS 26, Ubuntu 24

### Flux version

v2.8.1

### Additional context

While this doesn't happen on every execution, I'm consistently running into this issue when running Flux in CI/CD pipelines. I'm running a busy multi-tenant environment with very frequent automated checks; these issues end up showing up rather frequently.

### Code of Conduct

rycli:
One half of changes necessary to resolve https://github.com/fluxcd/flux2/issues/5781 ; makes use of changes introduced in the `pkg` dependency by https://github.com/fluxcd/pkg/pull/1153

As suggested, this change keeps the current behavior for building kustomizations internally as part of `flux build kustomization ...`, but allows users to opt-in with the `--in-memory-build` flag, which will use `filesys.MakeFsInMemory()`. 

To keep things a bit cleaner, I opted to add a filesystem "backend" interface; defaults to "on disk", uses "in memory" if opted in. This keeps the rest of the codebase relatively untouched.

### DIFF
--- a/cmd/flux/build_kustomization.go
+++ b/cmd/flux/build_kustomization.go
@@ -72,6 +72,7 @@ type buildKsFlags struct {
 	strictSubst       bool
 	recursive         bool
 	localSources      map[string]string
+	inMemoryBuild     bool
 }
 
 var buildKsArgs buildKsFlags
@@ -85,6 +86,8 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	buildKsCmd.Flags().BoolVarP(&buildKsArgs.recursive, "recursive", "r", false, "Recursively build Kustomizations")
 	buildKsCmd.Flags().StringToStringVar(&buildKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
+	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", true,
+		"Use in-memory filesystem during build.")
 	buildCmd.AddCommand(buildKsCmd)
 }
 
@@ -130,6 +133,7 @@ func buildKsCmdRun(cmd *cobra.Command, args []string) (err error) {
 			build.WithStrictSubstitute(buildKsArgs.strictSubst),
 			build.WithRecursive(buildKsArgs.recursive),
 			build.WithLocalSources(buildKsArgs.localSources),
+			build.WithInMemoryBuild(buildKsArgs.inMemoryBuild),
 		)
 	} else {
 		builder, err = build.NewBuilder(name, buildKsArgs.path,
@@ -140,6 +144,7 @@ func buildKsCmdRun(cmd *cobra.Command, args []string) (err error) {
 			build.WithStrictSubstitute(buildKsArgs.strictSubst),
 			build.WithRecursive(buildKsArgs.recursive),
 			build.WithLocalSources(buildKsArgs.localSources),
+			build.WithInMemoryBuild(buildKsArgs.inMemoryBuild),
 		)
 	}
 
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/fluxcd/pkg/chartutil v1.23.0
 	github.com/fluxcd/pkg/envsubst v1.5.0
 	github.com/fluxcd/pkg/git v0.46.0
-	github.com/fluxcd/pkg/kustomize v1.28.0
+	github.com/fluxcd/pkg/kustomize v1.29.0
 	github.com/fluxcd/pkg/oci v0.63.0
 	github.com/fluxcd/pkg/runtime v0.103.0
 	github.com/fluxcd/pkg/sourceignore v0.17.0
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -45,6 +45,7 @@ import (
 
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
 	"github.com/fluxcd/pkg/kustomize"
+	buildfs "github.com/fluxcd/pkg/kustomize/filesys"
 	runclient "github.com/fluxcd/pkg/runtime/client"
 	ssautil "github.com/fluxcd/pkg/ssa/utils"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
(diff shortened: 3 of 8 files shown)


rycli:
Small note: I see that the DCO workflow is failing... though I added the `Signed-off-by` line with "no alias" as requested. The workflow doesn't fail in this manner on the other PR; something in the configuration?

stefanprodan:
See https://github.com/fluxcd/flux2/pull/5794/checks?check_run_id=69071054594 right at the end it tells you the expected signature.

internal/build/build.go
```diff
@@ -65,6 +65,52 @@ const (
...
+type inMemoryBackend struct{}
+
+func (inMemoryBackend) Generate(gen *kustomize.Generator, dirPath string) (filesys.FileSystem, string, kustomize.Action, error) {
+	manifest, kfilePath, action, err := gen.GenerateManifest(dirPath)
+	if err != nil {
+		return nil, "", action, err
+	}
+
+	memFS := filesys.MakeFsInMemory()
+	if err := loadDirToMemFS(dirPath, memFSRoot, memFS); err != nil {
```
stefanprodan: Is `dirPath` here pointing to `$PWD`? If not, then the build would fail since resources can be point to parent dirs. Let's add a test for this where a kustomization.yaml has:

```
resources:
- ../../some.yaml
```

rycli: My understanding is that these are the paths that are loaded into the builder `resourcesPath`, wherever that may be (`$PWD` or whatever is specified as a local source?); so indeed this would fail if there's a referenced resource that's outside of that scope... What would the desired behavior be in this case; I'm assuming no path restrictions whatsoever?

I'll see if I can implement lazy loading instead, since that would sidestep these issues

stefanprodan: The secureFS we use for building the overlay restricts the refs to the working dir. Now memFS will miss files from the repo root right?

go.mod
```diff
@@ -267,3 +267,5 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.2-0.20260122202528-d9cc6641c482 // indirect
 )
+
+replace github.com/fluxcd/pkg/kustomize => github.com/rycli/fluxcd-pkg/kustomize v0.0.0-20260329153243-0671cee33351
```
rycli: Temporary override to allow testing with the proposed changes to `pkg/kustomize`

internal/build/build.go
rycli: From what I can see this is equivalent to how the current behavior is implemented; i.e., references outside of cwd fail but otherwise would resolve correctly. However in the process of looking into it I realized that the MakeFsOnDiskSecure example might also work better...

I've tried to take a separate approach; introduced a parallel "fs_memory" module that lazy-loads files on read and intercepts the writes; https://github.com/fluxcd/pkg/pull/1153/changes#diff-1e124e1285616703f52e715892784d4537e42152dfe9c02af1978225cbfd170f

Figured by making it like this, it can be composed; `MakeFsOnDiskSecure` wraps `MakeFsOnDisk`, and `MakeFsInMemory` wraps `MakeFsOnDiskSecure`. They both perform their respective checks.

I've also added a couple tests as suggested!

stefanprodan: Nice! Can you also test the recursive build with the in-memory flag to make sure we're not breaking it?

rycli:
@stefanprodan pushed the tests; I added an extra combination to the existing entries in `TestBuild*`. Seems to work 'transparently' as intended 😄

stefanprodan:
@rycli `fluxcd/pkg/kustomize@v1.29.0` contains the in-memory FS. Please bump then rebase your PR.

go.mod
```diff
@@ -267,3 +267,5 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.2-0.20260122202528-d9cc6641c482 // indirect
 )
+
+replace github.com/fluxcd/pkg/kustomize => github.com/rycli/fluxcd-pkg/kustomize v0.0.0-20260329192052-94b031e1aca6
```
stefanprodan: please remove this, then run `make tidy` and force push

rycli: Wups, snuck back in while squashing. Removed.

cmd/flux/build_kustomization.go
```diff
@@ -85,6 +86,8 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	buildKsCmd.Flags().BoolVarP(&buildKsArgs.recursive, "recursive", "r", false, "Recursively build Kustomizations")
 	buildKsCmd.Flags().StringToStringVar(&buildKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
+	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", false,
```
stefanprodan: I'm considering flipping this to `true`, the extra memory usage is negligible for the environments where the CLI runs (dev machines & CI runners). @rycli do you have some large overlay to run the CLI on?

rycli: I've already been running a version of these fixes in CI for a few weeks, without issue. The largest overlays I have would be some exported CRDs; didn't notice any particular issues. ~10mb max.

Intuitively; shouldn't this be essentially equivalent to the current implementation? The reading into memory is passed through to the real fs, so basically unchanged, only the `kustomization.yaml` files are being written, which is negligible data.

stefanprodan: Right the bulk of the memory usage comes from the build itself, the kustomization files can not be large. Let's flip this to `true` by default then.

rycli: How would you like to implement this w.r.t. flag semantics?

stefanprodan: Set the default to `true`, allowing users to opt-out. This feature will be released in Flux 2.9 so it's Ok to change the behaviour if it's for a better UX.

rycli: Cool. I've updated the tests; they now have 'on disk' variants instead of the other way around.

stefanprodan (approved):
LGTM

Thanks @rycli 🥇

stefanprodan:
@rycli if you have time, could you please look into finding a solution for https://github.com/fluxcd/flux2/issues/5439

cmd/flux/build_kustomization.go
```diff
@@ -85,6 +86,8 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	buildKsCmd.Flags().BoolVarP(&buildKsArgs.recursive, "recursive", "r", false, "Recursively build Kustomizations")
 	buildKsCmd.Flags().StringToStringVar(&buildKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
+	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", true,
```
matheuscscp, line 89: This is a breaking change, we need to default to `false`

matheuscscp: Filed a bug here: https://github.com/fluxcd/flux2/issues/5968

PR #5795, opened 2026-03-30
[RFC-0013] Flux CLI Plugin System

stefanprodan:
This RFC proposes a plugin system for the Flux CLI that allows external CLI tools to be discoverable and invocable as `flux <name>` subcommands. Plugins are installed from a centralized catalog hosted on GitHub, with SHA-256 checksum verification and automatic version updates.

The implementation of this RFC can be tests here: https://github.com/fluxcd/flux2/tree/plugin-system

### DIFF
--- a/rfcs/0013-cli-plugin-system/README.md
+++ b/rfcs/0013-cli-plugin-system/README.md
@@ -1,10 +1,10 @@
-# RFC-XXXX Flux CLI Plugin System
+# RFC-0013 Flux CLI Plugin System
 
-**Status:** provisional
+**Status:** implementable
 
 **Creation date:** 2026-03-30
 
-**Last update:** 2026-04-01
+**Last update:** 2026-04-13
 
 ## Summary
 


rfcs/0013-cli-plugin-system/README.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+#### Plugin Author Publishes a Plugin
+
+As a plugin author, I want to submit my tool to the Flux plugin catalog so that
+Flux users can install it with `flux plugin install <name>`.
+
+1. Release binary with GoReleaser (produces tarballs/zips + checksums)
+2. Submit a PR to `fluxcd/plugins` with `plugins/<name>.yaml`
+3. Subsequent releases are picked up by automated polling workflows
```
matheuscscp, line 237: I would extend this section to outline a process where the plugin author is responsible for stamping new versions picked up by the automation

stefanprodan: Added  in the last commit, please check.

matheuscscp: LGTM!

rfcs/xxxx-cli-plugin-system/README.md
```diff
@@ -0,0 +1,320 @@
...
+## Proposal
+
+### Plugin Discovery
+
+Plugins are executables prefixed with `flux-` placed in a single plugin directory.
+The `flux-<name>` binary maps to the `flux <name>` command. For example,
+`flux-operator` becomes `flux operator`.
+
+The default plugin directory is `~/.fluxcd/plugins/`. Users can override it with the
+`$FLUXCD_PLUGINS` environment variable. Only this single directory is scanned.
```
hiddeco: Nit:

```suggestion
`FLUXCD_PLUGINS` environment variable. Only this single directory is scanned.
```

rfcs/0013-cli-plugin-system/README.md
```diff
@@ -0,0 +1,320 @@
...
+serves as the plugin catalog. Each plugin has a YAML manifest:
+
+```yaml
+apiVersion: cli.fluxcd.io/v1beta1
+kind: Plugin
+name: operator
+description: Flux Operator CLI
+homepage: https://fluxoperator.dev/
+source: https://github.com/controlplaneio-fluxcd/flux-operator
+bin: flux-operator
```
hiddeco, line 103: On Windows, this most likely will be suffixed by `.exe`. Additionally, I know about projects that e.g. have a unique binary name per architecture or platform, or do not produce an archive for them at all.

It feels like the current state of the document does not properly cover:

1. The contract of how a plugin should be released (archived, uniform binary name, `.exe` suffixing magic for Windows).
2. Additional controls to be able to provide platform level overwrites, direct consumption of binary release artifacts, etc.

rfcs/xxxx-cli-plugin-system/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+The `internal/plugin` package uses dependency injection (injectable `ReadDir`, `Stat`,
+`GetEnv`, `HomeDir` on a `Handler` struct) for testability. Tests mock these functions
+directly without filesystem fixtures.
+
+### Plugin Directory
+
+- **Default**: `~/.fluxcd/plugins/` -- auto-created by install/update commands
+  (best-effort, no error if filesystem is read-only).
+- **Override**: `$FLUXCD_PLUGINS` env var replaces the default directory path.
```
hiddeco: Nit:

```suggestion
- **Override**: `FLUXCD_PLUGINS` env var replaces the default directory path.
```

rfcs/0013-cli-plugin-system/README.md
stefanprodan: Currently in the PoC when the OS is set to Windows we append `.exe` and we expect it to be in a `.zip` instead of `.tgz` (linux, macos). Going to add support for binaries too, and use the `bin` field to rename the upstream asset to `flux-<bin>`.

Thanks for the feedback, I'll update the RFC soon with the new procedure.

stefanprodan: I have updated the RFC and I've added direct binary URL support to the PoC. Please check.

rfcs/xxxx-cli-plugin-system/README.md
```diff
@@ -0,0 +1,334 @@
...
+        checksum: sha256:9712026094a5...
+```
+
+The plugin manifest includes metadata (name, description, homepage, source repo), the binary name
+(`bin`), and a list of versions with platform-specific download URLs and checksums.
+
+The download URLs can point to one of the following formats:
+
+- An archive containing the binary (`tar`, `tar.gz` or `zip`), with the binary at the root of the archive. The binary name inside the archive must match the `bin` field in the manifest.
+- An optional field `extractPath` can be specified in the `platforms` entry to indicate a subdirectory inside the archive where the binary is located (e.g., `extractPath: bin/flux-operator`). If `extractPath` is not specified, the CLI looks for the binary at the root of the archive.
```
hiddeco:
```suggestion
- An optional `extractPath` field can be specified in a `platforms` entry to override this default, either because the binary has a different name on this platform, or because it is nested in a subfolder rather than at the root of the archive. It accepts an absolute path to a file within the archive (e.g., `bin/flux-operator`).
```

PR #5828, opened 2026-04-09
Add `--show-source` to `flux get ks` and `flux get hr` 

### LINKED ISSUE #2692: Show source name as additional printer column

rafaelperoco:
When looking at a cluster with Kustomizations or HelmReleases from different sources it is hard to identify which source each resource refers to. This adds a `--show-source` flag to both `flux get kustomization` and `flux get helmrelease` that displays the source Kind/Namespace/Name.

Follows the same format used in `get sources external`.

For HelmReleases, supports both `ChartRef` and `Chart.Spec.SourceRef` source references.

Fixes #2692

### DIFF
--- a/cmd/flux/get_helmrelease.go
+++ b/cmd/flux/get_helmrelease.go
@@ -28,13 +28,22 @@ import (
 	helmv2 "github.com/fluxcd/helm-controller/api/v2"
 )
 
+type getHelmReleaseFlags struct {
+	showSource bool
+}
+
+var getHrArgs getHelmReleaseFlags
+
 var getHelmReleaseCmd = &cobra.Command{
 	Use:     "helmreleases",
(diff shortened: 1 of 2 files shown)


rafaelperoco:
@stefanprodan this implements the --show-source flag you suggested in #2692.

stefanprodan:
For consistency we should have the same flag for `flux get hr`. @rafaelperoco can you do that as well?

rafaelperoco:
> For consistency we should have the same flag for `flux get hr`. @rafaelperoco can you do that as well?

sure

rafaelperoco:
done @stefanprodan

stefanprodan (approved):
LGTM

Thanks @rafaelperoco 🏅

makkes:
Fantastic! Thanks for handling this @rafaelperoco.

PR #5831, opened 2026-04-10
Add `--ns-follows-kube-context` global flag for using the kubeconfig context namespace

### LINKED ISSUE #3453: Flux CLI tool doesn't respect context namespace

jtyr:
## Problem

The `flux` CLI ignores the namespace set in the current kubeconfig context and always defaults to `flux-system` when `--namespace` is not explicitly provided. This means users who set a namespace via `kubectl config set-context --current --namespace=...` still have to pass `-n` on every `flux` command.

Fix: #3453

## Fix

Resolve the namespace from the kubeconfig context when neither `--namespace` nor `FLUX_SYSTEM_NAMESPACE` is set. The `flux-system` default is preserved as the final fallback.

The resulting priority order (highest to lowest):

1. `--namespace` / `-n` flag
2. `FLUX_SYSTEM_NAMESPACE` environment variable
3. Kubeconfig context namespace
4. `flux-system`

The context lookup runs in `PersistentPreRunE` (after flag parsing), so it correctly respects `--context` and `--kubeconfig` flags when determining which context to read from.

## Testing

Added unit tests covering:

- Namespace returned from the current context
- `--context` flag selecting a different context's namespace
- Context with no namespace set (empty string)
- Non-existent context
- `FLUX_SYSTEM_NAMESPACE` taking precedence over context namespace
- `--namespace` flag taking precedence over context namespace
- Fallback to `flux-system` when context has no namespace

### DIFF
--- a/cmd/flux/main.go
+++ b/cmd/flux/main.go
@@ -100,6 +100,16 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.`,
   # Uninstall Flux and delete CRDs
   flux uninstall`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// If opted in via --ns-follows-kube-context flag or
+		// FLUX_NS_FOLLOWS_KUBE_CONTEXT env var, and --namespace was not
+		// explicitly set, respect the namespace from the kubeconfig context.
+		if !cmd.Flags().Changed("namespace") &&
+			(rootArgs.nsFollowsKubeContext || os.Getenv("FLUX_NS_FOLLOWS_KUBE_CONTEXT") != "") {
+			if ctxNs := getKubeconfigContextNamespace(kubeconfigArgs); ctxNs != "" {
+				*kubeconfigArgs.Namespace = ctxNs
+			}
+		}
+
 		ns, err := cmd.Flags().GetString("namespace")
 		if err != nil {
 			return fmt.Errorf("error getting namespace: %w", err)
@@ -116,10 +126,11 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.`,
 var logger = stderrLogger{stderr: os.Stderr}
 
 type rootFlags struct {
-	timeout      time.Duration
-	verbose      bool
-	pollInterval time.Duration
-	defaults     install.Options
+	timeout              time.Duration
+	verbose              bool
+	pollInterval         time.Duration
+	nsFollowsKubeContext bool
+	defaults             install.Options
 }
 
 // RequestError is a custom error type that wraps an error returned by the flux api.
@@ -139,6 +150,8 @@ var kubeclientOptions = new(runclient.Options)
 func init() {
 	rootCmd.PersistentFlags().DurationVar(&rootArgs.timeout, "timeout", 5*time.Minute, "timeout for this operation")
 	rootCmd.PersistentFlags().BoolVar(&rootArgs.verbose, "verbose", false, "print generated objects")
+	rootCmd.PersistentFlags().BoolVar(&rootArgs.nsFollowsKubeContext, "ns-follows-kube-context", false,
+		"use the namespace from the kubeconfig context instead of the default flux-system namespace, can also be set via FLUX_NS_FOLLOWS_KUBE_CONTEXT env var")
 
 	configureDefaultNamespace()
 	kubeconfigArgs.APIServer = nil // prevent AddFlags from configuring --server flag
@@ -203,6 +216,26 @@ func main() {
 		logger.Failuref("%v", err)
 		os.Exit(1)
 	}
+}
+
+// getKubeconfigContextNamespace returns the namespace from the current
+// kubeconfig context, or an empty string if it cannot be determined.
+func getKubeconfigContextNamespace(cf *genericclioptions.ConfigFlags) string {
+	rawConfig, err := cf.ToRawKubeConfigLoader().RawConfig()
+	if err != nil {
+		return ""
+	}
+
+	currentContext := rawConfig.CurrentContext
+	if cf.Context != nil && *cf.Context != "" {
+		currentContext = *cf.Context
+	}
+
+	if ctx, ok := rawConfig.Contexts[currentContext]; ok {
+		return ctx.Namespace
+	}
+
+	return ""
 }
 
 func configureDefaultNamespace() {
--- a/cmd/flux/main_context_ns_test.go
+++ b/cmd/flux/main_context_ns_test.go
@@ -1,0 +1,221 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+func TestGetKubeconfigContextNamespace(t *testing.T) {
+	tests := []struct {
+		name           string
+		kubeconfig     string
+		context        string
+		expectedResult string
+	}{
+		{
+			name: "returns namespace from current context",
+			kubeconfig: `apiVersion: v1
+kind: Config
+current-context: my-context
+contexts:
+- name: my-context
+  context:
+    cluster: my-cluster
+    namespace: custom-ns
+clusters:
+- name: my-cluster
+  cluster:
+    server: https://localhost:6443
+`,
+			expectedResult: "custom-ns",
+		},
+		{
+			name: "returns empty when context has no namespace",
+			kubeconfig: `apiVersion: v1
+kind: Config
+current-context: my-context
+contexts:
+- name: my-context
+  context:
+    cluster: my-cluster
+clusters:
+- name: my-cluster
+  cluster:
+    server: https://localhost:6443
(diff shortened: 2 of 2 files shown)


stefanprodan:
This is a major breaking change that we can not accept. Imagine how many CI pipelines and custom scripts we are going to break with this, e.g. you have `flux bootstrap` in your pipeline, the default namespace in kubeconfig is to some namespace, with this change new clusters will end up with Flux in the selected namespace instead of `flux-system`.

matheuscscp:
We had lots of discussions around this over the years. This is a bug that became expected behavior at this point, so we cannot break it.

-------

However, across discussions someone suggested an env var for opting into this behavior, which would not be a breaking change and I'd be ok with it. Env var or flag, not sure what's best, though.

jtyr:
Thanks for the feedback, @stefanprodan. I understand your concern.

This behavior was discussed and classified as a bug in #3453, with the consensus being that the kubeconfig context namespace should be respected (see https://github.com/fluxcd/flux2/issues/3453#issuecomment-2541446153). That said, I understand the concern about breaking existing CI pipelines and scripts that rely on `flux-system` being the implicit default.

Would an opt-in approach work for now? For example, introducing a `FLUX_NS_FOLLOWS_KUBE_CONTEXT` environment variable that users can set to enable the context namespace resolution, while keeping the current behavior as the default. This way existing workflows remain unaffected and users who want this can explicitly opt in, with the option to make it the default in a future major version.

Or would you prefer a different approach entirely?

jtyr:
Thanks for joining the discussion, @matheuscscp. How about supporting both?

- A `--ns-follows-kube-context` flag, which can also be set via an alias (`alias flux='/usr/bin/flux --ns-follows-kube-context'`)
- A `FLUX_NS_FOLLOWS_KUBE_CONTEXT` environment variable (`export FLUX_NS_FOLLOWS_KUBE_CONTEXT=1`)

This follows the existing pattern where `FLUX_SYSTEM_NAMESPACE` provides an env var alternative. The flag gives per-command control, while the env var allows setting it globally in a shell profile or CI environment.

Would that be acceptable?

matheuscscp:
I have no objections to this solution

jtyr:
@matheuscscp @stefanprodan I have updated the implementation to make the context namespace resolution opt-in. It can be enabled via one of the following:

- `--ns-follows-kube-context` flag (e.g. `alias flux='flux --ns-follows-kube-context'`)
- `FLUX_NS_FOLLOWS_KUBE_CONTEXT` env var (e.g. `export FLUX_NS_FOLLOWS_KUBE_CONTEXT=1`)

Without either of these, the existing behavior is preserved (`flux-system` remains the default).

matheuscscp (approved):
LGTM

cmd/flux/main.go
```diff
@@ -100,6 +100,17 @@ Command line utility for assembling Kubernetes CD pipelines the GitOps way.`,
   # Uninstall Flux and delete CRDs
   flux uninstall`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// If opted in via --ns-follows-kube-context flag or
+		// FLUX_NS_FOLLOWS_KUBE_CONTEXT env var, and --namespace was not
+		// explicitly set and FLUX_SYSTEM_NAMESPACE env var is not set,
+		// respect the namespace from the kubeconfig context.
+		if (rootArgs.nsFollowsKubeContext || os.Getenv("FLUX_NS_FOLLOWS_KUBE_CONTEXT") != "") &&
+			!cmd.Flags().Changed("namespace") && os.Getenv("FLUX_SYSTEM_NAMESPACE") == "" {
```
makkes: I'm not sure I understand why FLUX_SYSTEM_NAMESPACE is checked here. I don't always care about that as my Flux resources may be scattered across multiple namespaces on the cluster. Why should this generally disregard the kubeconfig NS if FLUX_SYSTEM_NAMESPACE is set?

cmd/flux/main_context_ns_test.go
```diff
@@ -0,0 +1,234 @@
...
+			nsFollowsFlag:    true,
+			expectedNamespace: "context-ns",
+		},
+		{
+			name:              "uses context namespace when opted in via env var",
+			nsFollowsEnv:     "1",
+			expectedNamespace: "context-ns",
+		},
+		{
+			name:              "FLUX_SYSTEM_NAMESPACE takes precedence over context namespace",
```
makkes: I don't believe this should be the expected behaviour as explained above.

cmd/flux/main.go
matheuscscp: Yeah, if FLUX_SYSTEM_NAMESPACE is not used when e.g. `-n` is set, then it makes no sense to be checked here, good catch!

jtyr: Fixed in the last commit.

cmd/flux/main_context_ns_test.go
jtyr: Fixed in the last commit.

makkes (approved):

jtyr:
@stefanprodan Please could you review this again?

jtyr:
@stefanprodan @matheuscscp Any idea why the E2E test is failing?

matheuscscp:
> @stefanprodan @matheuscscp Any idea why the E2E test is failing?

@jtyr I investigated this for a while with Opus 4.7 and this is how far I got:

```
  Failure: The Run tests step (make test --tags=unit) fails in cmd/flux with:                                                                                                                                 
  - TestTrace/* — "deployments.apps 'podinfo' not found" (and similar for HelmRelease)                                                                                                                        
  - TestTree/* — "kustomizations... 'flux-system' not found"                                                                                                                                                  
                                                            
  Root cause: the new test file cmd/flux/main_context_ns_test.go (added in the PR) leaks state into subsequent tests. I reproduced it locally:                                                                
                                                                                                                                                                                                              
  - TestTrace alone → passes                                                                                                                                                                                  
  - TestContextNamespaceOptIn alone → passes                                                                                                                                                                  
  - Both together → TestTrace fails with "not found" against the envtest API server                                                                                                                           
                                                                                                                                                                                                              
  The new TestContextNamespaceOptIn test mutates the package-global kubeconfigArgs (a *genericclioptions.ConfigFlags): swaps KubeConfig/Context pointers, calls configureDefaultNamespace(), calls            
  getKubeconfigContextNamespace() (which invokes kubeconfigArgs.ToRawKubeConfigLoader()). Even though the test's t.Cleanup restores the visible pointer fields, ConfigFlags internally caches a clientConfig  
  (see cli-runtime/pkg/genericclioptions/config_flags.go — clientConfig + clientConfigLock), and that cache is not reset. The cached loader was bound to the per-subtest temp kubeconfig (which got           
  t.TempDir-deleted), and bisecting the 6 subtests shows the failure only triggers when multiple subtests run together — consistent with cumulative cache poisoning.

  Fix direction: the new test shouldn't mutate the package-global kubeconfigArgs. Either build a local *genericclioptions.ConfigFlags for the test (e.g. genericclioptions.NewConfigFlags(false)) and refactor
   getKubeconfigContextNamespace to take it as a parameter, or have cleanup replace kubeconfigArgs with a freshly constructed instance (and then re-kubeconfigArgs.KubeConfig = &testEnv.kubeConfigPath) so
  the internal cache is discarded.
```

jtyr:
Thanks for providing the analysis, @matheuscscp. I have refactored `getKubeconfigContextNamespace` to take a `*genericclioptions.ConfigFlags` parameter, eliminating its dependency on the package-global `kubeconfigArgs`. The `getKubeconfigContextNamespace`, called from `PersistentPreRunE`, now passes `kubeconfigArgs` explicitly. Tests now construct their own local `genericclioptions.NewConfigFlags(false)` instance per subtest. This means:

- The internal `clientConfig` cache lives only on the local instance.
- No mutation of `kubeconfigArgs` whatsoever.
- No `t.Cleanup` needed for `kubeconfigArgs` state restoration.
- Subsequent tests like `TestTrace` and `TestTree` see a clean global `kubeconfigArgs`.

Please run the tests again so we can see if it passes now.

jtyr:
Cool, tests pass now. Can we merge it now? ;o)

matheuscscp:
> This is a major breaking change that we can not accept. Imagine how many CI pipelines and custom scripts we are going to break with this, e.g. you have `flux bootstrap` in your pipeline, the default namespace in kubeconfig is to some namespace, with this change new clusters will end up with Flux in the selected namespace instead of `flux-system`.

@stefanprodan The implementation is no longer a breaking change, it's opt-in now. Can we merge?

stefanprodan (approved):

matheuscscp:
@jtyr Please squash the commits into one matching the PR title and force-push so we can merge :pray:

jtyr:
@matheuscscp I have squashed the commits 👍

matheuscscp:
@jtyr Thanks very much! You may not realize, but this is quite a feature! Probably top 10 requested by users historically. We should do a bit of noise about it! Can you please open a PR here: https://github.com/fluxcd/website/blob/main/content/en/roadmap.md?plain=1#L108-L113

makkes:
Absolutely agree that this is a feature much longed for! Thanks @jtyr.

PR #5833, opened 2026-04-10
fix: handle multiple symlinks to same target in build artifact

### LINKED ISSUE #5832: `flux build artifact --resolve-symlinks` fails to bundle content when multiple symlinks point to the same target
## Description
When using the `--resolve-symlinks` flag with `flux build artifact` (or `flux push artifact`), the command fails to include the contents of a directory if that directory is targeted by more than one symlink within the source path.

This happens because the `copyDir` function in `cmd/flux/build_artifact.go` uses a global `visited` map to track absolute paths and prevent infinite loops (cycles). However, once a directory is visited via the first symlink, it remains in the `visited` map permanently for the duration of the command. When the traversal encounters a second symlink pointing to the same absolute path, it incorrectly identifies it as a cycle and skips the contents, resulting in empty directories in the final artifact.

## Proposed Fix
The `visited` map should only track the current recursive path (to break actual cycles) rather than acting as a global cache for the entire traversal. 

The fix is to remove the absolute path from the `visited` map once the function returns. Using `defer` ensures that the record is cleaned up after the directory branch has been fully walked, allowing the same target to be reached via other valid paths.

**Implementation Plan:**
Modify `cmd/flux/build_artifact.go` in the `copyDir` function:

```go
func copyDir(srcDir, dstDir string, visited map[string]bool) error {
    // ... absolute path resolution ...
    if visited[abs] {
        return nil // break the cycle
    }
    visited[abs] = true
+   defer delete(visited, abs) // Remove from visited after walking this branch
    
    // ... rest of the copy logic ...
}
```

### Steps to reproduce

1. Create a directory structure where two different symlinks point to the same "base" directory:
   ```bash
   mkdir -p repro/base
   echo "apiVersion: v1" > repro/base/manifest.yaml
   
   # Create two different symlinks to the same base
   ln -s ./base repro/app-a
   ln -s ./base repro/app-b
   ```

2. Build the artifact using the `--resolve-symlinks` flag:
   ```bash
   flux build artifact --path ./repro --output repro.tgz --resolve-symlinks
   ```

3. Inspect the contents of the generated artifact:
   ```bash
   tar -ztvf repro.tgz
   ```

**Actual Results:**
- `app-a/manifest.yaml` exists.
- `app-b/` directory exists but is **EMPTY**.
- `base/` directory exists but is **EMPTY**.

### Expected behavior

- All three directories (`app-a/`, `app-b/`, and `base/`) should contain the `manifest.yaml` file.

_No response_

### OS / Distro

Ubuntu 25.10

### Flux version

Affected versions: All versions containing the --resolve-symlinks feature.

### Flux check

N/A

### Git provider

_No response_

_No response_

### Additional context

- This bug silently break artifacts for users employing "Kustomize-style" or "Nix-style" structures where common bases are symlinked multiple times.

### Code of Conduct

Iam-Karan-Suresh:
### Summary

Fixes an issue where `flux build artifact --resolve-symlinks` fails to include contents when multiple symlinks point to the same target directory.

Followup: #5724

---

### Problem

The `copyDir` function uses a global `visited` map to track visited absolute paths and prevent cycles. However, when a directory is referenced by multiple symlinks:

* The first traversal marks the path as visited
* Subsequent traversals treat it as a cycle
* This results in skipped content and empty directories in the final artifact

---

### Root Cause

The `visited` map is used globally across the entire traversal instead of being scoped to the current recursion path. This causes valid paths to be incorrectly skipped.

---

### Solution

Scope the `visited` map to the current recursion path by removing entries after traversal using `defer`.

```go
visited[abs] = true
defer delete(visited, abs)
```

This ensures:

* Cycles are still prevented
* Multiple valid symlink paths are processed correctly

---

### Reproduction Steps

```bash
mkdir -p repro/base
echo "apiVersion: v1" > repro/base/manifest.yaml

ln -s ./base repro/app-a
ln -s ./base repro/app-b

./bin/flux build artifact --path ./repro --output repro.tgz --resolve-symlinks
tar -ztvf repro.tgz
```

---

### Expected Behavior

All directories should contain the file:

* `app-a/manifest.yaml`
* `app-b/manifest.yaml`
* `base/manifest.yaml`

---

### Actual Behavior (before fix)

* `app-a/manifest.yaml` ✅
* `app-b/` ❌ empty
* `base/` ❌ empty

---

### Testing

* Verified fix using local build of Flux CLI
* Confirmed all symlink paths are correctly included in the artifact
* Ensured no regression in cycle detection

---

closes: #5832

### DIFF
--- a/cmd/flux/build_artifact.go
+++ b/cmd/flux/build_artifact.go
@@ -172,7 +172,7 @@ func copyDir(srcDir, dstDir string, visited map[string]bool) error {
 		return nil // break the cycle
 	}
 	visited[abs] = true
-
+	defer delete(visited, abs)
 	entries, err := os.ReadDir(srcDir)
 	if err != nil {
 		return err
--- a/cmd/flux/build_artifact_test.go
+++ b/cmd/flux/build_artifact_test.go
@@ -179,3 +179,39 @@ func Test_resolveSymlinks_cycle(t *testing.T) {
 	_, err = os.Stat(filepath.Join(resolved, "cycle", "cycle", "cycle"))
 	g.Expect(os.IsNotExist(err)).To(BeTrue())
 }
+
+func Test_resolveSymlinks_multipleLinksSameTarget(t *testing.T) {
+	g := NewWithT(t)
+
+	// Create source directory with a real file inside a dir
+	srcDir := t.TempDir()
+	targetDir := filepath.Join(srcDir, "target")
+	g.Expect(os.MkdirAll(targetDir, 0o755)).To(Succeed())
+	g.Expect(os.WriteFile(filepath.Join(targetDir, "file.yaml"), []byte("data"), 0o644)).To(Succeed())
+
+	// Create a directory with multiple symlinks pointing to targetDir
+	symlinkDir := t.TempDir()
+
+	// Link 1
+	link1 := filepath.Join(symlinkDir, "link1")
+	g.Expect(os.Symlink(targetDir, link1)).To(Succeed())
+
+	// Link 2
+	link2 := filepath.Join(symlinkDir, "link2")
+	g.Expect(os.Symlink(targetDir, link2)).To(Succeed())
+
+	// Resolve symlinks
+	resolved, cleanupDir, err := resolveSymlinks(symlinkDir)
+	g.Expect(err).To(BeNil())
+	t.Cleanup(func() { os.RemoveAll(cleanupDir) })
+
+	// Verify link1 has the file
+	content, err := os.ReadFile(filepath.Join(resolved, "link1", "file.yaml"))
+	g.Expect(err).To(BeNil())
+	g.Expect(string(content)).To(Equal("data"))
+
+	// Verify link2 ALSO has the file
+	content2, err := os.ReadFile(filepath.Join(resolved, "link2", "file.yaml"))
+	g.Expect(err).To(BeNil())
+	g.Expect(string(content2)).To(Equal("data"))
+}


stefanprodan:
Please add a test that proves this fixes the issue.

cmd/flux/symlink_bug_test.go
```diff
@@ -0,0 +1,49 @@
+package main
```
stefanprodan: move this to build_artifact_test.go

stefanprodan (changes requested):

cmd/flux/symlink_bug_test.go
Iam-Karan-Suresh: @stefanprodan  i made changes according to your instructions

stefanprodan:
Please run `make fmt` and force push

Iam-Karan-Suresh:
Hi @stefanprodan, I’ve addressed the formatting issues and updated the changes accordingly. I’d appreciate it if you could take a moment to review. Thanks!

stefanprodan:
Rebase with upstream main and force push please.

stefanprodan:
You need to sync your fork and rebase

Iam-Karan-Suresh:
Hi @stefanprodan, I’ve synced my fork with upstream main, rebased the branch, and force-pushed the latest changes. 
Apologies for the earlier mistakes — I’ve corrected them and will make sure to avoid these in future PRs. Thanks for your guidance!

stefanprodan (approved):
LGTM

Thanks @Iam-Karan-Suresh for the fix 🏅

PR #5841, opened 2026-04-11
docs: Add AI Coding Assistants Guidance 

stefanprodan:
Add guidance for AI-assisted contributions to the Flux project.  Part of: https://github.com/fluxcd/flux2/issues/5846

TL;DR
- No Agent ID in `Signed-off-by` and `Co-authored-by` tags (only humans can legally certify DCO)
- Add Agent ID with `Assisted-by: <agent>/<model>` tag (required on all commits done with AI)

Preview: https://github.com/fluxcd/flux2/blob/ai-contrib-guidelines/CONTRIBUTING.md#ai-coding-assistants-guidance

References:
- [Linux Kernel](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-assistants.rst)
- [Kubernetes](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,47 @@ The main repositories in the Flux project are:
 - [fluxcd/image-automation-controller](https://github.com/fluxcd/image-automation-controller): Kubernetes operator for patching container image tags and digests in Git repositories
 - [fluxcd/website](https://github.com/fluxcd/website): The Flux documentation website accessible at <https://fluxcd.io/>
 
+## AI Coding Assistants Guidance
+
+Using AI Agents to help write your PR is acceptable, but as the author, you are responsible
+for understanding the code and the documentation you submit. Please review all the AI-generated
+content and make sure it follows the guidelines in this document before submitting your PR.
+
+All Flux repositories contain an `AGENTS.md` file. You must point your AI Agent to
+`AGENTS.md` and ask it to follow the guidelines and conventions described there.
+
+Trim down the verbiage in the PR description, commit messages and code comments.
+When engaging with Flux maintainers please refrain from using AI Agents to
+generate responses, we want to talk to you, not to your AI Agent.
+
+AI Agents **must not** add `Signed-off-by` or `Co-authored-by` tags to the commit message.
+Only humans can legally certify the Developer Certificate of Origin ([DCO](https://developercertificate.org/)).
+
+You should disclose the use of AI Agents in the description of your PR and
+in the commit message using the `Assisted-by: AGENT_NAME/LLM_VERSION` tag.
+
+Adding the `Assisted-by` tag to the commit message can be done with:
+
+```sh
+git commit -s -m "Your commit message" --trailer "Assisted-by: <agent>/<model>"
+```
+
+**Note** that the `Signed-off-by` tag is set via the `-s` flag using your real name and email
+(`user.name` and `user.email` must be set in Git config).
+
+Example of a commit message disclosing the use of AI assistance:
+
+```text
+Add version info to plugin listing
+
+Add a version column to the `flux plugin list` table output and populate
+it with the semantic version info extracted from the plugin's recipe file.
+For plugins installed via symlinks, the version is set to `unknown`.
+
+Signed-off-by: Jane Doe <jane.doe@example.com>
+Assisted-by: copilot/gpt-5.4
+```
+
 ## Certificate of Origin
 
 By contributing to this project you agree to the Developer Certificate of Origin (DCO).
@@ -59,8 +100,8 @@ For the Flux project we prefer the following rules:
 - Limit the subject to 50 characters, start with a capital letter and do not end with a period.
 - Explain what and why in the body, if more than a trivial change; wrap it at 72 characters.
 - Use the imperative mood in the subject line (e.g., "Add support for X" instead of "Added support for X" or "Adds support for X").
-- Do not include the issue number in the commit message, use the PR description instead (e.g., "Fixes #123" or "Closes #123").
-- Do not include GitHub mentions (e.g., `@username` or `@team`) within the commit message.
+- Do not include GitHub mentions to issues in the commit message, use the PR description instead (e.g., "Fixes #123" or "Closes #123").
+- Do not include GitHub mentions to accounts (e.g., `@username` or `@team`) within the commit message.
 
 ## Pull Request Process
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,154 +1,120 @@
...
+- [fluxcd/website](https://github.com/fluxcd/website): The Flux documentation website accessible at <https://fluxcd.io/>.
 
-If you are entirely new to Flux and the GitOps Toolkit,
-you might want to take a look at the [introductory talk and demo](https://www.youtube.com/watch?v=qQBtSkgl7tI).
+## AI Coding Assistants Guidance
 
-This project is composed of:
+Using AI Agents to help write your PR is acceptable, but as the author, you are responsible
+for understanding the code and the documentation you submit. Please review all the AI-generated
+content and make sure it follows the guidelines in this document before submitting your PR.
```
relu, line 31: I think it would be useful to suggest directing the agent to this file, that could solve most of the conformance issues.

stefanprodan: Yes, we'll add `AGENTS.md` and other agent-specific files, once we agree the guidance is OK.

CONTRIBUTING.md
```diff
@@ -1,154 +1,117 @@
...
-- [source-controller](https://github.com/fluxcd/source-controller): Kubernetes operator for managing sources (Git, OCI and Helm repositories, S3-compatible Buckets)
-- [source-watcher](https://github.com/fluxcd/source-watcher): Kubernetes operator for advanced source composition and decomposition patterns
-- [kustomize-controller](https://github.com/fluxcd/kustomize-controller): Kubernetes operator for building GitOps pipelines with Kustomize
-- [helm-controller](https://github.com/fluxcd/helm-controller): Kubernetes operator for building GitOps pipelines with Helm
-- [notification-controller](https://github.com/fluxcd/notification-controller): Kubernetes operator for handling inbound and outbound events
-- [image-reflector-controller](https://github.com/fluxcd/image-reflector-controller): Kubernetes operator for scanning container registries
-- [image-automation-controller](https://github.com/fluxcd/image-automation-controller): Kubernetes operator for patches container image tags in Git
+You should disclose the use of AI Agents in the description of your PR and
+in the commit message using the `Assisted-by: AGENT_NAME LLM_VERSION` tag.
```
makkes: I've seen this format suggested in other projects but I'm not quite sure of the value of knowing which LLM version or which specific agent was used.

makkes (approved):
🎯

CONTRIBUTING.md
stefanprodan: There is a lot of value in this from a statistical perspective. Also a great hint to us, if you see non-reasoning modules, there is almost a guarantee the code is 100% AI Slop unless the contributor actually rewrote it.

matheuscscp: This is going to disrupt my workflow: `git commit -asm "something"` (I type most commit commands by myself)

I'd prefer requiring disclosure only in the PR description, which I already take time to write by hand anyway.

Also, Kubernetes explicitly denied the use of `assisted-by`:

> Listing AI tooling as a co-author, co-signing commits using an AI tool, or using the assisted-by, co-developed or similar commit trailer is not allowed.

I don't see why we need to do the opposite. Stats based on PRs should be enough.

stefanprodan: I don't agree with Kubernetes stance here, I think Linux does it the right way. Having the tag in Git commits allows tools to creates stats based on Git history instead of being vendor-locked into a proprietary API like PRs and subject to GitHub rate limits.

Also we'll have an `AGENTS.md` file that will instruct the AI to add the `Assisted-by`, zero effort from the contributor.

matheuscscp: Is there a native git feature to make this easier? I don't wanna be forced to use AI to run destructive git commands

relu: One approach might be to use the [commit-msg git hook](https://git-scm.com/docs/githooks.html#_commit_msg). Use an environment variable to toggle the field.

`.git/hooks/commit-msg`:
```sh
#!/bin/sh

# Only append if ASSISTED_BY is set
if [ -z "$ASSISTED_BY" ]; then
  exit 0
fi

COMMIT_MSG_FILE="$1"
CURRENT_MSG=$(cat "$COMMIT_MSG_FILE")

# Avoid double-appending (e.g. on --amend)
if grep -q "^Assisted-By:" "$COMMIT_MSG_FILE"; then
  exit 0
fi

# Append with a blank line separator (Git trailer convention)
printf "%s\n\nAssisted-By: %s\n" "$CURRENT_MSG" "$ASSISTED_BY" > "$COMMIT_MSG_FILE"
```

Note that I didn't test this, treat it as an example.

relu (approved):
Looks reasonable and straightforward. 👍

CONTRIBUTING.md
matheuscscp: Looks reasonable :+1:

matheuscscp (approved):
LGTM

hiddeco (approved):

CONTRIBUTING.md
stefanprodan: @matheuscscp I've wrote a skill here, please check it: https://github.com/fluxcd/agent-skills/pull/13

Using `git commit -sm "<msg>" --trailer "Assisted-by: <agent>/<model>"` makes GitHub display it in the PR body automatically. No need to write it by hand in the PR 🎉

stealthybox: > Also we'll have an AGENTS.md file that will instruct the AI to add the Assisted-by, zero effort from the contributor.

notably, many models incorrectly report which model they are, so we may get some lossy info here.

stealthybox (approved):
lgtm

Noting that our stance on "assisted-by" commit messages differs from Kubernetes.
I'm okay with this.

Note that agents following AGENTS.md may incorrectly report their own model on commits:
https://github.com/fluxcd/flux2/pull/5841#discussion_r3068972822

CONTRIBUTING.md
stefanprodan: Yeah it's best effort. I tested the [skill](https://github.com/fluxcd/agent-skills/pull/13) with Claude Code and it works.

stefanprodan:
> Noting that our stance on "assisted-by" commit messages differs from Kubernetes.

The Kubernetes team has seen people abusing this tag for marketing reasons. As in: they fix a typo, the PR gets merged, then they use the commit to advertise their shitty AI tool. But this can happen with PRs as well.

If during the review we detect this behaviour, then we should close the PR and ban the user from the FluxCD org.

PR #5845, opened 2026-04-12
Add `--ignore-not-found` to `flux diff ks`

### LINKED ISSUE #5439: [Feature] Allow `flux diff ks` when the `Kustomization` does not already exist

rycli:
Closes: https://github.com/fluxcd/flux2/issues/5439

@stefanprodan happy to look at this, since I've been poking around this codebase neighborhood ;-)


cmd/flux/diff_kustomization_test.go
```diff
@@ -138,6 +138,48 @@ func TestDiffKustomization(t *testing.T) {
 	}
 }
 
+// TestDiffKustomizationNotDeployed tests `flux diff ks` when the Kustomization
+// CR does not exist in the cluster but is provided via --kustomization-file.
+// Reproduces https://github.com/fluxcd/flux2/issues/5439
+func TestDiffKustomizationNotDeployed(t *testing.T) {
```
stefanprodan, line 144: Assuming some of the objects already exists in the cluster, and we are moving them to Flux, will the diff show them? I think a valid test would be having an existing configmap in-cluster and a new configmap in git. The diff should show the labels added to existing one and the new configmap as created.

rycli: Added a test; this is indeed already picked up 🙂

One thing I did spot: the `kustomize.toolkit.fluxcd.io/namespace` value is empty (see the 'golden file' output I added); which makes sense given the namespace is not included in the test manifest. We could reassign the 'local' Kustomization's namespace to default, since you'll always have one; but it's not _really_ guaranteed, depending on how you deploy / apply overlays.

If we're going for correctness, I'd just leave it as-is, since it's technically right that there's no known Kustomization namespace.

stefanprodan: Another test we could add is new namespace + configmap, the CM should show created and not error out IMO

rycli: Just to clarify: what exactly would we be testing in that case?

To expand on what I ran into: if a local Kustomization file is diff'd, and the resources it would contain/start managing already exist on the cluster, these are correctly shown as only having the Flux labels added (`kustomize.toolkit.fluxcd.io/name` and `kustomize.toolkit.fluxcd.io/namespace:`). Since these labels refer to the "parent" Kustomization that (would) manage them, they're determined based on the values of the local Kustomization at that moment.

However, if the Kustomization manifest doesn't specify the namespace of the Kustomization itself (as is the case in a lot of the tests, and could be the case if namespaces are overwritten in e.g., a `kustomization.yaml` file ), we can't determine what it would actually be (and therefore the label value is an empty string). So no actual errors here, and arguably intended behavior.

stefanprodan: Sorry for the confusion. What I'm trying to determine is what happens if the new Kustomization (that doesn't exist in the cluster) creates a namespace and objects inside the namespace. Assuming we have a namespace.yaml and configmap.yaml, would the diff say `namespace x created` and `configmap y created`?

rycli: Added some more tests for this; diffing a 'create namespace' alone works, 'create namespace+configmap' fails with `ConfigMap/new-ns/app-config not found: namespaces "new-ns" not found`. Makes sense if we're using SSA on the individual resources.

This is a bit outside of the area of the codebase I'm familiar with... the behaviour seems logical/correct to me given the current implementation, but what's the desired output? Namespaces would need to get special handling to have this work.

stefanprodan: We should follow `kubectl diff --server-side` as close as possible. But we can leave this for another PR.

rycli: As far as I can tell, this is also consistent with the `kubectl diff` behaviour... but maybe indeed out of scope for this issue. I'll tidy up the PR for now :-)

rycli:
@stefanprodan I've squashed the commits; kept the namespace-related tests for future reference, but can also drop them if preferred.

I'm not particularly inspired by `--allow-new-kustomization` as a flag name, so if you have any ideas there...

cmd/flux/diff_kustomization.go
```diff
@@ -78,6 +79,8 @@ func init() {
 	diffKsCmd.Flags().StringToStringVar(&diffKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
 	diffKsCmd.Flags().BoolVar(&diffKsArgs.inMemoryBuild, "in-memory-build", true,
 		"Use in-memory filesystem during build.")
+	diffKsCmd.Flags().BoolVar(&diffKsArgs.allowNewKustomization, "allow-new-kustomization", false,
```
stefanprodan: Maybe `--ignore-not-found` would be better?

rycli: Renamed!

stefanprodan (approved):
LGTM

Thanks @rycli 🏅

Issue #4075, opened 2023-07-15
HelmRelease should support SOPS encryption directly


SebTardif:
### Describe the bug

Even if we have workaround described here -> https://github.com/fluxcd/flux2/discussions/1537

Users of SOPS are used to not have to change file format, or move part of configurations to be able to encrypt the few things that need encryption.

It should be straightforward to support that HelmRelease be decrypted on the fly if it's encrypted.

### Steps to reproduce

When assuming this is already implemented, the error is a little vague, it's saying: dry-run failed, error: failed to create typed patch object (...; helm.toolkit.fluxcd.io/v2beta1, Kind=HelmRelease

### Expected behavior

SOPS should just work, without having user to be confused by error message, had to Google, then implement workarounds, splitting a consistent and coherent YAML to many.

### OS / Distro

Linux

### Flux version

flux: v0.41.2

### Flux check

► checking prerequisites
✗ flux 0.41.2 <2.0.1 (new version is available, please upgrade)
✔ Kubernetes 1.27.1 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.31.1
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.35.0
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.33.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.36.0
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta2
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1beta2
✔ helmcharts.source.toolkit.fluxcd.io/v1beta2
✔ helmreleases.helm.toolkit.fluxcd.io/v2beta1
✔ helmrepositories.source.toolkit.fluxcd.io/v1beta2
✔ kustomizations.kustomize.toolkit.fluxcd.io/v1beta2
✔ ocirepositories.source.toolkit.fluxcd.io/v1beta2
✔ providers.notification.toolkit.fluxcd.io/v1beta2
✔ receivers.notification.toolkit.fluxcd.io/v1beta2
✔ all checks passed

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kingdonb:
[...]

PR #5869, opened 2026-04-27
build/diff: strip SOPS metadata from non-Secret resources

SebTardif:
## Summary

This PR makes `flux build kustomization` and `flux diff kustomization` safer and more compatible with SOPS-encrypted non-Secret manifests (e.g. `HelmRelease`, `ConfigMap`) by stripping top-level `.sops` metadata from non-Secret resources in CLI build output.

## Why

When `.sops` metadata is left in non-Secret resources, server-side apply dry-run can fail with schema validation errors on CRDs (e.g. `.sops: field not declared in schema`).

## Changes

- Extend `maskSopsData` in `internal/build/build.go` to clear top-level `.sops` for non-Secret resources when detected.
- Keep existing behavior for `Secret` resources.
- Preserve `ENC[...]` ciphertext values (no decryption performed in CLI build step).
- Add unit coverage in `internal/build/build_test.go`.
- Add build command tests and fixtures in `cmd/flux/build_kustomization_test.go` and `cmd/flux/testdata/build-kustomization/*`.
- Add `flux create kustomization` decryption golden coverage in `cmd/flux/create_kustomization_test.go`.

## Validation

- `make fmt vet`
- `make test`

Related context: https://github.com/fluxcd/flux2/issues/4075

### DIFF
--- a/cmd/flux/build_kustomization.go
+++ b/cmd/flux/build_kustomization.go
@@ -68,6 +68,7 @@ type buildKsFlags struct {
 	kustomizationFile string
 	path              string
 	ignorePaths       []string
+	stripSopsMetadata bool
 	dryRun            bool
 	strictSubst       bool
 	recursive         bool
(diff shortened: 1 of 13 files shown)


matheuscscp:
Hmmmm encrypting non-Secret resources is so bad...

SebTardif:
@matheuscscp > Hmmmm encrypting non-Secret resources is so bad...
"so bad..." is way too strong
It's all about tradeoff. In practice, if credentials showing up in values of helm chart can be encrypted more easily, then overall things are more secure. We covered some of that in #4075. It seems [stefanprodan](https://github.com/stefanprodan) was open to the idea if he didn't have to code it himself...

I worked in many projects were the credentials are showing up in git and logs. I want to make it easy to avoid that. Ending up with credentials hidden inside Kubernetes that is not in a secret and that nobody has practical access, is something my projects accept totally.

stefanprodan:
I'm OK with this change if we put it behind an opt-in flag like `--strip-sops-metadata`. While diff will no longer error out at dry-run, it will always detect drift, which I guess is fine.

SebTardif:
Implemented follow-up in ce8d6384:

- Added opt-in `--strip-sops-metadata` to both `flux build kustomization` and `flux diff kustomization`.
- Kept default behavior unchanged: non-Secret top-level `.sops` metadata is now preserved unless the flag is set.
- Preserved existing Secret masking behavior.
- Added/updated unit coverage for both opt-in and default paths.

Also verified the two Copilot findings are handled:
- structural lookup for top-level `sops` field
- proper error propagation from `PipeE`

PR #5881, opened 2026-05-06
include source-watcher in install.yaml manifests

tmmorin:
The source-watcher controller is currently not included in the manifest.yaml which is part of a release bundle.

Example: https://github.com/fluxcd/flux2/releases/download/v2.8.6/install.yaml

This PR updates the `./manifests/install` kustomization to include the source-watcher, so that an installation done with the `install.yaml` which is part of a release will include this controller.

### DIFF
--- a/manifests/install/kustomization.yaml
+++ b/manifests/install/kustomization.yaml
@@ -9,6 +9,7 @@ resources:
   - ../bases/helm-controller
   - ../bases/image-reflector-controller
   - ../bases/image-automation-controller
+  - ../bases/source-watcher
   - ../rbac
   - ../policies
 transformers:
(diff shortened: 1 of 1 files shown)


matheuscscp:
In `flux bootstrap` or `flux install` you need to use `--components-extra=source-watcher`.

tmmorin:
> In `flux bootstrap` or `flux install` you need to use `--components-extra=source-watcher`.

Our use case is that we directly consume the manifests from a kustomization.yaml (used by a Kustomization to let flux self-manage itself):

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - https://github.com/fluxcd/flux2/releases/download/v2.8.6/install.yaml
```

I honestly thought that it was a simple omission to not have this specific source-watcher controller as part of the `install.yaml`, because it's the only one missing. 

Also, the controllers which can be enabled with  `flux (bootstrap|install) --components-extra=...` seem to be all already part of `install.yaml`, except the `source-watcher`.

matheuscscp:
You have a good point!

matheuscscp (approved):
LGTM! 🚀 

Thanks for the fix, @tmmorin!

PR #5885, opened 2026-05-10
Add --add-ignore-paths flag to artifact commands

### LINKED ISSUE #5144: `flux push artifact --ignore-paths` should be added on top of the default
### Describe the bug

The default ignore paths are very useful for most cases:

https://github.com/fluxcd/flux2/blob/0cc68e0521dfcee719aad5866aff6c10e0eb45c6/cmd/flux/build_artifact.go#L56

```
.git/,.gitignore,.gitmodules,.gitattributes,*.jpg,*.jpeg,*.gif,*.png,*.wmv,*.flv,*.tar.gz,*.zip
```

I don't want to re-specify all of them when using the `--ignore-paths`, I want to add more ignore paths on top of them.

Not using these defaults should must still be supported somehow, but adding paths on top of the defaults would be very useful.

### Steps to reproduce

`flux push artifact --ignore-paths=some/path ...`

### Expected behavior

I don't want to re-specify all of them when using the `--ignore-paths`, I want to add more ignore paths on top of them.

_No response_

### OS / Distro

N/A

### Flux version

v2.4.0

### Flux check

N/A

### Git provider

_No response_

_No response_

### Additional context

_No response_

### Code of Conduct

ruslan-shaydullin:
Closes #5144

## What

Adds `--add-ignore-paths` to `flux build artifact`, `flux push artifact`,
and `flux diff artifact`. It appends user-provided patterns to whatever
ignore set is currently in effect, instead of forcing a full replacement
like `--ignore-paths` does.

## How it composes

- `--ignore-paths X` → X (replace, existing behaviour, unchanged)
- `--add-ignore-paths Y` → defaults + Y
- both → X + Y (no defaults)

## Tests

Unit tests for `composeIgnorePaths` covering empty input, ignore-paths
only, add-ignore-paths only, both flags combined, duplicates retained,
and the defaults+add common case.

## Local testing notes

On Apple Silicon the Makefile's `setup-envtest use latest --arch=amd64`
step fails (`bad CPU type in executable`). I ran envtest manually with
`--arch=arm64` and pointed `KUBEBUILDER_ASSETS` at the arm64 dir; the
full `cmd/flux/...` test suite passes. CI on linux/amd64 should be
unaffected.

### DIFF
--- a/cmd/flux/build_artifact.go
+++ b/cmd/flux/build_artifact.go
@@ -52,10 +52,27 @@ type buildArtifactFlags struct {
 	output          string
 	path            string
 	ignorePaths     []string
+	addIgnorePaths  []string
 	resolveSymlinks bool
 }
 
 var excludeOCI = append(strings.Split(sourceignore.ExcludeVCS, ","), strings.Split(sourceignore.ExcludeExt, ",")...)
+
+// composeIgnorePaths returns the concatenation of two ignore-path slices,
+// preserving order. It is used by build, push and diff artifact commands
+// to combine --ignore-paths (which replaces the defaults) with
+// --add-ignore-paths (which appends to whatever ignore set is currently
+// in effect). Duplicate patterns are passed through unchanged since the
+// underlying gitignore matcher treats redundant patterns as no-ops.
+func composeIgnorePaths(ignorePaths, addIgnorePaths []string) []string {
+	if len(ignorePaths) == 0 && len(addIgnorePaths) == 0 {
+		return nil
+	}
+	result := make([]string, 0, len(ignorePaths)+len(addIgnorePaths))
+	result = append(result, ignorePaths...)
+	result = append(result, addIgnorePaths...)
+	return result
+}
 
 var buildArtifactArgs buildArtifactFlags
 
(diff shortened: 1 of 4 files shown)


ruslan-shaydullin:
Hi @matheuscscp — quick ping on this. It implements the additive flag from #5144 ("We could have a different flag that uses the defaults as well"): `--add-ignore-paths` appends to the defaults rather than replacing them like `--ignore-paths`, so `--ignore-paths` behaviour is unchanged.

Scope note: the issue is about `push artifact`, but I applied the flag to `build`, `push`, and `diff artifact` too, since they share the same ignore-path handling. Happy to narrow it to just `push` if you'd prefer to keep it minimal.

matheuscscp:
@ruslan-shaydullin Looking good, please rebase and force-push (no merge commits!) :pray:

ruslan-shaydullin:
Done! Rebased onto main, clean history. Thanks for taking a look 🙏

matheuscscp (approved):
LGTM! 🚀

matheuscscp:
Thanks @ruslan-shaydullin! Are you on hurry to have this feature? Technically it only ships in Flux 2.10 mid Q4...

ruslan-shaydullin:
No hurry at all. Thanks for merging!

PR #5892, opened 2026-05-12
Add CLI support for OCIRepository.spec.layerSelector in flux create source oci

dme86:
This changes the CLI to use a single ``--layer-selector`` flag for
  ``flux create source oci``.

  The flag encodes both layer media type **and** operation in the form
  ``<media-type>:<operation>``.

  The operation is validated against the supported API values:

  - extract
  - copy

  When provided, the command renders ``spec.layerSelector`` in the exported
  OCIRepository.

  ### Testing
```
  go test ./cmd/flux/... --tags=unit -run TestCreateSourceOCI -v
```
  ### Notes

  This only adds CLI support for fields already supported by the
  OCIRepository API and source-controller. It does not change controller
  behavior or the API.

### DIFF
--- a/cmd/flux/create_source_oci.go
+++ b/cmd/flux/create_source_oci.go
@@ -53,6 +53,12 @@ var createSourceOCIRepositoryCmd = &cobra.Command{
     --verify-provider=cosign \
     --verify-subject="^https://github.com/stefanprodan/podinfo/.github/workflows/release.yml@refs/tags/6.6.2$" \
     --verify-issuer="^https://token.actions.githubusercontent.com$"
+
+  # Create an OCIRepository for a Helm chart layer
+  flux create source oci valkey-cluster \
+    --url=oci://example.com/charts/valkey \
+    --tag=0.11.6 \
+    --layer-selector=application/vnd.cncf.helm.chart.content.v1.tar+gzip:copy
 `,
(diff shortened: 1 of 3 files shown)


matheuscscp:
I'd rather have a single flag `--later-selector` that encodes both media type and operation in a single string, CLIs should not be verbose

dme86:
@matheuscscp
> I'd rather have a single flag `--later-selector` that encodes both media type and operation in a single string, CLIs should not be verbose

Thanks, that makes sense! 
I updated the CLI to use a single ``--layer-selector=<media-type>:<operation>`` flag as you suggested, force-pushed the branch, and revised the PR description to match the new shape.

matheuscscp:
Is the media type guaranteed not to have `:`?

dme86:
@matheuscscp
> Is the media type guaranteed not to have `:`?

OCI media types follow MIME syntax, so ``:`` is safe here.

- https://datatracker.ietf.org/doc/html/rfc6838#section-4.2

OCI also refers back to RFC 6838 for valid media types:
- https://github.com/opencontainers/image-spec/blob/main/manifest.md

matheuscscp (approved):
LGTM! 🚀 

Thanks, @dme86!

matheuscscp:
@dme86 Can you please rebase and force-push for the merge? :pray:

PR #5909, opened 2026-05-22
Validate Helm source URL schemes

immanuwell:
Reject unsupported URL schemes in `flux create source helm` before the command generates or applies a HelmRepository.

Repro before this patch:

```sh
flux create source helm podinfo --url=git://example.com/charts --export
```

That printed a HelmRepository with `spec.url: git://...`, but the CRD only accepts `http`, `https`, and `oci` URLs. Bit of a papercut, easy to hit when using `--export`.

Now the command fails early with a clear error and keeps the existing `https` and `oci` export paths working.

Tests:

```sh
make tidy fmt vet
make test TEST_ARGS='-run TestCreateSourceHelm -count=1'
```

### DIFF
--- a/cmd/flux/create_source_helm.go
+++ b/cmd/flux/create_source_helm.go
@@ -114,8 +114,15 @@ func createSourceHelmCmdRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @immanuwell

PR #5920, opened 2026-05-29
Allow signing commits using SSH key

hiddeco:
Depends on: 

- https://github.com/fluxcd/pkg/pull/1141
- https://github.com/fluxcd/pkg/pull/1222
- https://github.com/fluxcd/image-automation-controller/pull/1035


stefanprodan:
@hiddeco can you please incorporate these changes in this PR https://github.com/fluxcd/flux2/pull/5930/changes

matheuscscp:
@hiddeco We start the controller releases on Monday, I'd love to get this PR in before that if possible :pray:

hiddeco:
@stefanprodan @matheuscscp https://github.com/fluxcd/image-automation-controller/pull/1035 and this PR have both been updated.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/fluxcd/cli-utils v1.2.1
 	github.com/fluxcd/go-git-providers v0.26.0
 	github.com/fluxcd/helm-controller/api v1.5.5
-	github.com/fluxcd/image-automation-controller/api v1.1.4
```
hiddeco: This has to be updated once https://github.com/fluxcd/image-automation-controller/pull/1035 ends up in a release.

matheuscscp (approved):
LGTM! 🚀

matheuscscp:
@hiddeco Codex just found a couple of nits:

  1. pkg/bootstrap/bootstrap_plain_git.go:576 can still panic for a public-only GPG keyring when no --gpg-key-id is supplied. The keyed branch checks entity.PrivateKey == nil, but the default branch selects
     keyRing[0] and then immediately calls entity.PrivateKey.Decrypt(...) at line 579. This should return an actionable error instead.

  2. cmd/flux/bootstrap.go:224 makes --ssh-signing-reuse-private-key require --private-key-file before GitHub/Gitea get a chance to reject the flag as unsupported. For flux bootstrap github --ssh-signing-
     reuse-private-key, the user first sees “requires --private-key-file”, even though that subcommand cannot reuse a supplied transport key at all. The command-specific unsupported error should win for
     those providers.

hiddeco:
@matheuscscp good shout, will address them in a bit.

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #5927, opened 2026-06-04
Validate plugin binary path

stefanprodan:
Reject plugin manifests whose bin field is empty, path-like, or missing the flux- prefix before installing. This prevents downloaded plugin artifacts from being written outside the plugin directory.

Thanks to **Shivam Kumar** for reporting this path traversal vulnerability before the release of the Flux plugin system.


matheuscscp (approved):
LGTM

Issue #5931, opened 2026-06-08
host architecture for local builds/envtest is amd64 on arm machines


stealthybox:
Local builds and tests across the Flux repos are hard-coded to `amd64`, which forces emulation/rosetta on apple silicon and other arm64 hosts.

- **`ENVTEST_ARCH`** pinned to `amd64` (some repos also have a Darwin override), so `setup-envtest` downloads amd64 test binaries and runs them under emulation.
- **`BUILD_PLATFORMS`** pinned to `amd64` (or a multi-arch list), so `make docker-build` produces a non-native image locally.

envtest now publishes native arm64 binaries (including `darwin/arm64`), so the amd64 pin and the Darwin override are no longer needed.

We should detect the local arch from the host go toolchain (`go env GOARCH`):
- set `ENVTEST_ARCH` to the host arch to speed up `make test`
- set `BUILD_PLATFORMS` to `linux/<host arch>` for `make docker-build`

overridable for cross-arch builds should still work the same
```
make docker-build BUILD_PLATFORMS=linux/amd64,linux/arm64,linux/arm/v7
```

multi-arch, release images are built by the `fluxcd/gha-workflows` release workflow,
not by `make docker-build`, so release artifacts are unaffected.

## Affected repos

- fluxcd/flux2
- fluxcd/source-controller
- fluxcd/kustomize-controller
- fluxcd/helm-controller
- fluxcd/notification-controller
- fluxcd/image-automation-controller
- fluxcd/image-reflector-controller
- fluxcd/source-watcher
- fluxcd/pkg

stealthybox:
notably, when I switched to using arm64 envtest, my local cache of the binary missed, and setup-envtest had to fetch from github.

since github was having issues this morning for me, their releases CDN was issuing 504's and failing my local tests despite the failing link working in my browser session.

it eventually just worked after a few minutes of confusion.

PR #5932, opened 2026-06-08
build: target host arch for local builds/envtest

stealthybox:
build fix. `ENVTEST_ARCH` was hardcoded to `amd64`, which forces emulation for the test binaries on arm64 machines (apple silicon).

envtest publishes native arm64 binaries now (including darwin/arm64), so the pin and Darwin override is no longer needed.

I ran all tests and docker-builds across the 9 repos via apple silicon

tracking: https://github.com/fluxcd/flux2/issues/5931


stefanprodan (approved):
LGTM

PR #5952, opened 2026-06-24
cmd: support `type!=status` in get --status-selector

### LINKED ISSUE #5951: Feature: Add a "not equals" condition to --status-selector
Currently, I have to run this to find all objects where Ready is not True:
```
flux get all -A --status-selector Ready=False && flux get all -A --status-selector Ready=Unknown
```
It would be nicer if I could run:
```
flux get all -A --status-selector Ready!=True
```

3uzbcqje:
Closes: #5951

`flux get --status-selector` only supported equality (`type=status`), so finding objects that are not in a given state required multiple invocations, e.g. listing everything that is not ready needed both `Ready=False` and `Ready=Unknown`.

Add support for a negated selector `type!=status`. Since all resource adapters delegate matching to the shared `statusMatches` helper and filtering is centralised in `getRowsToPrint`, negation is implemented purely in the parse/filter layer by inverting the match result. This covers every resource type and the `--watch` path without touching the per-resource adapters.

A missing condition is treated as not-matching by `statusMatches` (Flux considers it "waiting to be reconciled"), so `Ready!=True` also surfaces objects that have no Ready condition yet, i.e. the complete not-ready set:

    flux get all -A --status-selector Ready!=True

### DIFF
--- a/cmd/flux/get.go
+++ b/cmd/flux/get.go
@@ -78,7 +78,7 @@ func init() {
 	getCmd.PersistentFlags().BoolVarP(&getArgs.noHeader, "no-header", "", false, "skip the header when printing the results")
 	getCmd.PersistentFlags().BoolVarP(&getArgs.watch, "watch", "w", false, "After listing/getting the requested object, watch for changes.")
 	getCmd.PersistentFlags().StringVar(&getArgs.statusSelector, "status-selector", "",
-		"specify the status condition name and the desired state to filter the get result, e.g. ready=false")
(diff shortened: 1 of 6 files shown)


matheuscscp (approved):
LGTM

(Correct is not always the best, pls ask the LLM to simplify the code as well. This `noFilter` var was already hard to read, after the LLM patch it was even harder)

3uzbcqje:
I'll do that next time. Thank you!

PR #5955, opened 2026-06-25
feat: Install Plugins alongside Flux setup in gh actions

### LINKED ISSUE #5933: Add plugins to Flux GitHub Action
After Flux 2.9 release we can allow users to install Flux plugins on GH runners with:

```yaml
steps:
  - name: Setup Flux CLI
    uses: fluxcd/flux2/action@main
    with:
      plugins: |
        mirror
        schema
        operator
```

gat786:
This PR adds the capability of installing plugins to the action which sets up Flux on a GH Action.

Closes: #5933

Adds two new variables to the GH Action input 

* plugins: [optional] Plugins to install alongside Flux. One per line: `<name>@<version|digest>`.
  Example: `mirror@0.0.1`
* plugin_dir: [optional] Directory where the plugins should be installed, setups the `FLUXCD_PLUGINS` env variable. 

And outputs 

* `plugin-dir` - Directory where the plugin binaries are present. This helps in users know the directory even when it was unspecified and they can use this value for further processing if need be.

the plugin support comes with a feature gate which is designed to exit with an error when plugins were specified but the flux binary installed does not support plugins yet (i.e. it has a version `< 2.9.0`.

### DIFF
--- a/action/action.yml
+++ b/action/action.yml
@@ -212,10 +212,9 @@ runs:
       env:
         PLUGINS: ${{ inputs.plugins }}
       run: |
-        # Read line by line
         echo "$PLUGINS" | while read -r PLUGIN; do
-          # if $PLUGIN contains some string value and is not empty
-          if [[ -n $PLUGIN ]] || [[ $PLUGIN ]]; then
+          trimmed="${PLUGIN//[[:space:]]/}"
+          if [[ -n "$trimmed" ]]; then
             echo Installing: "$PLUGIN"
             flux plugin install "$PLUGIN"
           fi
--- a/action/action.yml
+++ b/action/action.yml
@@ -12,17 +12,32 @@ inputs:
     description: "arch can be amd64, arm64 or arm"
     required: false
     deprecationMessage: "No longer required, action will now detect runner arch."
+  plugins:
+    description: >
+      Plugins to install alongside Flux. One per line: `<name>@<version|digest>`. Example: `mirror@0.0.1`
+    required: false
+  plugin-dir:
+    description: >
+      Directory where the plugins should be installed, setups the `FLUXCD_PLUGINS` env variable.
+      defaults to be installed alongside flux installation directory in $RUNNER_TOOL_CACHE in a child dir named `/plugins`
+    required: false
   bindir:
     description: "Alternative location for the Flux binary, defaults to path relative to $RUNNER_TOOL_CACHE."
     required: false
   token:
     description: "Token used to authenticate against the GitHub.com API."
     required: false
+outputs:
+  plugin-dir:
+    description: >
+      Directory where the plugins were installed by the action.
+    value: ${{ steps.set-outputs.outputs.plugin-dir }}
 runs:
   using: composite
   steps:
     - name: "Download the binary to the runner's cache dir"
       shell: bash
+      id: setup-flux-bin
       env:
         VERSION: "${{ inputs.version }}"
         FLUX_TOOL_DIR: "${{ inputs.bindir }}"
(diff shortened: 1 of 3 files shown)


gat786:
I did run some tests here -> https://github.com/gat786/flux2/actions/runs/28187933002

if it helps with anything!

matheuscscp (commented):
Hey @gat786, thanks very much for this PR!

We need to add an `if:` field to the plugin step so it runs only if `inputs.plugins` is specified.

Also, let's rename `plugin_dir` to `plugin-dir`, which is the GHA convention. We should also define a sensible default for this input, as people will probably not have an opinion about it. Finally, the effective `plugin-dir` should be an output of this action, because even though people will not want to think about it to choose one, they will still need the default directory chosen by the action e.g. for the `mirror` plugin, which for some commands expects the config file to be a sibling of the binary in the filesystem.

matheuscscp:
Hey @gat786 please squash, rebase and force-push, this PR should have a single commit :pray:

gat786:
@matheuscscp after a few back and forths me donez!

matheuscscp:
Looks like the `if:` blocks are wrong, I think they don't need the `${{` and `}}`

gat786:
oh no, lemme check

gat786:
it is interesting that they do not allow using `{{`, as far as i remember it used to be a norm. In any case, I fixed that issue along with it I separated the code out a bit and tested it to make sure it doesn't break this time.

matheuscscp:
Hey @gat786 this feature needs a very important gate: only the Flux CLI v2.9+ will support it. I think we need to detect the version and check if 2.9+ is satisfied, in order to print a nice error message about this requirement not being met, and that also shows the installed version.

matheuscscp:
Also wondering why are the other env vars needed i.e. those beyond `FLUXCD_PLUGINS`

gat786:
Hey the FLUXCD_PLUGINS is constructed using those other variables and when the other variables are not exported the construction does not happen properly, i.e. the os, arch and version entries remain empty. 

I can remove the other exports by constructing fluxcd plugins directory in the same step as installation but since installation of plugins is going to be condition based I had to do it in its own step and for that exports need to happen.

matheuscscp:
I think we can use step inputs/outputs for this, and not export envs. This way it stays internal.

gat786:
Hey I added a feature gate and did some testing on it as well across platforms it does seem to work according to me, please check whenever you can!

action/action.yml
```diff
@@ -146,3 +164,63 @@ runs:
...
+      id: plugin-support-gate
+      shell: bash
+      if: inputs.plugins != ''
+      env:
+        V1: ${{ steps.setup-flux-bin.outputs.installed-flux-version }}
+      run: |
+        if [ "$(printf '%s\n%s' "$V1" "2.9.0" | sort -V | tail -n1)" = "$V1" ]; then
+          echo plugin-support="yes" >> $GITHUB_OUTPUT
+        else
+          echo plugin-support="no" >> $GITHUB_OUTPUT
```
matheuscscp: I think this gate should error out hard instead of skipping plugin installation, it's the whole point. Otherwise the step will show "green" and it will be hard to understand why my plugins are not working if the installation step succeeded. This feature is opt-in by specifying the `plugins` field, so erroring out is not gonna break anybody that isn't opting in.

gat786: Makes complete sense, I added an exit 1, when plugin support is not available and it will now exit with a failure.

matheuscscp (approved):
LGTM! 🚀 

@gat786 Thanks very much for the contribution!

gat786:
Thank you so much!

gat786:
I would like to participate even more in the upcoming future! Learnt a lot from this implementation!

PR #5962, opened 2026-06-30
Upgrade go-git-providers to 0.29.0

matheuscscp:


matheuscscp:
We will do this for Flux 2.10

PR #5973, opened 2026-07-03
fix(goreleaser): make Homebrew formula pass brew audit

trademomentumllc:
Fixes two common `brew audit --strict` failures in the generated Homebrew formula:

- Description no longer starts with the formula name.
- Test `system` call uses separate arguments instead of a shell string.

This ensures future releases of Flux publish a compliant formula to fluxcd/homebrew-tap.

Related to fluxcd/homebrew-tap#17.


stefanprodan:
@trademomentumllc please signoff your commit, see: https://github.com/fluxcd/flux2/pull/5973/checks?check_run_id=85051958719

Issue #5968, opened 2026-07-01
Bug: `flux build kustomization` now defaults to in-memory build


matheuscscp:
The in-memory build locks the file system at `$PWD`:

```bash
  flux build kustomization app \
    --path=/tmp/flux-example/app \
    --kustomization-file=/tmp/flux-example/app.yaml \
    --namespace=default \
    --dry-run
```

Fails with:

```
  fs-security-constraint abs /tmp/flux-example/app:
  path '/tmp/flux-example/app' is not in or below '/builds/example/repo'
```

Workaround: `--in-memory-build=false`

This is a breaking change introduced here: https://github.com/fluxcd/flux2/pull/5794

stefanprodan:
I think we should lock at `--path` level.

rycli:
This is an (admittedly unintended) side effect of extending the securefs “backend” when creating the “memory” backend. We could switch it out to simply be a separate backend altogether; keeps the benefit of the in-memory building without this regression. I’ll need to quickly jump into the code again, but from memory it was quite easily “composable”.

I’d be happy to look at patching this, or implementing the `—path` lock as suggested by @stefanprodan, if you’d consider keeping it! And add a regression test in any case.

I’ve been running the memory building mode for a few months; the memory build as default has been great for our developers to avoid explicitly needing to understand the multiple issues highlighted in the original PR, in order to then switch to the patched build… and there’s a nice performance boost 😇

rycli:
Specifically, this regression is caused by this default; https://github.com/fluxcd/flux2/blob/006ecb6b0401525e703585df9e87d6d0d2ed34bd/internal/build/build.go#L114

Switching to the non-secure filesystem will likely resolve it. I can prep and test a fix in the coming couple days?

matheuscscp:
@rycli I see no reason to change the default behavior here. Locking the root to a path or not, I imagine in-memory build can still be a problem for large builds that don't entirely fit in memory. We already reverted the default to the previous behavior here: #5969

If we want to address the concern about the path where we lock the root we need to do it ASAP in the same / next Flux patch release, as we can still frame it as a bug. If we take too long the current behavior will become expected behavior and we will miss the opportunity to change it as a bugfix.

Stefan is also concerned that Windows users will have problems using this flag. I'm not sure what that means but it seems important to consider.

Can you please look after this ASAP for the next patch? It might as well happen early next week.

rycli:
[...]

PR #5974, opened 2026-07-04
Fix `--in-memory-build` restricting disk access

rycli:
Addresses https://github.com/fluxcd/flux2/issues/5968

The PR that introduced the bug https://github.com/fluxcd/flux2/pull/5794 implemented a new `--in-memory-build` option; it wraps a 'filesystem' instance, specifically the `buildfs.MakeFsOnDiskSecure`. This restricts access to files in PWD.

The fix is pretty straightforward; use the unrestricted `MakeFsOnDisk` option. I've also updated one of the tests accordingly, as it wasn't testing the desired behavior.

---

Side-note: as far as I can see, the `MakeFsOnDiskSecure` logic is only implemented for the `flux bootstrap` command, which is likely why this issue didn't occur previously 🤔

### DIFF
--- a/cmd/flux/build_kustomization.go
+++ b/cmd/flux/build_kustomization.go
@@ -86,7 +86,7 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	buildKsCmd.Flags().BoolVarP(&buildKsArgs.recursive, "recursive", "r", false, "Recursively build Kustomizations")
 	buildKsCmd.Flags().StringToStringVar(&buildKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
-	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", false,
+	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", true,
 		"Use in-memory filesystem during build.")
 	buildCmd.AddCommand(buildKsCmd)
 }
--- a/cmd/flux/diff_kustomization.go
+++ b/cmd/flux/diff_kustomization.go
@@ -77,7 +77,7 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	diffKsCmd.Flags().BoolVarP(&diffKsArgs.recursive, "recursive", "r", false, "Recursively diff Kustomizations")
 	diffKsCmd.Flags().StringToStringVar(&diffKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
-	diffKsCmd.Flags().BoolVar(&diffKsArgs.inMemoryBuild, "in-memory-build", false,
+	diffKsCmd.Flags().BoolVar(&diffKsArgs.inMemoryBuild, "in-memory-build", true,
 		"Use in-memory filesystem during build.")
 	diffKsCmd.Flags().BoolVar(&diffKsArgs.ignoreNotFound, "ignore-not-found", false,
 		"Ignore Kustomization not found errors on the cluster when diffing.")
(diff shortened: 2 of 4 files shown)


internal/build/build.go
```diff
@@ -106,16 +106,7 @@ func (inMemoryFsBackend) Generate(gen *kustomize.Generator, dirPath string) (fil
...
-	cwd, err := os.Getwd()
-	if err != nil {
-		return nil, "", action, fmt.Errorf("failed to get working directory: %w", err)
-	}
-
-	diskFS, err := buildfs.MakeFsOnDiskSecure(cwd)
-	if err != nil {
-		return nil, "", action, fmt.Errorf("failed to create secure filesystem: %w", err)
-	}
-	fs := buildfs.MakeFsInMemory(diskFS)
```
matheuscscp, line 118: Was this never in scope in the original issue? Why are we switching from secure to non-secure? Was this shipped entirely unintentionally?

cmd/flux/build_kustomization.go
```diff
@@ -86,7 +86,7 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	buildKsCmd.Flags().BoolVarP(&buildKsArgs.recursive, "recursive", "r", false, "Recursively build Kustomizations")
 	buildKsCmd.Flags().StringToStringVar(&buildKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
-	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", false,
+	buildKsCmd.Flags().BoolVar(&buildKsArgs.inMemoryBuild, "in-memory-build", true,
```
matheuscscp, line 89: You know, for as much as a I want Flux to have better defaults, we don't normally implement a new feature and make it replace a default behavior in Flux in the same release. See for example the `FluxStorage` feature gate in image-reflector-controller in Flux 2.9. It allows opting into a feature that we introduced and are planning to make the default in Flux 2.10 because we think it's pretty good and should become the default. I don't think this in-memory build should rush and skip that graduation phase.

cmd/flux/diff_kustomization.go
```diff
@@ -77,7 +77,7 @@ func init() {
 		"When enabled, the post build substitutions will fail if a var without a default value is declared in files but is missing from the input vars.")
 	diffKsCmd.Flags().BoolVarP(&diffKsArgs.recursive, "recursive", "r", false, "Recursively diff Kustomizations")
 	diffKsCmd.Flags().StringToStringVar(&diffKsArgs.localSources, "local-sources", nil, "Comma-separated list of repositories in format: Kind/namespace/name=path")
-	diffKsCmd.Flags().BoolVar(&diffKsArgs.inMemoryBuild, "in-memory-build", false,
+	diffKsCmd.Flags().BoolVar(&diffKsArgs.inMemoryBuild, "in-memory-build", true,
```
matheuscscp, line 80: Ditto

internal/build/build.go
rycli: The previous implementation of the file access logic did not in fact make use of the "secure layer". It came up when implementing the in-memory filesystem, and it was pointed by @stefanprodan that the underlying `pkg` implementation should respect/use the secure filesystem; see discussion on the corresponding PR https://github.com/fluxcd/pkg/pull/1153#discussion_r3031740146

The underlying topic of restricting the root/loading restrictions was explicitly discussed in https://github.com/fluxcd/flux2/pull/5794#discussion_r3006344085 . Looking back at it though, I think the error happened due to misunderstanding of the actual behaviour at the time; kustomization building did _not_ in fact respect the secure root previously, but we thought it did (https://github.com/fluxcd/flux2/pull/5794#discussion_r3006441768).

cmd/flux/build_kustomization.go
rycli: For what it's worth, it may be good to point out that this feature was originally reported & implemented as a bug fix; it addresses some real issues we ran into (see original related issue: https://github.com/fluxcd/flux2/issues/5781).

Making it a default was suggested by @stefanprodan in the follow-up discussion on the PR: https://github.com/fluxcd/flux2/pull/5794#discussion_r3066111220

Ultimately I believe this was a genuine bug, borne from an incorrect shared assumption, as opposed to any kind of intention to change user-facing behavior. I can switch it false, if that's the way forward though.

PR #5975, opened 2026-07-06
Add version-file input to the Flux CLI action

somaz94:
The Flux CLI action can now read the version from a file via a new `version-file` input, so the version can be pinned in a `.tool-versions` (asdf) or plain version file instead of hardcoded in the workflow. The input is ignored when `version` is set, so existing workflows are unaffected.

Validation:
- YAML lint of `action/action.yml` and the workflow: pass
- Local simulation of the resolution logic (plain file, `.tool-versions` with `flux`/`flux2` key, `v` prefix, missing-file and empty-file errors) behaves as expected
- Added a `version-file` CI job in `.github/workflows/action.yaml` exercising both fixture formats

AI assistance was used for this change, disclosed via the `Assisted-by` commit trailer.


.github/workflows/action.yaml
```diff
@@ -28,6 +28,33 @@ jobs:
...
+      - name: Setup flux from a .tool-versions file
+        uses: ./action
+        with:
+          version-file: action/testdata/.tool-versions
+      - name: Assert .tool-versions file
+        shell: bash
+        run: |
+          set -euo pipefail
+          flux -v
+          flux -v | grep -q '2.8.8'
```
stefanprodan: Let's make the `.tool-versions` install 2.9.2 and test here the version

somaz94:
Done — the `.tool-versions` fixture now installs `flux 2.9.2` (distinct from the plain `.flux-version` fixture's `2.8.8`), and the assert step checks for `2.9.2`. The differing versions prove the CLI version is resolved from the `.tool-versions` file rather than coincidentally matching a shared value. Pushed in ba9a4c6e.

stefanprodan:
Please squash the 2 commits and force push, your last commit has no signoff.

somaz94:
Squashed into a single signed-off commit and force-pushed — DCO should be green now. Regarding the `.tool-versions` version: the fixture already installs a distinct `flux 2.9.2` (vs `2.8.8` in the plain `.flux-version` fixture), and the CI job asserts it with `flux -v | grep -q '2.9.2'`, so the resolved version is verified end-to-end. Thanks for the review!

stefanprodan (approved):
LGTM

Thanks @somaz94

Issue #5879, opened 2026-05-05
Extend DependencyReference to support arbitrary Kubernetes resource dependencies


vecil:
### Motivation

This was first raised in [fluxcd/kustomize-controller#242](https://github.com/fluxcd/kustomize-controller/issues/242) and gained much more traction in [fluxcd/flux2#3364](https://github.com/fluxcd/flux2/issues/3364), now one of the most upvoted feature requests in the Flux project.

`spec.dependsOn` currently only works between resources of the same kind. A `Kustomization` can depend on another `Kustomization`, and a `HelmRelease` can depend on another `HelmRelease`, but you can't mix them or express a dependency on any other Kubernetes resource.

Using `spec.healthChecks` as a workaround has real limits. Health checks run post-apply and affect when a resource becomes Ready, which in turn unblocks dependants. However, they do not gate the resource’s own reconciliation. Gating on another resource requires that resource to define appropriate health checks, which you neither control nor can guarantee.

The Flux Operator already solved this in the [`ResourceSet`](https://fluxcd.control-plane.io/operator/resourceset/#dependency-management) API. Bringing it to the Applier APIs (`Kustomization` and `HelmRelease`) would open it up to the wider Flux user base who don't (yet) use the Operator, reducing the need for wrapper `Kustomization` resources used solely to enforce apply ordering.

### Proposed changes

Extend the `DependencyReference` type with `apiVersion` and `kind` fields to allow dependencies on arbitrary Kubernetes resources.

~~Add a `ready` field, mirroring the Flux Operator's `ResourceSet` `Dependency` type. This allows gating on existence rather than readiness. It is particularly useful for resources that do not expose a `Ready` condition, without requiring custom health checks.~~

#### <ins>Edit:</ins>
The existing `readyExpr` field can be used to gate on existence rather than readiness by setting its value to `"true"`. Since this expression always evaluates to true, no readiness check will be performed. Note that this approach has no effect when the `AdditiveCELDependencyCheck` feature gate is enabled on an Applier controller. As a result, the proposed `ready` field will not be added to the API.

When these fields are omitted, reconciliation should apply defaults so existing `Kustomization` and `HelmRelease` entries in `dependsOn` continue to behave as before.

```diff
  spec:
    dependsOn:
      - name: cert-manager
        namespace: flux-system
+     - apiVersion: apiextensions.k8s.io/v1
+       kind: CustomResourceDefinition
+       name: clusterissuers.cert-manager.io
[...]

PR #5988, opened 2026-07-10
[RFC-XXXX] Advanced Dependency Management

vecil:
Relates: #5879

Preview: https://github.com/vecil/flux2/blob/docs/propose-rfc-0014-advanced-dependency-management/rfcs/xxxx-advanced-dependency-management/README.md

This proposes an RFC that defines how Applier controllers (Kustomization, HelmRelease) could handle multi-hop dependencies on arbitrary resource kinds, including cross-namespace and cluster-scoped resources, with support for implicit (kstatus-based), explicit (user-defined CEL expressions), and existence-only dependency checks, while also proposing to unify the Applier and ResourceSet APIs around dependency management.

The document clarifies each design choice alongside already-debated alternatives, noting backward-compatibility throughout and surfacing caveats (e.g. interactions between kstatus-based built-in checks and resources lacking a `Ready` condition) with proposed mitigations for each.

### DIFF
--- a/rfcs/xxxx-advanced-dependency-management/README.md
+++ b/rfcs/xxxx-advanced-dependency-management/README.md
@@ -1,0 +1,371 @@
+# RFC-XXXX Advanced Dependency Management
+
+**Status:** provisional
+
+<!--
+Status represents the current state of the RFC.
+Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+-->
+
+**Creation date:** 2026-07-10
+
+**Last update:** 2026-07-14
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


artem-nefedov:
Does this cover the long-requested feature to handle deletion in the correct order?
It's probably worth specifying in Goals or Non-Goals in the RFC.

ringods:
Projects offering core infrastructure services for Kubernetes often make use of other core services too:
* most of these services depending on monitoring services
* monitoring services depending on deployed storage
* longhorn offering this storage, but a helm release of this can block because monitoring services haven't been deployed.

I find it pretty hard to model the correct order of such services. I'm also pretty new to building out a fully equipped Kubernetes cluster.

Still, this RFC would expand the possibilities to use `dependsOn` on a more granular level.

PR #6000, opened 2026-07-19
Guard against Kustomization tree cycles

### LINKED ISSUE #5998: flux tree: unbounded recursion when Kustomization inventories reference each other
### Describe the bug

`flux tree kustomization` has no guard against cyclic references between Kustomization inventories. `treeKustomization` (cmd/flux/tree_kustomization.go) recurses into every child Kustomization found in `status.inventory`; the only skip is a direct self-reference. If two Kustomizations apply *each other's* manifests, their inventories reference each other and the command recurses A→B→A→… issuing API GETs in a tight loop until the `--timeout` context (default 5m) expires, then exits with:

```
✗ failed to find object: Get "https://.../kustomizations/treebomb-a": context deadline exceeded
```

and no output. With the default timeout that is 5 minutes of hammering the API server for nothing.

### Steps to reproduce

Git repository layout:

```
treebomb/a/ks-b.yaml   # Kustomization treebomb-b, path ./treebomb/b, same GitRepository
treebomb/b/ks-a.yaml   # Kustomization treebomb-a, path ./treebomb/a, same GitRepository
```

1. `kubectl apply` a Kustomization `treebomb-a` with `path: ./treebomb/a`.
2. It applies `treebomb-b`, which in turn applies (adopts) `treebomb-a`. Both become Ready, each holding the other in `status.inventory`.
3. `flux tree kustomization treebomb-a` → spins until `--timeout`, then `context deadline exceeded`, no output.

Verified on flux v2.8.8 (Kubernetes v1.34); the code path is unchanged on main.

### Expected behavior

A visited-set in the recursion: print each Kustomization node once (optionally marking the back-reference) and stop descending. Related: #5291 proposes ordering tree nodes by `dependsOn` — an implementation would touch the same recursion and needs the same cycle-tolerance.

### Flux version

v2.8.8

AkashKumar7902:
Closes #5998

## What

Guard `flux tree kustomization` recursion with the current traversal
path. Cyclic inventory paths now stop at an ancestor instead of querying
the API server until the command times out.

Path-local tracking keeps shared non-cyclic subtrees traversable under
independent branches.

## Tests

- `make tidy fmt vet && make test`

## Local testing

- Built `flux` from `e8594afcd6c29c9a573464892d0c645c1e2cdf65`.
- Against a disposable Kubernetes 1.36.2 envtest API server, seeded a
  two-object Kustomization inventory cycle. The base binary exited 1 at
  the two-second timeout; the candidate exited 0 and printed `cycle-a`
  followed by `cycle-b` once each.

Assisted-by: codex/gpt-5

### DIFF
--- a/cmd/flux/testdata/tree/kustomizations.yaml
+++ b/cmd/flux/testdata/tree/kustomizations.yaml
@@ -86,3 +86,38 @@ status:
       status: "True"
       type: Ready
 ---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: cycle-a
+  namespace: {{ .fluxns }}
+spec:
+  interval: 5m
+  path: ./cycle-a
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: flux-system
+status:
+  inventory:
+    entries:
+      - id: {{ .fluxns }}_cycle-b_kustomize.toolkit.fluxcd.io_Kustomization
+        v: v1
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: cycle-b
+  namespace: {{ .fluxns }}
+spec:
+  interval: 5m
+  path: ./cycle-b
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: flux-system
+status:
+  inventory:
+    entries:
+      - id: {{ .fluxns }}_cycle-a_kustomize.toolkit.fluxcd.io_Kustomization
+        v: v1
--- a/cmd/flux/testdata/tree/tree-cycle.golden
+++ b/cmd/flux/testdata/tree/tree-cycle.golden
@@ -1,0 +1,3 @@
+Kustomization/{{ .fluxns }}/cycle-a
+└── Kustomization/{{ .fluxns }}/cycle-b
+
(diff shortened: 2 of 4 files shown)


gecube:
Hi @AkashKumar7902,
Thanks for taking the time to work on this and for trying to address the issue I reported!
One question came to mind though: how does this approach behave with more complex dependency graphs?
For example, not just a simple A <-> B cycle, but something like:

```
A -> B -> C -> D -> ... -> Z -> A
```

or even graphs where multiple cycles are interconnected, for example:

```
        +---- C ----+
        |           |
A -> B -+-> D -> E -+-> F
^        \         /
|         \-> G <-/
+------------- H --+
```

or:

```
A -> B -> C -> D
^         |    |
|         v    |
F <- E <- G <- H
```

or a graph with nested/overlapping cycles:

```
A -> B -> C -> D -> E
     ^    |         |
     |    v         v
     H <- G <- F <- I
          ^         |
          +---------+
```

Those are usually much harder to detect correctly than the trivial two-node case.
More generally, dependency graph analysis is a surprisingly difficult problem. Garbage collection and cycle detection are areas where, as far as I know, the industry still has to balance correctness against performance. Simple loop-detection approaches often miss more complicated topologies, while more advanced graph algorithms can become quite expensive in terms of CPU and memory on large dependency graphs.
It would be interesting to understand where this implementation sits on that spectrum and whether there are any known limitations regarding longer or overlapping cycles.

PR #6016, opened 2026-07-27
Add digests flag to plugin search, which displays SHA values for all os/arch per binary

### LINKED ISSUE #6008: Improve UX to install/pin plugins by SHA
The new plugin system allows to specify plugin versions using: (implicit) latest, by version and by SHA.

In the last case (by SHA), it is currently not possible to get the required SHA value directly from the Flux CLI. [The PR](https://github.com/fluxcd/flux2/pull/5872) that introduced pinning by SHA fetches these value from the GitHub release. This works, but is not obvious to users who are not familiar with the internal workings of the plugin system.

I would argue that one can place similar levels of trust in the Flux CLI and the GitHub API. This would allow us to fetch and display the SHA for a given version directly in the Flux CLI during `flux plugin search`.

---

I would propose to make two changes: 

1. Add a new flag `--digests` to `plugin search` which displays the SHA **in addition** to the version. Similar to what [docker image ls](https://docs.docker.com/reference/cli/docker/image/ls/#digests) does
2. Add high level documentation to https://fluxcd.io/flux/cli-plugins/ which explains the trade-off between the different version references and the intended use-cases and workflows.

If you agree with these changes I am happy to take a stab at a PR for this!

datosh:
### Changes

+ Added `--digests` flag to `flux plugin search`

```
› ./bin/flux plugin search --digests      
operator  0.57.0  Flux Operator CLI
├── darwin/amd64   sha256:dea268f9ae341c81e77bd5dda25f0c94a543c6acf629cd79e2de30f9e999a94b
├── darwin/arm64   sha256:4dc9b3026f3b2d9fc7f13d9f6f5805036cff7d0af94d684abc928431c914138a
├── linux/amd64    sha256:abd705d2772912c482379103163df9537b4828326d29be114ec1da1cbf7a81b3
├── linux/arm64    sha256:ac2ac47b55dc42459ab8de71ddb1e8687597e80320dbba43ace5d9904f219d1d
├── windows/amd64  sha256:d0161b5aabc2016a12bbcd2763d8da0a59668e0e163df73ffafc57bfaa3010ff
└── windows/arm64  sha256:0b9fe5e1f600426f0f1896d7c8aeee4ca1ba47f87f3e2c1ea9800fd847f21b50
schema  0.11.0  Kubernetes schema extraction and manifests validation
├── darwin/amd64   sha256:0ec55397e49a2ac76a34ae04f99cdfd26ec814fd1475fa2d821611be16294548
├── darwin/arm64   sha256:7ef017fa6d123e7fa9805668fcea2d7baca2a61e1a6f08bb21a48ae57f055b96
├── linux/amd64    sha256:f208a99e829c5bb1df1ae290438161a7bc3ce1d2caad949a8a9c09bfb2a496ac
├── linux/arm64    sha256:602325d27b036abd74c166c857705a578338bf2d1eada81928ac3192e3a140ae
├── windows/amd64  sha256:9ae033656824163e82068393164a102b044d98a7567054cb448a6f05b56ce2ca
└── windows/arm64  sha256:87313c25de041f169f680cc9054f49f0a9209259ec8902446f3fcd622dc73dc8
mirror  0.8.0  Helm charts, OCI artifacts and container images mirroring
├── darwin/amd64   sha256:52f028aa0d60a335219fe9b06f18288d2a082163ce0661af4d66d7133c7edbf0
├── darwin/arm64   sha256:4e4a974f6acaa15e808381d7fa19ca03ce2b0556675da99b51f51599d3839836
├── linux/amd64    sha256:4862e9c4a2c9ac6bfe4976766ec425b015f2e6730adebcf95184d1d94beb5b35
├── linux/arm64    sha256:71222a7bcfe0715da9481845b34b60df68034c1be386d59c3c3cef8c76fca6f3
├── windows/amd64  sha256:cdd399ce097b4ae8f0a92f126d73672be7c9fb80c4728e960e966a26c981d2ff
└── windows/arm64  sha256:c0339e7c64f56c348117fdd0d04d0b6ce5f618dabff1c1d35cde16a3fc058d03
```

+ Added an optional `[@version]` field to `flux plugin search [query]`

```
› ./bin/flux plugin search operator@0.45.0
operator  0.45.0  Flux Operator CLI
├── darwin/amd64   sha256:8f8f6d04adb6a2451be1e5585753cf111536cf4a95b3f4716680d6b14f6bb667
├── darwin/arm64   sha256:cd85d5d84d2646a0bf991713fd2ee83e5c27a629094be447d4acd67fe5c54dea
├── linux/amd64    sha256:96198da9690965df56488a2a4a8b39cf56e2dc217a622446d5f00b1425f89fe8
├── linux/arm64    sha256:9c20eeca130f4a3374951dedfe12b89d7b093b521560f37f09bea39748a62142
├── windows/amd64  sha256:9712026094a5c99efbbd8feabbf7bc30b737e9f93e0f534f14129b4ac1ed796b
└── windows/arm64  sha256:9481a7a8d8111c8e9ff8de99bd10ef653aeeb7c2271e778b2c3e75c19b876f9a
```

Example output from https://github.com/fluxcd/flux2/pull/6016/commits/6b6f078614a6e89eeff376551c6708d88fbe25d6

### Implementation Decisions

I would like to call out these decisions and invite feedback on the direction here!

#### Tree View

`flux plugin search` uses the columns NAME, DESCRIPTION, and INSTALLED. To pin by hash, we need at least two (additional) pieces of information: OS/ARCH and DIGEST.

To efficiently display that information and prevent duplicate values, a tree-based view is used for digests.

#### Implied --digests

`flux plugin search operator` uses the standard output format.

`flux plugin search operator --digests` uses the "pinning relevant" output format. 

`flux plugin search name@1.2.3` implies `--digests` and therefore presents the "pinning relevant" information, since the user wants more information about a specific version. 

Are we okay with this, or should we require the explicit use of --digests?

#### Potentially Large Output

For each new plugin added to the catalogue, `flux plugin search --digests` will produce 6 rows of output, more if we decide to support more architectures or operating systems.

I think this is fine, since `--digests` has to be set explicitly, and filtering down using the plugin name is possible.

#### Sequential Fetching

To display the SHA values, the manifest for each plugin has to be fetched. To keep the implementation simple, I opted for a simple loop instead of parallel fetching.

I think this is reasonable right now, but it might be slow if the number of plugins grows substantially.

fixes #6008

### DIFF
--- a/cmd/flux/main_test.go
+++ b/cmd/flux/main_test.go
@@ -455,6 +455,7 @@ func resetCmdArgs() {
 	imageUpdateArgs = imageUpdateFlags{}
 	installArgs = newInstallFlags()
 	kustomizationArgs = NewKustomizationFlags()
+	pluginSearchArgs = pluginSearchFlags{}
 	receiverArgs = receiverFlags{}
 	resumeArgs = ResumeFlags{}
 	rhrArgs = reconcileHelmReleaseFlags{}
--- a/cmd/flux/plugin.go
+++ b/cmd/flux/plugin.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
--- a/cmd/flux/plugin.go
+++ b/cmd/flux/plugin.go
@@ -112,7 +112,7 @@ func newCatalogClient() *plugin.CatalogClient {
 }
 
 func newPluginSpinner(message string) *spinner.Spinner {
-	s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
+	s := spinner.New(spinner.CharSets[14], 100*time.Millisecond, spinner.WithWriterFile(os.Stderr))
 	s.Suffix = " " + message
 	return s
 }
--- a/cmd/flux/plugin_search.go
+++ b/cmd/flux/plugin_search.go
@@ -95,13 +95,14 @@ func pluginSearchCmdRun(cmd *cobra.Command, args []string) error {
 	var rows [][]string
 	if digests {
 		header = []string{"NAME", "VERSION", "OS/ARCH", "DIGEST"}
-		rows, err = pluginDigestRows(catalogClient, entries, version)
+		var warnings []error
+		rows, warnings = pluginDigestRows(catalogClient, entries, version)
+		for _, w := range warnings {
+			logger.Warningf("%s", w)
+		}
 	} else {
 		header = []string{"NAME", "DESCRIPTION", "INSTALLED"}
 		rows = pluginCatalogRows(entries)
-	}
-	if err != nil {
-		return err
 	}
 
 	if len(rows) == 0 {
@@ -135,8 +136,9 @@ func pluginCatalogRows(entries []plugintypes.CatalogEntry) [][]string {
 }
 
 // pluginDigestRows fetches the manifest of every entry and returns one row per
-// os/arch for the binary of the requested version.
-func pluginDigestRows(catalogClient *plugin.CatalogClient, entries []plugintypes.CatalogEntry, version string) ([][]string, error) {
+// os/arch for the binary of the requested version. Any errors encountered when
+// fetching plugin information are recorded and plugins are skipped in output.
+func pluginDigestRows(catalogClient *plugin.CatalogClient, entries []plugintypes.CatalogEntry, version string) ([][]string, []error) {
 	if len(entries) == 0 {
 		return nil, nil
 	}
@@ -146,14 +148,17 @@ func pluginDigestRows(catalogClient *plugin.CatalogClient, entries []plugintypes
 	defer sp.Stop()
 
 	var rows [][]string
+	var warnings []error
 	for _, entry := range entries {
 		manifest, err := catalogClient.FetchManifest(entry.Name)
 		if err != nil {
-			return nil, err
+			warnings = append(warnings, err)
+			continue
 		}
 
 		pv, err := plugin.ResolveVersion(manifest, version)
 		if err != nil {
+			warnings = append(warnings, err)
 			continue
 		}
 
@@ -167,5 +172,5 @@ func pluginDigestRows(catalogClient *plugin.CatalogClient, entries []plugintypes
 		}
 	}
 
-	return rows, nil
+	return rows, warnings
 }
--- a/cmd/flux/plugin_search.go
+++ b/cmd/flux/plugin_search.go
@@ -17,49 +17,112 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/fluxcd/flux2/v2/internal/plugin"
+	plugintypes "github.com/fluxcd/flux2/v2/pkg/plugin"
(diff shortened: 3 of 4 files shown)


matheuscscp:
Thanks very much for this PR @datosh!

codex/gpt-5.5 xhigh found a couple of nits:

* This command's main output is the table, and with `--digests` it starts doing network calls, hence why it makes sense to have the spinner. Codex suggested printing the spinner in stderr to separate the streams (table goes to stdout)
* We are ignoring the errors from `plugin.ResolveVersion(manifest, version)`, and (this one I saw myself) we are failing the whole command on `catalogClient.FetchManifest(entry.Name)`. I'm wondering if we shouldn't both capture all errors and print them all on stderr instead of failing the whole command. Then we can either skip the rows that failed or print a blank or indicative msg that it errored out.

datosh:
Thanks for the feedback @matheuscscp 

> * This command's main output is the table, and with `--digests` it starts doing network calls, hence why it makes sense to have the spinner. Codex suggested printing the spinner in stderr to separate the streams (table goes to stdout)

`plugin install` has the same issue, so I opted to fix this in the shared `spinner.New`.

> * We are ignoring the errors from `plugin.ResolveVersion(manifest, version)`, and (this one I saw myself) we are failing the whole command on `catalogClient.FetchManifest(entry.Name)`. I'm wondering if we shouldn't both capture all errors and print them all on stderr instead of failing the whole command. Then we can either skip the rows that failed or print a blank or indicative msg that it errored out.

Good catch! I have added code to record the errors and print them as warnings. Let me know if you agree with that approach!

matheuscscp:
Thanks @datosh! One last nit: what happens if all digest lookups fail? Wondering if the command output will make sense in that case :thinking:

datosh:
> Thanks @datosh! One last nit: what happens if all digest lookups fail? Wondering if the command output will make sense in that case 🤔

Since we have the `len(rows) == 0` guard in place, it won't print the table, but of course, one line for each failed lookup. I would say this is acceptable behaviour. Would you prefer something else? 

The exit code is `0`, which might be reasonable to change if `len(warnings) == len(plugins)`...? 

---

Do you agree with the other "implementation decisions" I laid out in the PR description?

matheuscscp:
@datosh Yep, they make sense!

About the failure for all digests, looks like we enter `if len(rows) == 0` and get one of:

  - no query: `No plugins found in catalog`
  - query present: `No plugins matching "..." found in catalog`

Which is a bit confusing, since there are plugins but we actually failed fetching digests for all of them...

matheuscscp (approved):
LGTM! 🚀 

We can merge this either way, the comment above is quite a corner case 👍

datosh:
I have switched the implementation to a tree view to reduce duplicate information.

I have also revisited the corner case you described @matheuscscp , and improved the error handling in  https://github.com/fluxcd/flux2/pull/6016/commits/92fa38d7e6a6fd77011cd5ef1e8acd2703c801c2

PTAL whenever you have the chance.

stefanprodan:
@datosh please update the PR body with the new tree format.

datosh:
> @datosh please update the PR body with the new tree format.

Thanks for the reminder - description is updated and matches current implementation.

PR #6020, opened 2026-07-29
Add support for 2.9 in `migrate -f`

matheuscscp:
Filing also a follow-up PR in the skill for minor releases.


stefanprodan (approved):
LGTM

PR #6074, opened 2026-09-16
Format reproducible created timestamp in UTC

lexfrei:
`flux push artifact --reproducible` says it sets the created timestamp to `1970-01-01T00:00:00Z`, but it formats `time.Unix(0, 0)`, which is in local time. So the `org.opencontainers.image.created` annotation depends on `TZ`, and the manifest digest too.

Same directory, `--source` and `--revision`, pushed under three time zones:

| TZ | Manifest digest | `org.opencontainers.image.created` |
| --- | --- | --- |
| UTC | `sha256:40af310a…` | `1970-01-01T00:00:00Z` |
| Europe/Moscow | `sha256:78e418f3…` | `1970-01-01T03:00:00+03:00` |
| Asia/Tokyo | `sha256:1d496cc8…` | `1970-01-01T09:00:00+09:00` |

Layer and config blobs are the same, only the manifest differs. A digest pinned on a machine outside UTC never matches what CI pushes from the same tree.

Fixed by formatting the zero time in UTC. All three pushes now give `sha256:40af310a…`, same as UTC users get today. The new test sets `time.Local` to UTC+9 and fails without the fix.

I used an LLM to help with this change and reviewed it myself.


Karthik-Chowdary (approved):
The explicit `UTC()` is necessary here: in this Go toolchain `time.Unix` returns a `Time` in `time.Local`, so RFC3339 formatting otherwise changes both the annotation and manifest digest with the process zone. The regression test exercises the real push/manifest path under a non-UTC `time.Local`, restores global state with cleanup, and asserts the canonical OCI annotation. I also checked that the production change is narrowly scoped to reproducible mode. A focused local run could not compile because this checkout lacks the generated `cmd/flux/manifests/*.yaml` embed inputs; I did not treat that environment failure as a test pass.

matheuscscp (approved):
LGTM! 🚀 

Thanks @lexfrei, nice one!

